PR #5, opened 2020-05-10
Add eslint/prettier with airbnb base with typescript integration

### LINKED ISSUE #1: Setup and config eslint and prettier
use eslint airbnb please

nabeel-:
Closes #1 

I opted to disable any errors that weren't obvious fixes and leave the warnings for now since they won't actually fail the eslint run. Happy to reconfigure/add/remove any rules!


package.json
```diff
@@ -8,7 +8,8 @@
     "build": "next build",
     "start": "next start",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "test": "eslint . --ext=tsx,ts"
```
mmahalwy, line 12: can you create a separate script for this to be just `lint` and `test` would be `lint` and soon to be jest tests too :)

mmahalwy:
wow @nabeel- thanks so much for the quick turn around :)

package.json
nabeel-: Done!

mmahalwy:
@nabeel- got a few more issues if you're up for it :)

PR #32, opened 2020-06-10
Allow theme to be typed for ease of consumption

nabeel-:
Follows the [styled-components typescript docs](https://styled-components.com/docs/api#typescript). It may be worth doing something more complex (e.g., a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions)) here if we envision the two themes diverging in attributes.


mmahalwy:
@nabeel- any luck with this? 🙏

nabeel-:
@mmahalwy Any insight into why the deploy is failing?

PR #41, opened 2020-06-26
Add pull_request event to GH action triggers and fix a lint warning

nabeel-:
Mainly to test a theory on the GH actions wonkiness


nabeel-:
oh. @mmahalwy it doesn't trigger since these are pushes to a _fork_ of the repo and not the repo itself 🤦

Issue #20, opened 2020-06-04
Add Husky as a pre-commit hook


mmahalwy:
https://www.npmjs.com/package/husky
- Run prettier before committing

nabeel-:
Have you thought about setting up [a github action](https://github.com/features/actions) to run tests or linting for PRs/commits? In addition to Husky for pre-commit hooks, we can have the PR build/run tests without having to set up travis/jenkins/another CI tool.

mmahalwy:
Yes. I'd like to. Can you own it @nabeel- ? I can give you collaborator access.

nabeel-:
Of course - happy to own it!

PR #45, opened 2020-06-27
Add husky & lint-staged

mmahalwy:
Fixes https://github.com/quran/quran.com-frontend-next/issues/20
cc @nabeel-

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,16 @@
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^4.0.0",
+    "husky": "^4.2.5",
     "prettier": "^2.0.5",
     "react-docgen-typescript-loader": "^3.7.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "start": "next start",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "lint": "eslint . --ext=tsx,ts",
+    "lint": "eslint . --ext=tsx,ts,json --fix",
```
nabeel-, line 12: Thoughts on breaking this out into a `lint:fix` script that will apply fixes? Seems a bit wonky to have CI (since it runs `yarn test`, which in turn runs `yarn lint`) auto-fix potential errors and not commit them anywhere.

mmahalwy: good call.

nabeel-: Do we need an eslint json plugin to support linting against `.json` files?

mmahalwy: removed this

mmahalwy:
@nabeel- addressed PR comments :)

PR #48, opened 2020-06-27
Allow empty for storybook deployment

mmahalwy:
@Anasss added this. Can we have a longer term solution for this? It's breaking CI on master atm


Anasss:
Thanks @mmahalwy! I'll have a look, we should commit only there is some diffs with origin/gh-pages! Also, I would rather keep only storybook/master on the gh-pages branch, it's cleaner

PR #70, opened 2020-07-03
Add redux-persist

tifa2UP:
### Summary
Added [redux-persist](https://github.com/rt2zz/redux-persist) to persist parts of our redux store in localstorage across sessions. This will be useful for adding features such as showing the last read position.

Followed the set-up guide here: https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist.

Will create a follow-up PR insha'Allah to demo the values being saved in localstorage and persisting across sessions. 

### Test Plan
Confirmed that a `persist:root` object is being created in localstorage (see screenshot). No observable changes in the app behavior. Tested the usage and confirmed that the whitelisted values persist in #71. 

### Screenshots

image

### DIFF
--- a/src/pages_/_app.tsx
+++ b/src/pages_/_app.tsx
@@ -2,19 +2,22 @@ import React from 'react';
 import { DefaultSeo } from 'next-seo';
 import { ThemeProvider } from 'styled-components';
 import { Provider } from 'react-redux';
+import { PersistGate } from 'redux-persist/integration/react';
 import { theme } from '../utils/styles';
 import { createSEOConfig } from '../utils/seo';
 import Footer from '../components/Footer';
-import store from '../redux/store';
+import store, { persistor } from '../redux/store';
 
 function MyApp({ Component, pageProps }): JSX.Element {
   return (
     <Provider store={store}>
-      <ThemeProvider theme={theme}>
-        <DefaultSeo {...createSEOConfig({})} />
-        <Component {...pageProps} />
-        <Footer />
-      </ThemeProvider>
+      <PersistGate loading={null} persistor={persistor}>
+        <ThemeProvider theme={theme}>
+          <DefaultSeo {...createSEOConfig({})} />
+          <Component {...pageProps} />
+          <Footer />
+        </ThemeProvider>
+      </PersistGate>
     </Provider>
   );
 }
--- a/src/pages_/_app.tsx
+++ b/src/pages_/_app.tsx
@@ -1,24 +1,20 @@
 import React from 'react';
 import { DefaultSeo } from 'next-seo';
 import { ThemeProvider } from 'styled-components';
-import { Provider } from 'react-redux';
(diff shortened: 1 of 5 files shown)


src/pages_/_app.tsx
```diff
@@ -2,19 +2,22 @@ import React from 'react';
...
 
 function MyApp({ Component, pageProps }): JSX.Element {
   return (
     <Provider store={store}>
-      <ThemeProvider theme={theme}>
-        <DefaultSeo {...createSEOConfig({})} />
-        <Component {...pageProps} />
-        <Footer />
-      </ThemeProvider>
+      <PersistGate loading={null} persistor={persistor}>
```
tifa2UP, line 14: The loading time is too small imo to add a loading component, it's helpful for apps that store their full redux store in localstorage. We can reassess and change it as our app gets bigger.

nabeel-: Is the loading time negligible across all devices? If there are some low-powered devices that might experience a noticeable delay we could get the best of both worlds by having a “delayed spinner” that will only kick in after a timeout. In that case, we avoid flashing a loading spinner on devices where the delay is minimal and show indications to the user that something is happening on lower-powered devices that take a little longer.

Not something to block this particular PR on - just a drive-by thought!

tifa2UP: 1) Yeah, the loading time is niglible on all devices, if there's no localstorage data, the `PersistGate` component loads the application right away. So there's no observable difference for first-time users and web crawlers. For returning users, the storage payload is pretty small, and the data is read locally, I'm suspecting that it'll be neglible too. 

2) I'm open to the idea of having a loading component that appears after 500ms or so, we should do a performance rundown once we have a working version of the app iA and identify the bottlenecks :)

mmahalwy: Hmm. after looking at the code, this will show empty on SSR, no?

tifa2UP: Good catch @mmahalwy. Created `Provider.tsx` to fix this issue. 

We now only load `<PersistGate />` on the client. I confirmed that the DOM is loaded for SSG, and that redux-persist still works as expected

AhmedCodeGuy:
Looks good to me!

tifa2UP:
> With changes to src/redux/Provider.tsx, this would still cause a diff in server rendering and client mounting. Might be some inconsistency here

This seems to be recommended approach to use redux-persist with SSR. Do you have concerns for not rendering `<PeristGate />` on the server?

mmahalwy:
> > With changes to src/redux/Provider.tsx, this would still cause a diff in server rendering and client mounting. Might be some inconsistency here
> 
> This seems to be recommended approach to use redux-persist with SSR. Do you have concerns for not rendering `<PeristGate />` on the server?

I feel like this probably would have an inconsistent flash. Let's try it though?

Issue #33, opened 2020-06-10
Eslint prettier rules not applying to json & yml files


hazem3500:
I found that when the prettier rules are inside eslint they don't work on files like `json` and `yml` so they won't be auto-formatted, maybe we can find a way to make eslint also work on `json` and `yml` files.

papasmile:
Asslaam alaykum wa rahmatullah,

Nice idea!  Firstly, as JSON files are supported via their own plugin in ESLint:

"you should run prettier over your json files in a separate step instead of trying to smush it into your eslint workflow" https://github.com/prettier/eslint-plugin-prettier/issues/294

That said, I was pretty easily able to add on `prettier` commands to existing `lint` and `lint:fix` scripts.  PR open at:  https://github.com/quran/quran.com-frontend-next/pull/74

njmulsqb:
Kindly assign this issue to me

papasmile:
Assalaam alaykum, I'd also be happy to assist, you @njmulsqb.  My PR was closed with no justification... but willing to give another go.

njmulsqb:
> Assalaam alaykum, I'd also be happy to assist, you @njmulsqb. My PR was closed with no justification... but willing to give another go.

Ws. Is this project active? Are some reviewers available to actively review PRs?

Issue #21, opened 2020-06-04
Add prettier.config.js file


mmahalwy:
Can be:
```js
module.exports = {
  "arrowParens": "always",
  "bracketSpacing": true,
  "jsxBracketSameLine": false,
  "printWidth": 100,
  "proseWrap": "always",
  "requirePragma": false,
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "all",
  "useTabs": false
};
```

Thoughts @nabeel- ?

nabeel-:
The rules lgtm!

As for the setup, if we want to use prettier as the formatter we'll need to remove the plugin from the `.eslintrc.yml` so we don't have conflicts when it comes to formatting.

The other option would be to add these rules as part of the `.eslintrc.yml`:

```
prettier/prettier:
  - 2
  - arrowParens: always
     bracketSpacing: true
     jsxBracketSameLine: false
     printWidth: 100
     proseWrap: always
     requirePragma: false
     semi: true
     singleQuote: true
     tabWidth: 2
     trailingComma: all
     useTabs: false
```

Do you have a preference one way or the other?

mmahalwy:
Hmm, you sure we can do both? So eslint would use the `prettier.config.js` and, I guess, the editor would read from that file too?

nabeel-:
> you sure we can do both

I'm not sure we can do both. The idea is to have eslint be responsible for formatting via [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier). We can add the rules to the `.eslintrc.yml` and we wouldn't need to have any `prettier.config.js` file.

> the editor would read from that file too

The editor will just need the `eslint` plugin (like [this one for VSCode](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)) and eslint will take care of running prettier under the hood. Prettier errors will surface as eslint errors and still potentially be auto-fixable.

PR #74, opened 2020-07-08
Add Prettier Support For JSON And YML

papasmile:
### Summary

Check and Clean `.yml` and `.json` files.

* Updates `prettier` configuration per `eslint-plugin-prettier` recommendations https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
* Adds on `prettier` cli commands for json and yml files
* Two commits:
  * updated config https://github.com/quran/quran.com-frontend-next/commit/04e502e3b1752518486469664b56b516c0057c7b 
  * actual formatting updates using new config https://github.com/quran/quran.com-frontend-next/commit/82be0d937f4d86f464668ab60546f4d9dce99960

### Test Plan

Try it out!

`yarn lint`

and you should see e.g.

```
Checking formatting...
All matched files use Prettier code style!
✨  Done in 13.39s.
```

or use

 `yarn lint:fix`.

to auto-fix

Unit tests still pass and sever starts normally.

### Screenshots

n/a


.prettierrc
```diff
@@ -0,0 +1,11 @@
...
+bracketSpacing: true
+jsxBracketSameLine: false
+printWidth: 100
+proseWrap: always
+requirePragma: false
+semi: true
+singleQuote: true
+tabWidth: 2
+trailingComma: all
+useTabs: false
```
tifa2UP, line 11: I recommend going with the default prettier options and not have a `prettierrc` to be consistent across different quran.com projects.

papasmile:
Salam, @tifa2UP .  Sorry, what command did you run to get those errors?  As noted I get zero....

mmahalwy:
cc @nabeel- on this

.prettierrc
nabeel-: It looks like this is just a copy-paste from the `.eslintrc.yml`, which are the rules outlined in #21. Do you think we should change any rules here (probably best for a follow-up in any case)?

package.json
```diff
@@ -9,8 +9,9 @@
     "start": "next start",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "lint": "eslint . --ext=tsx,ts,js,jsx",
-    "lint:fix": "yarn lint --fix",
+    "lint": "yarn lint:eslint && prettier --check '**/*.(yml|json)'",
+    "lint:eslint": "eslint . --ext=tsx,ts,js,jsx",
+    "lint:fix": "yarn lint:eslint --fix && prettier --write '**/*.(yml|json)'",
```
nabeel-, line 14: This is unfortunately the kind of thing the current setup wanted to avoid. It's odd that eslint would be doing the same work for the file extensions provided as prettier is doing in the one-off cli calls.

.prettierrc
papasmile: +1 for getting rid of `.prettierrc` if no one is hung op on those preferences :)

package.json
papasmile: Salam @nabeel- et al, I referenced plugin owner's comments in #33, but to repeat:

> You shouldn't pass non-js files into eslint unless you've got a plugin to explicitly handle that (and even then interaction with this plugin gets a little weird). Instead you should run prettier over your json files in a separate step instead of trying to smush it into your eslint workflow... eslint-plugin-prettier aims to be just a rule - it deliberately doesn't want to start wrangling ESLint preprocessors as that's out of scope for what I want it to to care about

https://github.com/prettier/eslint-plugin-prettier/issues/294

In other words: *ESlint is for code, not config and other things, while prettier is more general.*  So if you want something cleaner, your going to have to write your own pagackge/scripts to wrap around eslint and prettier together to achieve a 'nice' interface.

In conclusion: What I've proposed is not overly long/ugly in terms of scripts.  But if other folks disagree now or after adding more stuff and would like to seem something cleaner then a next simple step would be some scripts with nice code comments (e.g. ./scripts/format-files) to clean up package.json

tifa2UP:
> Salam, @tifa2UP, would like to get you to clarify #74 (comment) before pressing forward to make sure we are in good shape, JAK

Thank you for the follow-up @papasmile, I added the build file under the eslint ignore list and no longer get these errors. It's good to be merged once @nabeel- gives a ship-it.

mmahalwy:
@papasmile can you rebase? Let's get this merged!

papasmile:
salam, @mmahalwy , sure thing!  Rebased, but it seems some folks have been adding optional fields with no default props :)  Someone want to fix those or shall I attempt???

papasmile:
salam, rebased @mmahalwy , someone fixed up those lint issues and have none (just 1 warning) so should be good to merge, sir

papasmile:
some funny permissions error on deploy-storybook...

papasmile:
asslaam alaykum wa rahmatullah @mmahalwy  and @tifa2UP if there is any way i can assist please advise, would be happy to make effort to get PRs moving insha Allah t'ala

PR #75, opened 2020-07-09
Refactor the QuranWord component

tifa2UP:
### Summary
Refactored the word component following the design in the design doc. The QuranWord component renders a word as it appears in a Mushaf. 

I put the different text rendering styles under separate components: `<UthmaniWordText>`, `<IndoPakWordText>`, and `<MadaniWordText>`. 

The (future) shared functionality such as highlighting, tooltips, and onClick events will live in the parent `QuranWord` component.  

These technically should be in a separate PR, but I made them while I was at it: 
- run prettier on `eslintrc.yml`
- disable required return types for TS, these could be automatically inferred by the compiler
- Added the root as an absolute path. We can do import from `'types/..'`, `'src/..'`, etc. This made me modify .eslintrc and the storybook webpack config
- Sorted the compilerOptions in `tscofig.json`

### Notes
- This currently does not support ayah markers (ayah number) for madani and indopak font, I'll consult with design to see if we want to reuse the Uthmani one. 
- I removed the font sizing from the component as that shouldn't be the property of the component. It will be should be set by the container. I will however apply a font scaling factor as the uthamni font appears smaller than the madani and indopak one.
- The text direction (rtl) should be set on the block level element (the Verse component), that'll be in a follow PR.

### Test Plan
Rendered different font scripts in the `<VerseText>` component, confirmed that the fonts load differently and as expected for the different font types. 

### Screenshots
Surah Alfatiha with fontFamily set to `uthmani`: 
image

Surah Ali-Imran with fontFamily set to `uthmani`:
image

You can view more variations in the storybook build


src/components/dls/QuranWord/IndoPakWordText.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import styled from 'styled-components';
+
+type IndoPakWordTextProps = {
+  text: string;
+};
+
+const IndoPakWordText = (props: IndoPakWordTextProps) => {
+  const { text } = props;
```
mmahalwy, line 9: prefer to destruct in the function arg :)

src/components/dls/QuranWord/QuranWord.tsx
```diff
@@ -1,60 +1,28 @@
...
-  //   word.highlight ? word.highlight : ''
-  // }`;
-
-  if (useTextFont) {
-    if (word.charType === WORD_TYPES.CHAR_TYPE_END) {
-      text = pad(word.verseKey.split(':')[1], 3, '0');
-    } else if (word.textMadani) {
-      text = word.textMadani;
-    }
+  if (fontStyle === 'uthmani') {
```
mmahalwy, line 17: why not do `if` `else` with return? 

```typescript
if (fontStyle === 'uthmani') {
  return <UthmaniWordText code={word.code} pageNumber={word.pageNumber} />;
}
```

tifa2UP: In the future, the return value will be: 
```
<WordWrapper ...>
	<WordText />
<WordWrapper>
```

WordWrapper will handle the shared functionality across different scripts such as highlighting a word, onClick, etc. 

Lmk if it makes sense!

src/components/dls/QuranWord/IndoPakWordText.tsx
tifa2UP: Is this a strong opinion >.<, this is the pattern that I'm used to but no strong opinions

mmahalwy: not a strong opinion but it's cleaner the other way 😓

src/components/dls/QuranWord/QuranWord.tsx
mmahalwy: ok

PR #76, opened 2020-07-10
Add the Translation and QuranPage view

tifa2UP:
### Summary
Refactored the verse component and added containers for two views:

Translation View: the view to display verses one by one with the translation (main view in v2)
Mushaf View: the reading view from v2

### Notes
- The Mushaf view doesn't break at the end of the Mushaf line, that'll be in a follow-up
- I've opted in for minimal CSS as the design is pending

### Screenshots

Translation view: 
image

Quran Page view: 
image


src/components/QuranReader/MushafView.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
mmahalwy, line 1: prefer to name this file something that's commonly understood. Better named as `PageView` for example. We want to avoid using Arabic words in English as not everyone spells it the same 😅

src/components/QuranReader/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import MushafView from './MushafView';
+import TranslationView from './TranslationView';
+
+type QuranReaderProps = {
+  view: 'translation' | 'reading';
+  verses: VerseType[];
+};
+
+const QuranReader = ({ view, verses }: QuranReaderProps) => {
+  if (view === 'reading') {
```
mmahalwy, line 12: prefer constant for `'reading'` please

src/components/Verse/VerseText.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import styled from 'styled-components';
+import VerseType from '../../../types/VerseType';
+import QuranWord from '../dls/QuranWord/QuranWord';
+
+type VerseTextProps = {
+  verse: VerseType;
+  fontStyle?: 'uthmani' | 'madani' | 'indopak';
```
mmahalwy, line 8: feels like this string literal should be a type somewhere? 
```ts
type FontStyleOptions = 'uthmani' | 'madani' | 'indopak';
```

src/components/QuranReader/MushafView.tsx
tifa2UP: Good called, renamed it QuranPage. Will try to avoid Arabic names moving forward insha'Allah

src/components/Verse/VerseText.tsx
tifa2UP: Yepppp, made it an enum and refactored the code to use it

PR #78, opened 2020-07-10
Save the readingMode in Redux

tifa2UP:
### Summary
Refactored the reading mode preference to live in redux.

### Test Plan
Confirmed that the state is now maintained in redux and persists across sessions

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -1,15 +1,18 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
 import VerseType from '../../../types/VerseType';
 import MushafView from './MushafView';
 import TranslationView from './TranslationView';
+import { selectReadingMode } from '../../redux/slices/QuranReader/readingMode';
 
 type QuranReaderProps = {
-  view: 'translation' | 'reading';
   verses: VerseType[];
 };
 
-const QuranReader = ({ view, verses }: QuranReaderProps) => {
-  if (view === 'reading') {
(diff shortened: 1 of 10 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -1,15 +1,18 @@
...
-  view: 'translation' | 'reading';
   verses: VerseType[];
 };
 
-const QuranReader = ({ view, verses }: QuranReaderProps) => {
-  if (view === 'reading') {
+const QuranReader = ({ verses }: QuranReaderProps) => {
+  const readingMode = useSelector(selectReadingMode);
+
+  if (readingMode === 'mushaf') {
```
mmahalwy, line 15: prefer is `'mushaf'` is a constant please 🙏

src/redux/slices/QuranReader/readingMode.ts
```diff
@@ -0,0 +1,21 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export type ReadingMode = 'translation' | 'mushaf';
+
+const initialState: ReadingMode = 'translation';
```
mmahalwy, line 5: constant please

src/components/QuranReader/index.tsx
tifa2UP: Good call, moved it to an enum.

src/redux/slices/QuranReader/readingMode.ts
tifa2UP: Good call, moved it to an enum.

PR #79, opened 2020-07-11
Add query string object parsing for makeUrl

tifa2UP:
### Summary
Added an optional utility that parses the query string parameters and appends it to the API call. E.g parses `{a: 1, b: 2}` to `?a=1&b=2`.

### Test Plan
Refactored the existing manual queryString usage to use the utility function, the API calls are formatted successfully.


src/utils/api.ts
```diff
@@ -1,2 +1,23 @@
...
+
+  // The following section parses the query params for convenience
+  // E.g. parses {a: 1, b: 2} to "?a=1&b=2"
+  const parametersStringified = JSON.stringify(parameters);
+  const formattedParameters = JSON.parse(
+    parametersStringified,
+    (key, val) => (typeof val !== 'object' && val !== null ? String(val) : val), // Converts the value to a string in each key/value pair
+  );
+
+  const queryParameters = `?${new URLSearchParams(formattedParameters).toString()}`;
```
mmahalwy, line 21: does this work on SSR?

src/utils/api.ts
```diff
@@ -1,2 +1,23 @@
...
+ * @param parameters optional query params, {a: 1, b: 2} is parsed to "?a=1&b=2"
+ */
+export const makeUrl = (path: string, parameters?: Record<string, unknown>) => {
+  if (!parameters) {
+    return `${apiUrl}${path}`;
+  }
+
+  // The following section parses the query params for convenience
+  // E.g. parses {a: 1, b: 2} to "?a=1&b=2"
+  const parametersStringified = JSON.stringify(parameters);
```
mmahalwy, line 15: feels lame but i rather use a lib here like https://www.npmjs.com/package/qs or https://www.npmjs.com/package/query-string. They handle objects, arrays, etc etc

src/utils/api.ts
tifa2UP: Yeah, next had a polyfill for it. It's just string parsing.

src/utils/api.ts
tifa2UP: qs is `10kb`, and query-string is `32kb`. I added qs since we'll likely reuse it in the future. Thank you for the suggestion!

PR #81, opened 2020-07-12
Add useSWRInfinite for paginating requests

tifa2UP:
### Summary
This is part 1 of 2 PRs to build infinite scrolling. `useSWR`'s pages feature is going to be deprecated and replaced with useSWRInfinite in `0.3.0`: https://github.com/vercel/swr/pull/435

`useSWRInfinite` allows us to paginate requests and get all the benefits of using `swr`. This feature is currently in beta, I've tested it and will move us back to the stable `swr` once it goes into the stable release (hopefully in a couple of weeks). 

I've refactored the data fetching logic from `[chapterId]` to `QuranReader` as this is where we'll be managing fetching more data as users scroll. 

### Test Plan
The initial data is fetched and rendered as expected. I've created a setTimeout to fetch more pages every 3 seconds, the data is fetched correctly.

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -1,23 +1,39 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import VerseType from '../../../types/VerseType';
+import ChapterType from 'types/ChapterType';
+import { fetcher } from 'src/api';
+import { useSWRInfinite } from 'swr';
+import { makeUrl } from 'src/utils/api';
+import VerseType from 'types/VerseType';
 import QuranPageView from './QuranPageView';
 import TranslationView from './TranslationView';
 import { selectReadingView } from '../../redux/slices/QuranReader/readingView';
 import { ReadingView } from './types';
 
 type QuranReaderProps = {
-  verses: VerseType[];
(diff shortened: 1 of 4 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -1,23 +1,39 @@
...
-  verses: VerseType[];
+  initialData: { verses: VerseType[] };
+  chapter: ChapterType;
 };
 
-const QuranReader = ({ verses }: QuranReaderProps) => {
+const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
+  const { data } = useSWRInfinite(
+    (index) => {
+      return makeUrl(`/chapters/${chapter.id}/verses`, { translations: 20, page: index }); // TODO: select the translation using the user preference
```
tifa2UP, line 21: In the follow-up PR, `index` will represent that API page we're fetching using the `size` and `setSize` methods from useSWRInfinite

mmahalwy:
@tifa2UP I was testing our `SWR` cause its look. That being said, we may not want to use it actually since we will make requests via redux actions and store it in redux store. That is, unless we can make SWR and redux play nicely together. WDYT?

PR #82, opened 2020-07-13
Implement infinite scrolling 🎉

tifa2UP:
### Summary
Added infinite scrolling and a few refactors: 

- Used `react-simple-infinite-scroll` to implement virtual scrolling*
- Created a custom `verseFetcher` to extract the verses array from the request

Changes that should be in a separate PR:
- Refactored a couple of props destructuring to the function parameters. cc: @mmahalwy ;)
- Removed the API responses into a standalone ts declarations file.
- Replaced `api.quran.com` with `quran.com/api/api` as former doesn't have an SSL certificate and the API requests are blocked on staging. Naveed is working on issues a certificate.

### React-simple-inifinite-scroll vs react-window*
@mmahalwy and @nabeel- I know that we had a discussion about opting in for simplicity and not using windowing unless we need to. After this implementation, I have a stronger opinion for having a follow-up PR to replace the infinite scroller with react-window for a few reasons: 

- The package size for react-window is smaller `24.2kb` vs `45.76kb`
- With long surah like Albaqarah, the FPS rate drops to `30fps` while scrolling, this is while our app is still bare minimum.
- The scrollbar wiggles and shifts up with every new API request
- React-simple-inifinite-scroll doesn't do `onScroll` handler cleanup, it attaches a new event listener every time you visit a new chapter page. 

### Test Plan
- Tested infinite scrolling in Surah Albaqarah as well as short surah - the ayahs load as expected
- Tested the FPS rate, it drops to 20-30 after the DOM starts getting big (e.g. surah albaqarah)
- Tested if the event listeners are being detached - they're not. Will create a follow-up to replace the library
- Tested if an API endpoint is being made out of bounds: no API calls are triggered at the end of the surah and no invalid API requests are made.

### Screenshots
Opted for a video to preserve the frame rate: https://drive.google.com/file/d/1OjD7MzaLn4RAV5ovHOrT9PHB-4vVjrf_/view?usp=sharing

You can try it on the staging link :)


types/APIResponses.ts
```diff
@@ -0,0 +1,24 @@
+import VerseType from './VerseType';
```
mmahalwy, line 1: i dig this file 🙏

PR #84, opened 2020-07-14
Replace the infinite scrolling library

tifa2UP:
### Summary
Replaced the `react-simple-infinite-scroll` library with `react-infinite-scroller`. Results: 

- `50-60` fps during scrolling, up from `20-30`fps
- Removes React 15 from the bundle
- `25kb` library size, down from `45kb`

### Notes 
`react-window` brings a solid 60 fps but generates an empty DOM for SSR. The amount of data in the dom is managed by the window size of the client, for crawlers that's zero and it doesn't render any data :/. 

Using the new library brings us the best of both worlds, simplicity and a good fps.

### Test Plan
Same as #82 

### Screenshots
Feel free to try scrolling on staging :) 

image


mmahalwy:
@tifa2UP this is awesome. TY for doing this. One thing I noticed is the API is REALLY slow. ~1.5s to load the next set of verses. @naveedahmad I wonder why?

tifa2UP:
@mmahalwy, yes! @naveed-ahmad one thing I've noticed is that switching the endpoint from `api.quran.com` to `quran.com/api/api` had a noticeable increase in response times, I didn't benchmark it but I'm happy to profile it.

PR #89, opened 2020-07-30
Align lines in the Quran Reader Page View

tifa2UP:
### Summary
Implemented the Page View (formerly known as "Reading Mode", and "Mushaf view") such that the lines match the lines in the Quran.

- Grouped the verses into "lines" based on the line property from the api.
- Set a `vw` dimensions that activates on smaller devices so that we retain the alignment on mobile screens. 
- Refactored the Page and Translation view into separate folders. 

### Test Plan
Scrolled through surat Al-baqarah on a desktop and mobile viewport. The layout is maintained in both

### Screenshots

| Before (Desktop v4) | After (Desktop v4) |
| ------ | ------ |
|image|image|

| Mobile (v2) | Mobile (v4) |
| ------ | ------ |
|image|image|


tifa2UP:
@mmahalwy merging this. Feel free to leave any comments and I'll create a follow-up PR

PR #98, opened 2020-09-12
Fixes for eslint property re-assign error, and other warnings

mahmoudfelfel:
### Summary
- Redux toolkit uses [immer](https://github.com/immerjs/immer/) to allow mutating the state inside the reducers, that conflicts with the `no-param-reassign` eslint rule, instead of turning it off, I added an ignore rule for the `draft` object that is usually passed to the reducer to mutate its property, which later becomes the new state.
- Fixed a few other minor warnings.

### Screenshots

| Before | After |
| ------ | ------ |

### DIFF
--- a/src/redux/slices/QuranReader/readingView.ts
+++ b/src/redux/slices/QuranReader/readingView.ts
@@ -7,9 +7,7 @@ export const readingViewSlice = createSlice({
   name: 'readingView',
   initialState,
   reducers: {
-    setReadingView: (state: ReadingView, action: PayloadAction<ReadingView>) => {
-      state = action.payload;
-    },
+    setReadingView: (draft, action: PayloadAction) => action.payload,
   },
 });
 
--- a/types/ChapterType.ts
+++ b/types/ChapterType.ts
@@ -9,7 +9,7 @@ interface ChapterType {
   nameSimple: string;
   nameArabic: string;
   chapterNumber: number;
-  translatedName: string;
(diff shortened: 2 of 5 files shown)


src/redux/slices/QuranReader/readingView.ts
```diff
@@ -7,9 +7,7 @@ export const readingViewSlice = createSlice({
   name: 'readingView',
   initialState,
   reducers: {
-    setReadingView: (state: ReadingView, action: PayloadAction<ReadingView>) => {
-      state = action.payload;
-    },
+    setReadingView: (draft, action: PayloadAction) => action.payload,
```
tifa2UP, line 10: Why are we doing this change?

types/ChapterType.ts
```diff
@@ -9,8 +9,8 @@ interface ChapterType {
   nameSimple: string;
   nameArabic: string;
   chapterNumber: number;
-  translatedName: any;
-  languageName: any;
+  translatedName: string;
+  languageName: string;
```
tifa2UP, line 13: Thank you!

src/redux/slices/QuranReader/readingView.ts
mahmoudfelfel: you don't need to do `state = `, if you are resetting the whole state, returning the new value is enough.
That will cause the re-assign to fail even with my change in place, as my change is only to allow properties re-assigning, not the entire parameter.

tifa2UP (approved):

Issue #6, opened 2020-05-11
Create mobile app page


mmahalwy:
Designs here: http://beta.quran.com/pages/apps

Shemsedin:
@mmahalwy  I will take this in sha Allah.

Shemsedin:
@mmahalwy I have done a PR please see https://github.com/quran/quran.com-frontend-next/pull/100

tifa2UP:
Chatted with @Shemsedin on his PR. This issue is also blocked on the design. I'm gonna close it until we have the new specs

PR #100, opened 2020-10-04
Feature/mobile app page

Shemsedin:
### Summary
This is about creating of Create mobile app page (issue#6) as per this [link](https://github.com/quran/quran.com-frontend-next/issues/6) 

### Screenshots

| Before | After |
| ------ | ------ |


tifa2UP:
Assalamu alaikum Shamsedin, JAK for creating the PR. it looks like the build is failing. Design is working on an overhaul for the design system and main pages (including the mobile apps). The goal is to have quran-next designed for an overall better experience and not just be a re-skin over the existing website. 

Would you mind if I keep the PR open until we get the specs from Design? This way it'd be easier to refactor them one at a time. Will update the Github issues accordingly

.idea/$CACHE_FILE$
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
abdullahceylan, line 1: I think you shouldn't send `.idea` folder as it's for your IDE. So can you add it to the `.gitignore`?

src/components/Pages/QuranApps/AppsPage.stories.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
abdullahceylan, line 1: Within React 17 and newer versions come with the new JSX transforms and you don't need to import `React` anymore.

You can update the other files in this PR as well.

.idea/$CACHE_FILE$
Shemsedin: @abdullahceylan  I have added this to gitignore

src/components/Pages/QuranApps/AppsPage.stories.tsx
Shemsedin: I tried to remove this but it `yarn lint:` through an error and then doesn't let me commit hence can't remove the import.

abdullahceylan: Salaam @Shemsedin,

If you add `react/jsx-uses-react: 0` right after line 47 in `.eslintrc.yml` you won't get an error anymore.

Shemsedin: I tried this but it is the same. Maybe if you tray on your end what works (on this branch) and let me know and I am happy to change that part.

PR #104, opened 2020-10-31
Replace <img /> usage with Next 10 <Image />

tifa2UP:
### Summary
Replaced the `` usage with the new Image component from Next.js 10. Removed the libraries we had for optimizing images since it's now handled automatically by Next 10.


mmahalwy (approved):

tifa2UP:
FYI, next-image doesn't have storybook support yet. It's planned to be fixed in nextjs `10.0.3` I'm going to merge to not block and create a follow-up once `10.0.3`is released

https://github.com/vercel/next.js/commit/5562daf7a160e62fd4f78eca257ba957ad64074c

PR #106, opened 2020-11-01
Fix no-document-viewport-meta and missing-document-component warnings

osamasayed:
### Summary
Next.js was showing two warnings regarding [no-document-viewport-meta](https://err.sh/next.js/no-document-viewport-meta ) and [missing-document-component](https://github.com/vercel/next.js/blob/master/errors/missing-document-component.md) when starting the dev server.

### Screenshots

| Before | After |
| ------ | ------ |
| Screen Shot 2020-11-01 at 19 17 46Screen Shot 2020-11-01 at 19 21 47|Screen Shot 2020-11-01 at 20 08 11|

### DIFF
--- a/src/pages_/_app.tsx
+++ b/src/pages_/_app.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import Head from 'next/head';
 import { DefaultSeo } from 'next-seo';
 import { ThemeProvider } from 'styled-components';
 import { theme } from '../utils/styles';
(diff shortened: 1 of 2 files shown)


src/pages_/_app.tsx
```diff
@@ -8,13 +9,18 @@ import ReduxProvider from '../redux/Provider';
...
   return (
-    <ReduxProvider>
-      <ThemeProvider theme={theme}>
-        <DefaultSeo {...createSEOConfig({})} />
-        <Component {...pageProps} />
-        <Footer />
-      </ThemeProvider>
-    </ReduxProvider>
+    <>
+      <Head>
```
abdullahceylan, line 13: `_app.tsx` shouldn't contain this as it's for initializing the pages.

src/pages_/_document.tsx
```diff
@@ -32,9 +32,8 @@ export default class MyDocument extends Document {
     const { styleTags } = this.props;
 
     return (
-      <html lang="en">
+      <Html lang="en">
```
abdullahceylan, line 35: Just a quick note: I think we should handle this dynamically based on the locale rather than using a static value.

PR #139, opened 2021-02-04
Remove extraneous word-spacing in uthmani text 

tifa2UP:
### Summary
Leftover from debugging the missing alif on page 307 in Firefox. Removed to reset the page layout

### Test Plan
The page now renders as before.


src/components/Verse/VerseText.tsx
```diff
@@ -2,18 +2,29 @@ import React from 'react';
...
+  const isUthmaniText = quranReaderStyles.quranFont === QuranFont.Uthmani;
+  const isFirstTwoPages = words[0].pageNumber <= 2; // We don't want to justify the text on the first 2 pages to resemble the madani mushaf
 
   return (
-    <StyledVerseTextContainer styles={quranReaderStyles}>
+    <StyledVerseTextContainer
+      styles={quranReaderStyles}
+      isQuranPage={isQuranPage}
+      isUthmaniText={isUthmaniText}
+      isFirstTwoPages={isFirstTwoPages}
```
naveed-ahmad: ill suggest using `alightCenter` instead of `isFirstTwoPages`. Then we can reuse this if we need to center align other surah(/113, /114) in future.

PR #146, opened 2021-02-13
Fix Typos + Replace Redux with xState for AudioPlayer

osamasayed:
### Summary

While Redux is a great library but in some cases it becomes quite complicated to visualize and cover all the possible states and sync the UI with the logic behind those states. An audio player is one of those scenarios as an audio player can be at only one state at a time, either it’s Loading, Playing, Paused, Ended playing or Failed loading. [Xstate](https://xstate.js.org/docs/) is a much more robust library to handle this case than Redux in my humble opinion. It takes advantage of state diagrams to help us visualize all the possible states, the transitions between them and what triggers state A to move to state B. In this commit I have came up with the below state diagram for the audio player and tied it with the UI. I have also added a couple of improvements and fixed a typo. This is an initial commit for the audio player and we can take it from there and improve it. I didn’t add any loading indicators because that will depend on the UX and didn’t want to add a loading indicator that is not compatible with the design. Please feel free to let me know your thoughts and any issues.

Screen Shot 2021-06-14 at 06 36 19


naveed-ahmad (approved):

osamasayed:
Alsalam Alikum @naveed-ahmad I have added one more commit to the current pull request to use Xstate for the audioPlayer instead of Redux. Please have a look and let me know your thoughts.

PR #167, opened 2021-05-24
Version 1 of Audio Implementation. Functionality only

tifa2UP:
Added functionality but no UI to the audio player


src/components/AudioPlayer/Slider.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
tifa2UP, line 1: place holder for v2

naveed-ahmad (approved):

PR #186, opened 2021-07-01
Qdc api

naveed-ahmad:
### Screenshots

| Before | After |
| ------ | ------ |


types/SlugType.ts
```diff
@@ -0,0 +1,6 @@
+interface SlugType {
```
tifa2UP, line 1: Drop "Type". Just `Slug`

types/TranslatedNameType.ts
```diff
@@ -0,0 +1,6 @@
+interface TranslatedNameType {
```
tifa2UP, line 1: ditto, just `TranslatedName`. all interfaces are types :D

types/ChapterType.ts
```diff
@@ -9,8 +12,9 @@ interface ChapterType {
   nameSimple: string;
   nameArabic: string;
   chapterNumber: number;
-  translatedName: Record<'name' | 'languageName', string>;
-  languageName: string;
+  translatedName: TranslatedNameType;
+  defaultSlug: SlugType;
+  slugs?: Array<SlugType>;
```
tifa2UP, line 17: No need to change, but for future reference, you can do `Slug[]` if it's easier

tifa2UP (dismissed):
💯

PR #194, opened 2021-07-15
Add Incremental Static Generation for chapter pages

osamasayed:
### Summary
The current behavior is that all chapters will be generated at build time. Not only each chapter but each language, and translation as well for each chapter which makes the building process slow. In this PR, we moved away from using SSG and added ISG instead. This will work as follows:

1. The build process will not pre-render any of the chapters's pages.
2. When a page is first requested, a static version of the page will be generated after calling the BE API and then returned to the browser. It will also be cached for any subsequent requests for the same page until the cache validation time expires which is set to 7 days since the data doesn't get updated that frequent.
3. Once the validation time runs out (after 7 days), the first request after that will still be served from the cache (the stale page) and in the background the ISG will trigger a re-generation of the page and in-case it's successful, it will invalidate the old cached version and replace it with the new version so that any subsequent requests to the same page will be served the new cached version. 

### Screenshots

| Before | After |
| ------ | ------ |
| [SSG] | [ISG] |


src/pages/[chapterId].tsx
```diff
@@ -18,31 +18,37 @@ const Chapter: NextPage<ChapterProps> = ({ chapterResponse: { chapter }, versesR
 
 export const getStaticProps: GetStaticProps = async ({ params }) => {
   const chapterId = String(params.chapterId);
-
-  const [chaptersResponse, chapterResponse, versesResponse] = await Promise.all([
-    getChapters(),
+  // we need to validate the chapterId first to save calling BE since we haven't set the valid paths inside getStaticPaths to avoid pre-rendering them at build time.
+  if (!isValidChapterId(chapterId)) {
+    return {
+      notFound: true,
```
Comment (tifa2UP) [inline src/pages/[chapterId].tsx:24]: nice

src/pages/[chapterId].tsx
```diff
@@ -18,31 +18,37 @@ const Chapter: NextPage<ChapterProps> = ({ chapterResponse: { chapter }, versesR
...
+  const [chapterResponse, versesResponse] = await Promise.all([
     getChapter(chapterId),
     getChapterVerses(chapterId, {
       wordFields: `verse_key, verse_id, page_number, location, ${QuranFont.QPCHafs}`,
     }),
   ]);
+  // if any of the APIs have failed due to internal server error, we will still receive a response but the body will be something like {"status":500,"error":"Internal Server Error"}.
+  if (chapterResponse.status === 500 || versesResponse.status === 500) {
+    return {
+      notFound: true,
```
Comment (tifa2UP) [inline src/pages/[chapterId].tsx:36]: Is there an alternate approach that we can follow when the backend is not responding? 

occasionally the backend might not give us a response, and a not found might be detrimental to SEO

Comment (osamasayed) [reply src/pages/[chapterId].tsx:36]: As per our discussion, in-case we have an error, we will return an error UI instead of the `QuranReader` and revalidate that after 35 seconds which will trigger the regeneration of the page again.

tifa2UP (changes requested):
Code lgtm, thank you! Good to go once the storybook error is resolved

PR #195, opened 2021-07-15
add 'rtl' attribute when setting the language to Arabic

osamasayed:
### Summary

We are using `next-translate` to add localization. When switching to Arabic, we need to let the browser know that the text direction of the element's content is `rtl` this is done by setting the `dir` attribute on `main` element like this: 
`<main dir='rtl'>`

### Test Plan

1. Visit `{url}/ar/about` and we should find the content shown correctly from right to left since we are changing the language to Arabic.
2. Visit `{url}/about` and we should find the content shown correctly from left to right.

### Screenshots

| Before | After |
| ------ | ------ |
| Screen Shot 2021-07-15 at 09 57 24 | Screen Shot 2021-07-15 at 09 56 49 |

### DIFF
--- a/src/components/RTLLayoutDetector.tsx
+++ b/src/components/RTLLayoutDetector.tsx
@@ -5,10 +5,10 @@ interface Props {
   children: ReactNode | ReactNode[];
 }
 
-const Layout: React.FC<Props> = ({ children }) => {
+const RTLLayoutDetector: React.FC<Props> = ({ children }) => {
   const { lang } = useTranslation();
 
   return <main dir={lang === 'ar' ? 'rtl' : 'ltr'}>{children}</main>;
 };
 
-export default Layout;
+export default RTLLayoutDetector;
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -1,0 +1,14 @@
+import { ReactNode } from 'react';
+import useTranslation from 'next-translate/useTranslation';
+
+interface Props {
+  children: ReactNode | ReactNode[];
+}
+
+const Layout: React.FC<Props> = ({ children }) => {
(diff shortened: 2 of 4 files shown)


tifa2UP (changes requested):
I like this approach, I have two suggestions:

1) can we move this logic to `_app.tsx` instead of having to duplicate using the layout component in every page.

2) should we rename the Layout component to RTLLayoutDetector or similar?

osamasayed:
> I like this approach, I have two suggestions:
> 
> 1. can we move the component usage to `_app.tsx` instead of having to duplicate using the layout component in every page.
> 2. should we rename the Layout component to RTLLayoutDetector or similar?

src/components/RTLLayoutDetector.tsx
```diff
@@ -0,0 +1,14 @@
...
+import useTranslation from 'next-translate/useTranslation';
+
+interface Props {
+  children: ReactNode | ReactNode[];
+}
+
+const RTLLayoutDetector: React.FC<Props> = ({ children }) => {
+  const { lang } = useTranslation();
+
+  return <main dir={lang === 'ar' ? 'rtl' : 'ltr'}>{children}</main>;
```
naveed-ahmad, line 11: can we also include:
- Uyghur
- Urdu
- Persian
- Hebrew
- Pashto
- Divehi 

all of these are rtl languages.

osamasayed: Got it thanks @naveed-ahmad . As per my discussion with Abdellatif, we will postpone this PR until pre-production release so that all pages, layouts and components are in place so that we can test this feature properly and make sure nothing breaks. but I will include those languages once I am back to this PR. Thanks a lot.

PR #197, opened 2021-07-17
add ability to select/change translations

osamasayed:
### Summary
This PR adds the ability to select/un-select multiple translations for the QuranReader.

### Test Plan

1. Open the Developer Utility.
2. `Show Translations` button should show.
3. Click on `Show Translations` button which should show a multiple `select` element that has the list of available translations.
4. Select the first 2 translations and they should appear below each other in the QuranReader.
5. Refresh the page, the 2 translations selected in step 4 should still show.
6. Open the Developer Utility again and click on `Show Translations` button again and the 2 translations selected in step 4 should still be selected.

### Screenshots

| Before | After |
| ------ | ------ |
| Before | After 1 |
|  | After 2 |
|  | After 3 |
|  | After 4 |

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,8 +1,9 @@
 import { camelizeKeys } from 'humps';
+import { TranslationsResponse } from 'types/APIResponses';
 import { makeUrl } from './utils/api';
 import Chapter from '../types/Chapter';
 import Verse from '../types/Verse';
-import { makeVersesUrl } from './utils/apiPaths';
+import { makeTranslationsUrl, makeVersesUrl } from './utils/apiPaths';
 
 export const fetcher = async function fetcher(input: RequestInfo, init?: RequestInit) {
   const res = await fetch(input, init);
@@ -47,4 +48,17 @@ export const getChapterVersesResponse = async (
   params?: Record<string, unknown>,
 ) => {
   return fetcher(makeVersesUrl(id, params));
+};
+
+/**
+ * Get the current available translations with the name translated in the current language.
+ *
+ * @param {string} language we use this to get translated names of authors in specific the current language.
+ *
+ * @returns {Promise<TranslationsResponse>}
+ */
+export const getAvailableTranslations = async (language: string): Promise<TranslationsResponse> => {
+  const payload = await fetcher(makeTranslationsUrl(language));
+
+  return camelizeKeys(payload) as TranslationsResponse;
 };
--- a/src/components/DeveloperUtility/TranslationAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationAdjustment.tsx
@@ -1,0 +1,81 @@
+import { useEffect, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import styled from 'styled-components';
+import { getAvailableTranslations } from 'src/api';
+import { TranslationsResponse } from 'types/APIResponses';
+import TranslationResource from 'types/TranslationResource';
+import useTranslation from 'next-translate/useTranslation';
+import {
+  selectCurrentTranslations,
+  setCurrentTranslations,
+} from 'src/redux/slices/QuranReader/translations';
+
+const StyledSelect = styled.select.attrs({
+  name: 'translations',
+  multiple: true,
+})`
+  background-color: white;
+  width: 100%;
+  height: 220px;
+`;
+
+const TranslationAdjustment = () => {
+  const dispatch = useDispatch();
+  const currentTranslations = useSelector(selectCurrentTranslations) as string[];
+  const { lang } = useTranslation();
+  const [translations, setTranslations] = useState<TranslationResource[]>([]);
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  useEffect(() => {
+    getAvailableTranslations(lang)
+      .then((res: TranslationsResponse) => {
+        // if there is an internal server error.
+        if (res.status === 500) {
+          setHasError(true);
+        } else {
+          setTranslations(res.translations);
+        }
+      })
+      .catch(() => {
+        setHasError(true);
+      });
+  }, [lang]);
+
+  if (hasError) {
+    return null;
+  }
+
+  /**
+   * Handle when and item(s) get selected or un-selected.
+   *
+   * @param {React.ChangeEvent<HTMLSelectElement>} e
+   */
+  const onSelectedChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    // extract the selected translation IDs as an arrays of strings
+    const selectedTranslations = Array.from(
+      e.target.selectedOptions,
+      (option) => option.value,
+    ) as string[];
+    dispatch({ type: setCurrentTranslations.type, payload: selectedTranslations });
+  };
+
+  return (
+    <div>
+      <button type="button" onClick={() => setIsExpanded((prevIsExpanded) => !prevIsExpanded)}>
+        <p>{isExpanded ? 'Hide' : 'Show'} Translations</p>
+      </button>
+      {isExpanded && (
+        <StyledSelect onChange={onSelectedChange} defaultValue={currentTranslations}>
+          {translations.map((translation) => (
+            <option key={translation.id} value={translation.id}>
+              {translation.translatedName.name}
+            </option>
+          ))}
+        </StyledSelect>
+      )}
+    </div>
+  );
+};
+
+export default TranslationAdjustment;
--- a/src/components/DeveloperUtility/TranslationsAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationsAdjustment.tsx
@@ -2,18 +2,27 @@ import { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import styled from 'styled-components';
 import { getAvailableTranslations } from 'src/api';
-import TranslationResource from 'types/TranslationResource';
+import AvailableTranslation from 'types/AvailableTranslation';
 import useTranslation from 'next-translate/useTranslation';
 import {
   selectTranslations,
   setSelectedTranslations,
 } from 'src/redux/slices/QuranReader/translations';
 
+/**
+ * Convert an array of numbers to an array of strings.
+ *
+ * @param {number[]} numbersArray
+ * @returns {string[]}
+ */
+const numbersToStringsArray = (numbersArray: number[]): string[] =>
+  numbersArray.map((number) => String(number));
+
 const TranslationsAdjustment = () => {
   const dispatch = useDispatch();
   const selectedTranslations = useSelector(selectTranslations) as number[];
   const { lang } = useTranslation();
-  const [translations, setTranslations] = useState<TranslationResource[]>([]);
+  const [translations, setTranslations] = useState<AvailableTranslation[]>([]);
   const [isExpanded, setIsExpanded] = useState(false);
   const [hasError, setHasError] = useState(false);
 
@@ -56,7 +65,10 @@ const TranslationsAdjustment = () => {
         <p>{isExpanded ? 'Hide' : 'Show'} Translations</p>
       </button>
       {isExpanded && (
-        <StyledSelect onChange={onSelectedTranslationsChange} defaultValue={selectedTranslations}>
+        <StyledSelect
+          onChange={onSelectedTranslationsChange}
+          defaultValue={numbersToStringsArray(selectedTranslations)}
+        >
           {translations.map((translation) => (
             <option key={translation.id} value={translation.id}>
               {translation.translatedName.name}
@@ -74,7 +86,7 @@ const StyledSelect = styled.select.attrs({
 })`
   background-color: white;
   width: 100%;
-  height: ${(props) => props.theme.heights.jumbo};
+  height: calc(5 * ${(props) => props.theme.spacing.mega});
 `;
 
 export default TranslationsAdjustment;
--- a/src/components/DeveloperUtility/TranslationsAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationsAdjustment.tsx
@@ -2,26 +2,16 @@ import { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import styled from 'styled-components';
 import { getAvailableTranslations } from 'src/api';
-import { TranslationsResponse } from 'types/APIResponses';
 import TranslationResource from 'types/TranslationResource';
 import useTranslation from 'next-translate/useTranslation';
 import {
-  selectCurrentTranslations,
-  setCurrentTranslations,
+  selectTranslations,
+  setSelectedTranslations,
 } from 'src/redux/slices/QuranReader/translations';
 
-const StyledSelect = styled.select.attrs({
-  name: 'translations',
-  multiple: true,
-})`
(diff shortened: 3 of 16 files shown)


src/api.ts
```diff
@@ -48,3 +49,16 @@ export const getChapterVersesResponse = async (
...
 };
+
+/**
+ * Get the current available translations with the name translated in the current language.
+ *
+ * @param {string} language we use this to get translated names of authors in specific the current language.
+ *
+ * @returns {Promise<TranslationsResponse>}
+ */
+export const getAvailableTranslations = async (language: string): Promise<TranslationsResponse> => {
```
naveed-ahmad, line 60: `getAvailableTranslations` is bit confusing.  Why not just `getTranslations`?

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const { lang } = useTranslation();
+  const [translations, setTranslations] = useState<TranslationResource[]>([]);
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  useEffect(() => {
+    getAvailableTranslations(lang)
+      .then((res: TranslationsResponse) => {
+        // if there is an internal server error.
+        if (res.status === 500) {
```
naveed-ahmad, line 34: I would handle this error in `getAvailableTranslations` and just return empty array if there is an error.

src/components/QuranReader/index.tsx
```diff
@@ -38,18 +40,19 @@ const verseFetcher = async (input: RequestInfo, init?: RequestInit) => {
...
         wordFields: `verse_key, verse_id, page_number, location, ${quranReaderStyles.quranFont}`,
-      });
-    },
+        translations: joinStringArray(currentTranslations),
+      }),
     verseFetcher,
     {
       initialData: initialData.verses,
-      revalidateOnFocus: true,
+      revalidateOnFocus: false, // disable auto revalidation when window gets focused
```
naveed-ahmad, line 54: Thank you for disabling this, so this was generating two request :) 

While adding v2 fonts, page won't render on initial page load. But when I click on dev utility, it was triggering the api request and rendering the page.

src/redux/slices/QuranReader/translations.ts
```diff
@@ -0,0 +1,22 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+const DEFAULT_TRANSLATIONS = ['20'];
```
naveed-ahmad, line 3: We've two default translations of English. 

20: Shaih International 
131:  Clear Quran

src/redux/store.ts
```diff
@@ -22,7 +23,7 @@ const persistConfig = {
   key: 'root',
   version: 1,
   storage,
-  whitelist: ['quranReaderStyles', 'readingView'], // Reducers defined here will be have their values saved in local storage and persist across sessions. See: https://github.com/rt2zz/redux-persist#blacklist--whitelist
+  whitelist: ['quranReaderStyles', 'readingView', 'translations'], // Reducers defined here will be have their values saved in local storage and persist across sessions. See: https://github.com/rt2zz/redux-persist#blacklist--whitelist
```
tifa2UP, line 26: nice

src/redux/slices/QuranReader/translations.ts
tifa2UP: I like that you used the ids instead of the names for the translations. It's more robust.

That should be a `number` not a `string` to match the api, right? You'll have to refactor it in a couple of places

src/redux/slices/QuranReader/translations.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const initialState: string[] = DEFAULT_TRANSLATIONS;
```
tifa2UP, line 5: ditto `number[]`

src/redux/slices/QuranReader/translations.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const translationsSlice = createSlice({
+  name: 'translations',
+  initialState,
+  reducers: {
+    setCurrentTranslations: (state, action: PayloadAction<string[]>) => {
```
tifa2UP, line 11: nit: is `selectedTranslations` better that `currentTranslations`?

src/redux/slices/QuranReader/translations.ts
```diff
@@ -0,0 +1,22 @@
...
+      // this is to handle if the user un-selects all translations, we will reset it to defaults.
+      return action.payload.length ? action.payload : DEFAULT_TRANSLATIONS;
```
tifa2UP, line 13: I do not think that we should do this, if the user unselects all translations, we should show nothing, similar to the current experience on the live Quran.com

There are two reasons:
1) the experience will feel buggy, if a user unchecks something and then it is selected all of suddent
2) there are legitimate use cases for people not wanting to have a translation in the translation view. (e.g. a user primarily uses QDC in arabic, and uses the translation view to play audio of specific verses)

src/components/QuranReader/index.tsx
```diff
@@ -38,18 +40,19 @@ const verseFetcher = async (input: RequestInfo, init?: RequestInit) => {
 
 const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
   const quranReaderStyles = useSelector(selectQuranReaderStyles);
+  const currentTranslations = useSelector(selectCurrentTranslations) as string[];
```
tifa2UP, line 43: why do you need the `as string[]` part? It should be done automatically by typescript, no?

src/components/QuranReader/index.tsx
```diff
@@ -38,18 +40,19 @@ const verseFetcher = async (input: RequestInfo, init?: RequestInit) => {
...
-      });
-    },
+        translations: joinStringArray(currentTranslations),
+      }),
     verseFetcher,
     {
       initialData: initialData.verses,
-      revalidateOnFocus: true,
+      revalidateOnFocus: false, // disable auto revalidation when window gets focused
+      revalidateOnMount: true, // enable automatic revalidation when component is mounted. This is needed when the translations inside initialData don't match with the user preferences and would result in inconsistency either when we first load the QuranReader with pre-saved translations from the persistent store or when we change the translations' preferences after initial load.
```
tifa2UP, line 55: does revalidate only rerenders if the data is different? 

If not, we'll need our own custom logic to prevent do the revalidation conditionally. Instead of setting it to always be `true`

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+const StyledSelect = styled.select.attrs({
+  name: 'translations',
+  multiple: true,
+})`
+  background-color: white;
+  width: 100%;
+  height: 220px;
+`;
+
+const TranslationAdjustment = () => {
```
tifa2UP, line 22: nit: `TranslationsAdjustment`

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+        setHasError(true);
+      });
+  }, [lang]);
+
+  if (hasError) {
+    return null;
+  }
+
+  /**
+   * Handle when and item(s) get selected or un-selected.
```
tifa2UP, line 50: nit: an

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @param {React.ChangeEvent<HTMLSelectElement>} e
+   */
+  const onSelectedChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
```
tifa2UP, line 54: nit: `onSelectedTranslationsChange`

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+    // extract the selected translation IDs as an arrays of strings
+    const selectedTranslations = Array.from(
+      e.target.selectedOptions,
+      (option) => option.value,
+    ) as string[];
```
tifa2UP, line 59: we should use `TranslationResource.id[]` instead of a generic `string[]`. 

This is generally a better practice and is more readable

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+    dispatch({ type: setCurrentTranslations.type, payload: selectedTranslations });
+  };
+
+  return (
+    <div>
+      <button type="button" onClick={() => setIsExpanded((prevIsExpanded) => !prevIsExpanded)}>
+        <p>{isExpanded ? 'Hide' : 'Show'} Translations</p>
```
tifa2UP, line 66: nit: add a `selected translations` text on the left of the button to match the other developer utility items

src/components/QuranReader/index.tsx
```diff
@@ -38,18 +40,19 @@ const verseFetcher = async (input: RequestInfo, init?: RequestInit) => {
 
 const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
```
tifa2UP, line 41: while changing translations, there's a brief moment where the default translation is shown, regardless of which translations are currently selected.

Is there a way to either: 
1) show the current translations until the new ones load
2) show a loading indicator instead of the default translations. This loading indicator can just be `loading...` text for now

src/api.ts
osamasayed: Maybe I can rename it to `getTranslationsResources` since it will match the API path `/resources/translations` and also we can include "resource" terminology as a distinguisher between a translation and a translation of a verse. In my commit, I already introduced a type called `TranslationResource` that represents one item inside the array of translations. What do you think @naveed-ahmad @tifa2UP ?

src/utils/string.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Join an array of string items by a comma.
+ *
+ * @param {string[]} stringsArray  e.g. ["1", "2", "3"]
+ * @returns  {string} e.g. "1, 2, 3"
+ */
+export const joinStringArray = (stringsArray: string[]): string => stringsArray.join(', ');
```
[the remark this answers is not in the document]
osamasayed: I know it's a one-liner, but I like to separate concerns 😃 I can remove it if you want me to.

src/redux/slices/QuranReader/translations.ts
osamasayed: > We've two default translations of English.
> 
> 20: Shaih International
> 131: Clear Quran

Okay got it, I will include 131 as well. The reason I included 20 is because the default value of the translations set inside`DEFAULT_VERSES_PARAMS` was set to 20 only.

osamasayed: > I like that you used the ids instead of the names for the translations. It's more robust.
> 
> That should be a `number` not a `string` to match the api, right? You'll have to refactor it in a couple of places

Okay got it. The reason I have it as a `string` is because when the selected option(s) changes, the type of the value inside each `option.value` inside `e.target.selectedOptions` is `string`. But no problem I can convert it to Number before setting it in the store.

src/components/DeveloperUtility/TranslationAdjustment.tsx
```diff
@@ -0,0 +1,81 @@
...
+} from 'src/redux/slices/QuranReader/translations';
+
+const StyledSelect = styled.select.attrs({
+  name: 'translations',
+  multiple: true,
+})`
+  background-color: white;
+  width: 100%;
+  height: 220px;
+`;
```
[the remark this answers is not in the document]
osamasayed: Okay got it.

src/components/QuranReader/index.tsx
osamasayed: No it doesn't resolve it automatically, it shows `any` as the type.

src/components/DeveloperUtility/TranslationsAdjustment.tsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  return (
+    <div>
+      Selected translations{' '}
+      <button type="button" onClick={() => setIsExpanded((prevIsExpanded) => !prevIsExpanded)}>
+        <p>{isExpanded ? 'Hide' : 'Show'} Translations</p>
+      </button>
+      {isExpanded && (
+        <StyledSelect onChange={onSelectedTranslationsChange} defaultValue={selectedTranslations}>
```
tifa2UP, line 59: use a helper function below to convert the numbers array to a string array

src/api.ts
tifa2UP: Chatted with @osamasayed, availableTranslations seems to be the clearest option. Let's go with it for now.

src/components/DeveloperUtility/TranslationsAdjustment.tsx
```diff
@@ -0,0 +1,80 @@
+import { useEffect, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import styled from 'styled-components';
+import { getAvailableTranslations } from 'src/api';
+import TranslationResource from 'types/TranslationResource';
```
tifa2UP, line 5: change to AvailableTranslation

src/components/QuranReader/TranslationView/index.tsx
```diff
@@ -36,8 +39,9 @@ const StyledTranslationView = styled.div`
   margin: ${(props) => props.theme.spacing.medium} auto;
 `;
 
-const StyledText = styled.div`
+const StyledText = styled.div<{ styles: QuranReaderStyles }>`
```
tifa2UP, line 42: Make line-height "unset"

src/components/QuranReader/TranslationView/index.tsx
```diff
@@ -1,21 +1,24 @@
 import React from 'react';
 import styled from 'styled-components';
+import { QuranReaderStyles } from 'src/redux/slices/QuranReader/styles';
 import Verse from '../../../../types/Verse';
 import VerseText from '../../Verse/VerseText';
 import Translation from '../../../../types/Translation';
 
 type TranslationViewProps = {
   verses: Verse[];
+  styles: QuranReaderStyles;
```
tifa2UP, line 10: nit: refactor to quranReaderStyles both here and in the VerseText

src/components/QuranReader/index.tsx
tifa2UP: In a separate PR, do the following changes:

1. In the redux store, rename translations to selectedTranslations
2. Add a state called isUsingDefaultTranslations, this checks whether the currently selected translations match the default ones, if yes, it'll be set to true, otherwise will be set to false.
3. If isUsingDefaultTranslations is false, do not use initial data
4. If there's no data (and no initial data) show a loading indicator

src/styles/styled.d.ts
```diff
@@ -64,6 +64,10 @@ type LineHeights = {
   jumbo: string;
 };
 
+type Heights = {
+  jumbo: string;
+};
+
```
tifa2UP, line 70: Delete this, use the `spacing` jumbo instead

tifa2UP (approved):
🚀

PR #198, opened 2021-07-19
Add verse page

osamasayed:
### Summary
This PR adds an incrementally static generated page where the user can navigate to a specific verse within a chapter. A specific verse can be accessed through navigation using `{chapterNumber}/{verseNumber}` path in the url e.g. {url}/2/6 which navigates to the 6th verse of chapter 2.

### Test Plan

1. Navigate to the 1st chapter through {url}/1.
2. The list of the verses should show and on the left side of each verse there should be a clickable section with the following format `{chapterNumber}:{verseNumber}` e.g. `1:3`.
3. Click on the section from step 2 for one of verses let's say the 3rd verse.
4. You will be navigated to a page that only shows the 3rd verse.
5. You can change the translations/font from the Developer Utility and it should reflect instantly.
6. Click on `Go Back` section on the top left side of the verse's page which should take you back to the verse's chapter.

### Screenshots

| Before | After |
| ------ | ------ |
| `{url}/1` Before | `{url}/1` After 1 |
|  |`{url}/1/3`After 2 |


sitemap-generator.js
```diff
@@ -1,12 +1,24 @@
...
 const range = require('lodash/range');
+const constants = require('./src/utils/const');
 
 const chapters = range(1, 115);
 
+const generateVersesPaths = () => {
+  const paths = [];
+  range(114).forEach((chapterId) => {
+    const chapterNumber = chapterId + 1;
+    range(constants.CHAPTERS_DETAILS[chapterId].versesCount).forEach((verseId) => {
```
tifa2UP, line 11: I do like having this sort of data be in repo, so that we wouldn't have to wait for an api call, though it shouldn't be living in `constants` 

Will follow-up with an alternate solution

src/components/QuranReader/VerseReader/index.tsx
```diff
@@ -0,0 +1,85 @@
+import styled from 'styled-components';
```
tifa2UP, line 1: We should reuse the same component for the chapters routes. 

The api request is different, otherwise we should treat them as identical.

The same applies for the Reader component

src/components/dls/QuranWord/TextWord.tsx
```diff
@@ -17,7 +17,7 @@ const UTHMANI_HAFS_FONTS = decamelizeKeys({
   textIndopak: 'IndoPak',
 });
 
-const getFontFaimly = (font: string, charType: string) => {
+const getFontFamily = (font: string, charType: string) => {
   if (charType === CharType.End) return DEFAULT_FONT_FAMILY;
 
   return UTHMANI_HAFS_FONTS[font] || DEFAULT_FONT_FAMILY;
```
tifa2UP, line 23: Thank you!

src/pages/[chapterId]/[verseId].tsx
```diff
@@ -0,0 +1,58 @@
...
+type VerseProps = {
+  hasError?: boolean;
+  verseResponse?: VerseResponse;
+};
+
+const Verse: NextPage<VerseProps> = ({ hasError, verseResponse }) => {
+  if (hasError) {
+    return <Error statusCode={500} />;
+  }
+  return <VerseReader initialData={verseResponse.verse} />;
```
Comment (tifa2UP) [inline src/pages/[chapterId]/[verseId].tsx:18]: ditto, should use the same component as the chapters route

src/utils/const.js
```diff
@@ -0,0 +1,121 @@
+/* eslint-disable import/prefer-default-export */
+const CHAPTERS_DETAILS = [
+  { versesCount: 7 },
+  { versesCount: 286 },
+  { versesCount: 200 },
+  { versesCount: 176 },
```
tifa2UP, line 6: I do like the local cache approach, though we should do the following:

1) Create a separate directory outside of src and call it `data`
2) Create a file called under data `Chapters.json`, this file will contain all 114 chapters as a json blob. with the key being the chapter number, and one the children being the verses count. 

In the future, we can add more data to be something like this: 

Screen Shot 2021-07-19 at 11 23 20 AM

types/Verse.ts
```diff
@@ -1,20 +1,34 @@
...
+  sajdahType: null;
+  imageUrl?: string;
+  imageWidth?: number;
+  v1Page?: number;
+  v2Page?: number;
+  codeV1?: string;
+  codeV2?: string;
   translations?: Translation[];
-  audio?: $TsFixMe;
+  audio?: AudioFile;
```
tifa2UP, line 31: This is not technically a file, we can call it `AudioResponse`

src/components/QuranReader/ReaderContainer.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface Props {
+  children: ReactNode | ReactNode[];
+}
+
+const ReaderContainer: React.FC<Props> = ({ children }) => {
+  const isSideBarVisible = useSelector(selectNotes).isVisible;
+  return (
+    <>
+      <ContextMenu />
+      <Container isSideBarVisible={isSideBarVisible}>{children}</Container>
```
tifa2UP, line 18: re: from our discussion earlier. isSideBarVisible should not be passed from the parent container.

Ideally we keep the props minimal and simple. 

And thank you for the container refactor!

tifa2UP (approved):
🚀🚢

PR #199, opened 2021-07-20
Don't use initialData when the default translations change

osamasayed:
### Summary

This PR improves UX by showing a loading indicator instead of falling back to `initialData` which contains the default translations (The Clear Quran and Sahih International) of the `QuranReader` while the user is waiting for the new translations he had selected to show.

### Test Plan

1. Make sure you remove `persist:root` from the local storage first if it has been set.
2. Go to a chapter's page e.g. `{url}/5`.
3. Change the reading view to `translation` from the Developer Utility.
4. Usually the API is pretty quick, so to slow it down, open Google Chrome's Dev Tools, go to `Network` tab then change from `No throttling` to `Slow 3G` for example.
5. Open the Developer Utility again and change the pre-selected translations, maybe select an additional one.
6. A loading indicator should show until the response is received and the QuranReader shouldn't show.
7. Refresh the page, the loading indicator should also show and the QuranReader shouldn't show.

### Screenshots

| Before | After |
| ------ | ------ |
| Old loading Indicator Old loading | New loading Indicator New loading |

### DIFF
--- a/src/components/DeveloperUtility/TranslationsAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationsAdjustment.tsx
@@ -1,14 +1,11 @@
 import { useEffect, useState } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch } from 'react-redux';
 import styled from 'styled-components';
 import { getAvailableTranslations } from 'src/api';
 import AvailableTranslation from 'types/AvailableTranslation';
 import useTranslation from 'next-translate/useTranslation';
-import {
-  selectTranslations,
-  setSelectedTranslations,
-  TranslationsSettings,
-} from 'src/redux/slices/QuranReader/translations';
+import { setSelectedTranslations } from 'src/redux/slices/QuranReader/translations';
+import { useSelectorTyped } from 'src/redux/store';
 
 /**
  * Convert an array of numbers to an array of strings.
@@ -21,7 +18,7 @@ const numbersToStringsArray = (numbersArray: number[]): string[] =>
 
 const TranslationsAdjustment = () => {
   const dispatch = useDispatch();
-  const { selectedTranslations } = useSelector(selectTranslations) as TranslationsSettings;
+  const { selectedTranslations } = useSelectorTyped((state) => state.translations);
   const { lang } = useTranslation();
   const [translations, setTranslations] = useState<AvailableTranslation[]>([]);
   const [isExpanded, setIsExpanded] = useState(false);
--- a/src/components/DeveloperUtility/TranslationsAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationsAdjustment.tsx
@@ -7,6 +7,7 @@ import useTranslation from 'next-translate/useTranslation';
 import {
   selectTranslations,
   setSelectedTranslations,
+  TranslationsSettings,
 } from 'src/redux/slices/QuranReader/translations';
 
 /**
@@ -20,7 +21,7 @@ const numbersToStringsArray = (numbersArray: number[]): string[] =>
 
 const TranslationsAdjustment = () => {
   const dispatch = useDispatch();
-  const selectedTranslations = useSelector(selectTranslations) as number[];
+  const { selectedTranslations } = useSelector(selectTranslations) as TranslationsSettings;
   const { lang } = useTranslation();
   const [translations, setTranslations] = useState<AvailableTranslation[]>([]);
   const [isExpanded, setIsExpanded] = useState(false);
--- a/src/components/DeveloperUtility/TranslationsAdjustment.tsx
+++ b/src/components/DeveloperUtility/TranslationsAdjustment.tsx
@@ -1,11 +1,14 @@
 import { useEffect, useState } from 'react';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import styled from 'styled-components';
 import { getAvailableTranslations } from 'src/api';
 import AvailableTranslation from 'types/AvailableTranslation';
 import useTranslation from 'next-translate/useTranslation';
-import { setSelectedTranslations } from 'src/redux/slices/QuranReader/translations';
-import { useSelectorTyped } from 'src/redux/store';
+import {
+  selectTranslations,
+  setSelectedTranslations,
+  TranslationsSettings,
+} from 'src/redux/slices/QuranReader/translations';
 
 /**
  * Convert an array of numbers to an array of strings.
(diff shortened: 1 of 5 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -41,7 +44,9 @@ const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
   const isVerseView = initialData.verses.length === 1;
   const isSideBarVisible = useSelector(selectNotes).isVisible;
   const quranReaderStyles = useSelector(selectQuranReaderStyles);
-  const selectedTranslations = useSelector(selectTranslations) as number[];
+  const { selectedTranslations, isUsingDefaultTranslations } = useSelector(
+    selectTranslations,
+  ) as TranslationsSettings;
```
tifa2UP, line 49: same here, avoid using `as`

src/redux/slices/QuranReader/translations.ts
```diff
@@ -1,14 +1,55 @@
...
+  if (newTranslations.length !== DEFAULT_TRANSLATIONS.length) {
+    return false;
+  }
+  for (let i = 0; i < newTranslations.length; i += 1) {
+    // if the current translation ID is not inside the default IDs' array, it means the user has changed the default translations.
+    if (DEFAULT_TRANSLATIONS.indexOf(newTranslations[i]) === -1) {
+      return false;
+    }
+  }
+  return true;
```
tifa2UP, line 36: I like this logic, 2 things: 

1. It's simpler to just do `_.isEqual(DEFAULT_TRANSLATIONS.sort(), newTranslations.sort());` in 1 line of code :) 

2. In general, try to move the helper functions to the bottom of the file (before the styles), to make the "main" part easier to follow

src/components/QuranReader/index.tsx
```diff
@@ -55,47 +60,60 @@ const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
...
   const readingView = useSelector(selectReadingView);
-  const pageLimit = isVerseView ? 1 : initialData.pagination.totalPages;
-  const verses = data.flat(1);
+  let body;
   let view;
-
-  if (readingView === ReadingView.QuranPage) {
-    view = <PageView verses={verses} />;
+  // if we are fetching the data (this will only happen when the user has changed the default translations so the initialData will be set to null).
+  if (!data) {
```
tifa2UP, line 72: I have mixed opinions here, thinking out loud. The nested if elses make the code a bit hard to follow. 


My first reaction is to remove the: 

```
 if (readingView === ReadingView.QuranPage) {
      view = <PageView verses={verses} />;
    } else {
      view = <TranslationView verses={verses} quranReaderStyles={quranReaderStyles} />;
    }

```

outside of the nested code block to flatten the structure. 


The other approach is to do something like this: 

```
if(!data) {
 return     
	<>
      <ContextMenu />
      <Container isSideBarVisible={isSideBarVisible}>
		<StyledLoading>loading...</StyledLoading>
	  </Container>
      <Notes />
    </>
}

// rest unchanged 
```

What do you think? No clear cut opinions

src/components/QuranReader/index.tsx
```diff
@@ -55,47 +60,60 @@ const QuranReader = ({ initialData, chapter }: QuranReaderProps) => {
...
+    // TODO: add a proper loading indicator
+    body = <StyledLoading>loading...</StyledLoading>;
```
tifa2UP, line 74: 🤔

The api loads pretty fast, so it's not a concern for now. 

Thinking out loud: the loading indicator currently blocks the entire container. Ideally, the loading indicator only blocks the translations component. 

When a user switches their translation, a good experience would be the following: 

1) the arabic verse is still visible
2) they pick up where they last left off, if they are reading verse 100, they'd still be on verse 100.

Probably going to be more work than the scope of this PR, but FYI.

src/components/DeveloperUtility/TranslationsAdjustment.tsx
```diff
@@ -20,7 +21,7 @@ const numbersToStringsArray = (numbersArray: number[]): string[] =>
 
 const TranslationsAdjustment = () => {
   const dispatch = useDispatch();
-  const selectedTranslations = useSelector(selectTranslations) as number[];
+  const { selectedTranslations } = useSelector(selectTranslations) as TranslationsSettings;
```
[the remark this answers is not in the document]
osamasayed: Okay got it I think it's about time to set the types for the store properly. I am working on that and will be included in the next commit.

src/redux/slices/QuranReader/translations.ts
osamasayed: Oh I forgot about the mighty swiss knife `lodash` :D  thanks for the suggestion.

src/components/DeveloperUtility/TranslationsAdjustment.tsx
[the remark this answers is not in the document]
osamasayed: Okay I agree with you we need to setup the redux slice properly. I will include that in the next commit

src/components/QuranReader/index.tsx
osamasayed: Okay got it. InshAllah in the future we can fine-tune this to only show the loading for the translations instead of the Quran text itself.

src/components/QuranReader/index.tsx
osamasayed: I would go for the second approach but we would be duplicating the `ContextMenu`, `Container` and `Notes` in both cases which is not a problem for now but if we want to add more components inside `QuranReader` in the future, maybe we will have to take the common components into a separate component for example `QuranReaderContainer` or something like that like the previous PR I did and we can reuse it but for now I can do the second approach. I will include it in the upcoming commit.

tifa2UP: +1 to the second approach

src/redux/slices/QuranReader/translations.ts
```diff
@@ -47,14 +19,13 @@ export const translationsSlice = createSlice({
   reducers: {
     setSelectedTranslations: (state, action: PayloadAction<number[]>) => ({
       ...state,
-      isUsingDefaultTranslations: isUsingDefaultTranslations(action.payload), // check if the user is using the default translations on each translation change.
+      // we need to before we compare because there is a corner case when the user changes the default translations then re-selects them which results in the same array as the default one but reversed e.g. instead of [20, 131] it becomes [131, 20].
+      isUsingDefaultTranslations: _.isEqual(DEFAULT_TRANSLATIONS.sort(), action.payload.sort()), // check if the user is using the default translations on each translation change.
```
tifa2UP, line 23: thank you!

tifa2UP (approved):

PR #200, opened 2021-07-21
Fix Quran reader not wrapping on long verses

osamasayed:
### Summary

When there is a long verse e.g. verse 75 of chapter 3, the Quran Reader doesn't wrap the Quranic text so it goes out of the viewport. This PR fixes this by wrapping the text on a new line when it exceeds the width of its container.

### Test Plan

1. Go to `{url}/3/75`.
2. The Quranic text should be wrapped across 2 lines.

### Screenshots

| Before | After |
| ------ | ------ |
| Before |After|


tifa2UP (approved):
lgtm 🚀

PR #201, opened 2021-07-23
add Dropdown DLS, verse settings dropdown and copying verse functionality

osamasayed:
### Summary
This PR adds non-advanced verse copying functionality and introduces `Dropdown` to QDC's DLS. A story of the `Dropdown` is also added.

### Test Plan
1. Go to the first chapter.
2. Click on a verse's settings icon.
3. A dropdown containing a menu that has a Copy item should appear.
4. Click on Copy.
5. The text of the verse should be copied into the clipboard and the text should change to "Copied!" instead of "Copy".
6. After 3 seconds, "Copied!" should be changed back to "Copy".

### Screenshots

| Before | After |
| ------ | ------ |
| Before| Copy |
| | Copied|


src/components/Verse/VerseSettings.tsx
```diff
@@ -0,0 +1,23 @@
...
+import VerseSettingsMenu from './VerseSettingsMenu';
+
+interface Props {
+  verse: Verse;
+}
+
+const VerseSettings: React.FC<Props> = ({ verse }) => (
+  <Dropdown overlay={<VerseSettingsMenu verse={verse} />}>
+    <IconContainer>
+      <MenuMoreIcon />
```
tifa2UP, line 14: 1) Let's call the 3 dots `OverflowMenu` instead of `MenuMore`

2) We should not use "Icons" here. We should use the higher-level component definitions instead. E.g. IconContainer -> Container. MenuMoreIcon -> OverflowMenu. Inside the overflow Menu, you're going to define the icon usage. Lmk if it makes sense!

src/components/Verse/VerseSettingsMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Verse from '../../../types/Verse';
+import VerseSettingsMenuItem from './VerseSettingsMenuItem';
+import CopyIcon from '../../../public/icons/copy.svg';
+
+interface Props {
+  verse: Verse;
+}
+
+const VerseSettingsMenu: React.FC<Props> = ({ verse }) => {
+  const [copied, setCopied] = useState(false);
```
tifa2UP, line 12: nit: `isCopied`, `setIsCopied`

src/components/Verse/VerseSettingsMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout;
```
tifa2UP, line 15: nit: is there a non nodeJS type definition that we can use?

src/components/Verse/VerseSettingsMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (copied === true) {
+      timeoutId = setTimeout(() => setCopied(false), 3 * 1000);
+    }
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [copied]);
+
+  const onCopyClicked = () => {
+    navigator.clipboard
```
tifa2UP, line 26: 1) Is navigator a browser object? If so, we'd need to wrap it around a browser check. Since next JS runs on the server as well.

2) It looks like Clipboard.writeText doesn't have support on all browsers :/. Is there a 3rd party library that we can use instead? https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write

src/hooks/useOutsideClickDetector.ts
```diff
@@ -0,0 +1,34 @@
...
+      document.addEventListener('mousedown', handleClickOutside);
+    }
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, [ref, onClickOutsideDetected, enableDetection]);
+};
+
+export default useOutsideClickDetector;
```
tifa2UP, line 34: Love this. Did you write or is it copied from somewhere? 

Also, can this work with multiple elements at the same time. Or can you only have a single global listener?

src/components/QuranReader/TranslationView/index.tsx
```diff
@@ -24,6 +25,7 @@ const TranslationView = ({ verses, quranReaderStyles }: TranslationViewProps) =>
           <Link as={`/${chapterId}/${verse.verseNumber}`} href="/[chapterId]/[verseId]" passHref>
             <StyledVerseLink>{verse.verseKey}</StyledVerseLink>
           </Link>
+          <VerseSettings verse={verse} />
```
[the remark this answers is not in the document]
osamasayed: Yes I also thought settings doesn't sound right :) . Thanks for the suggestion, will change it.

src/components/Verse/VerseSettingsMenu.tsx
osamasayed: Well the other option is `ReturnType<typeof setTimeout>` but when you hover on it, it still shows `NodeJS.Timeout` since the type that `clearTimeout` accepts is `NodeJs.Timeout`:

`declare function clearTimeout(timeoutId: NodeJS.Timeout): void;`

src/components/Verse/VerseSettings.tsx
naveed-ahmad: We used `copy-to-clipboard` for writing text to clipboard in [v2](https://github.com/quran/quran.com-frontend-v2/blob/c33675c1994cecf8041134ee2209fe1dbcf2badc/app/javascript/controllers/verse_controller.js#L74)

src/components/Verse/VerseSettingsMenu.tsx
naveed-ahmad: We used `copy-to-clipboard` for writing text to clipboard in [v2](https://github.com/quran/quran.com-frontend-v2/blob/c33675c1994cecf8041134ee2209fe1dbcf2badc/app/javascript/controllers/verse_controller.js#L74)

src/components/Verse/VerseSettingsMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+      .writeText(
+        // remove the last element which is the verse's number then join all of them into 1 string
+        verse.words
```
naveed-ahmad, line 29: Copying v2, and v3 glyph codes don't make any sense. We can fetch text uthmani for each ayah using `fields` query string in [makeVersesUrl](https://github.com/quran/quran.com-frontend-next/blob/4ba85e5cdb7be1e2d8deeb2d9141d042808c26ec/src/utils/apiPaths.ts#L11)

Then copy Uthmani text of ayah.

src/hooks/useOutsideClickDetector.ts
osamasayed: No it's written from scratch :) . Also it works with multiple elements at once so that at any given time, only one dropdown will be visible because if you click at any verse's actions icon, it's considered you clicked outside of all other verse actions' icons which will cause them to close (if any is opened).

src/components/Verse/VerseSettings.tsx
osamasayed: Okay got it.

src/components/Verse/VerseSettingsMenu.tsx
osamasayed: JAK for pointing that out. I assumed that the user would want to copy the verse in the current selected Quran font. I will include `text_uthmani` under `fields` when calling the API.

src/components/Verse/VerseSettingsMenu.tsx
osamasayed: There are a few libraries that do polyfilling for copying to clipboard for example: [clipboard-copy](https://github.com/feross/clipboard-copy), [clipboard-polyfill](https://github.com/lgarron/clipboard-polyfill) and [copy-to-clipboard](https://github.com/sudodoki/copy-to-clipboard) but I think we can go with `copy-to-clipboard` since it's already tried and tested. What do you think? @tifa2UP

osamasayed: There are a few libraries that do polyfilling for copying to clipboard for example: [clipboard-copy](https://github.com/feross/clipboard-copy), [clipboard-polyfill](https://github.com/lgarron/clipboard-polyfill) and [copy-to-clipboard](https://github.com/sudodoki/copy-to-clipboard) but I think we can go with `copy-to-clipboard` since it's already tried and tested. What do you think? @tifa2UP

osamasayed: I just tried `copy-to-clipboard`, it doesn't seem to work neither on Chrome nor on Safari for some reason :/ `clipboard-copy` does work well.

tifa2UP (approved):
🚀

PR #202, opened 2021-07-24
Add Modal DLS

osamasayed:
### Summary

This PR add `Modal` component to QDC's DLS a story has been included as well to showcase different use-cases for the component.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -12,6 +12,17 @@ module.exports = {
     reactDocgen: false,
   },
   webpackFinal: async (config) => {
+    // Default rule for images /\.(svg|ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|cur|ani|pdf)(\?.*)?$/
+    const fileLoaderRule = config.module.rules.find(rule => rule.test && rule.test.test && rule.test.test('.svg'));
+    // we should exclude handling .svg files by file-loader.
+    fileLoaderRule.exclude = /\.svg$/;  
+    // and add our own loader (@svgr/webpack)
+    config.module.rules.push({
+      test: /\.svg$/,
+      enforce: 'pre',
+      loader: require.resolve('@svgr/webpack'),
+    });
+    // add a loader to handle ts|tsx files.
     config.module.rules.push({
       test: /\.(ts|tsx)$/,
       use: [
--- a/src/components/dls/Modal/Modal.stories.tsx
+++ b/src/components/dls/Modal/Modal.stories.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 
-import Modal from './Modal';
+import Modal, { ModalSize } from './Modal';
 
 export default {
   title: 'dls|Modals',
@@ -61,15 +61,16 @@ export default {
         category: 'Optional',
       },
     },
-    width: {
-      defaultValue: 520,
-      description: 'The width of the modal in pixels.',
+    size: {
+      defaultValue: ModalSize.Medium,
+      description: 'The size of the modal.',
       table: {
         category: 'Optional',
       },
-      control: { type: 'number' },
+      options: [ModalSize.Small, ModalSize.Medium, ModalSize.Large, ModalSize.XLarge],
+      control: { type: 'select' },
     },
-    closeIcon: {
+    customCloseIcon: {
       description: 'Custom close icon',
       table: {
         category: 'Optional',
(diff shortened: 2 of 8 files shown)


.storybook/main.js
```diff
@@ -12,6 +12,17 @@ module.exports = {
     reactDocgen: false,
   },
   webpackFinal: async (config) => {
+    // Default rule for images /\.(svg|ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|cur|ani|pdf)(\?.*)?$/
+    const fileLoaderRule = config.module.rules.find(rule => rule.test && rule.test.test && rule.test.test('.svg'));
```
tifa2UP, line 16: nit: should this be rules instead of rule?

src/components/dls/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+        category: 'Optional',
+      },
+    },
+    title: {
+      description: 'The title of the modal.',
+      table: {
+        category: 'Optional',
+      },
+    },
+    width: {
```
tifa2UP, line 64: ideally we don't specify a width. We instead have a `size` prop with preset widths like: `default` `large`.

This constraint is good for consistency

src/components/dls/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+      defaultValue: 520,
+      description: 'The width of the modal in pixels.',
+      table: {
+        category: 'Optional',
+      },
+      control: { type: 'number' },
+    },
+    closeIcon: {
```
tifa2UP, line 72: rename to `customCloseIcon` otherwise it might be understood as a boolean prop that takes true/false

src/components/dls/Modal/Modal.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useCallback, useRef, ReactNode, useEffect } from 'react';
+import useKeyPressedDetector from 'src/hooks/useKeyPressedDetector';
+import useOutsideClickDetector from 'src/hooks/useOutsideClickDetector';
+import styled from 'styled-components';
+import Close from '../../../../public/icons/close.svg';
+import Button, { ButtonSize } from '../Button/Button';
+
+interface Props {
```
tifa2UP, line 8: I tried out the modal locally. Overall it works really well masha'Allah. I noticed a couple of things: 

1) the x button is not scaled correctly:

2021-07-24 at 14 29 16@2x

2) The vertically scrolling and horizontally scrolling modals should not behave the way they do. There should be a max width specified by `size` prop, and there should be a max height (perhaps a % of the viewport). If the children size exceeds the modal size, they should overflow (scroll within the container). This is the effect when I try to overflow the gestalt (pinterest design system library) modal: 

2021-07-24 at 14 36 38@2x

.storybook/main.js
osamasayed: But `fileLoaderRule` actually refers to only one rule that handles `.svg` and uses `file-loader` out of the webpack rules that storybook uses by default. This rule in the [default webpack config](https://github.com/storybookjs/storybook/blob/next/lib/builder-webpack4/src/preview/base-webpack.config.ts) is as follows:

```
{
          test: /\.(svg|ico|jpg|jpeg|png|apng|gif|eot|otf|webp|ttf|woff|woff2|cur|ani|pdf)(\?.*)?$/,
          loader: require.resolve('file-loader'),
          options: {
            esModule: false,
            name: isProd
              ? 'static/media/[name].[contenthash:8].[ext]'
              : 'static/media/[path][name].[ext]',
          },
},
```
So `fileLoaderRule` refers to that rule that we will later modify to exclude `.svg` files and add our own loader.

src/components/dls/Modal/Modal.tsx
osamasayed: Regarding 1st point it's an issue with Storybook and `Button` component not with `Modal` itself. The button looks fine in the UI but it's shifted a bit when it's viewed as a story.

Screen Shot 2021-07-25 at 08 24 44

This is how the close button for the modal looks without Storybook: 

Screen Shot 2021-07-25 at 08 26 40

src/components/dls/Modal/Modal.stories.tsx
osamasayed: I thought having the width in pixels would give a lot of flexibility when using the modal versus when we have pre-defined limited values because they might not match exactly what the user of the `Modal` is looking for. I was looking at the [antd's component API](https://ant.design/components/modal/#API) for modal and they got it defined in pixels. 

Screen Shot 2021-07-25 at 11 00 22

tifa2UP (approved):

PR #203, opened 2021-07-24
fix QCF font

naveed-ahmad:
### Summary
- Fixed QCF font rendering issue. Last value of lodash [range](https://lodash.com/docs/4.17.15#range) wasn't inclusive. [1,1]  or [114,114] would yield an empty array. And no font face would generate. 
- Justify center ayah for few pages to match the printed Mushaf layout.

### Test Plan

1. Go to `{url}/114`.
2. Change the layout to page view
3. Last two lines should be centered align

### Screenshots

| Before | After |
| ------ | ------ |
| Before |After|


src/components/Verse/VerseText.tsx
```diff
@@ -1,24 +1,26 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import styled from 'styled-components';
 import { useSelector } from 'react-redux';
 import Word from 'types/Word';
 import { selectReadingView } from 'src/redux/slices/QuranReader/readingView';
 import QuranWord from '../dls/QuranWord/QuranWord';
 import { selectQuranReaderStyles, QuranReaderStyles } from '../../redux/slices/QuranReader/styles';
 import { ReadingView } from '../QuranReader/types';
+import isCenterAlignPage from './pageUtils';
```
tifa2UP, line 9: nit: `isCenterAlignedPage`

tifa2UP (dismissed):

PR #218, opened 2021-08-07
Use JSON instead of YAML for ESLint

osamasayed:
### Summary
This PR uses JSON instead of YAML for `.eslintrc` file.


tifa2UP (approved):
🛳️

PR #221, opened 2021-08-08
Add sass and classnames

osamasayed:
### Summary
This PR adds [`sass`](https://www.npmjs.com/package/sass) and [`classnames`](https://www.npmjs.com/package/classnames) packages as we decided to move away from using `styled-components` and will be using [sass](https://sass-lang.com/) which is [supported natively](https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css) by Next.js.


package.json
```diff
@@ -44,6 +46,7 @@
     "react-scroll": "^1.7.16",
     "react-use": "^13.27.0",
     "redux-persist": "^6.0.0",
+    "sass": "^1.37.5",
```
tifa2UP, line 49: this should be a devDependency

tifa2UP (approved):

PR #245, opened 2021-08-10
Update Storybook to use SCSS files

osamasayed:
### Summary
This PR updates the settings and version of Storybook for us to be able to use it with `.scss` modules.


.storybook/main.js
```diff
@@ -1,52 +1,37 @@
...
+        cssLoaderOptions: {
+          modules: {
+            auto: true
+          }
+        }
+      }
+    },
+    "@storybook/addon-links",
+    "@storybook/addon-essentials",
+    "@storybook/addon-a11y"
```
tifa2UP, line 19: Did you remove this part on purpose?

```
 '@storybook/addon-storysource',
 '@storybook/addon-a11y/register',
```

.storybook/main.js
```diff
@@ -1,52 +1,37 @@
...
   ],
   typescript: {
-    reactDocgen: false,
+    reactDocgen: false
+  },
+  "core": {
+    "builder": "webpack5"
   },
   webpackFinal: async (config) => {
```
tifa2UP, line 27: good riddance, didn't like custom webpack setups in the first place. 

Thank you!

tifa2UP (approved):
Code looks good mA. Waiting for CI to pass

.storybook/main.js
osamasayed: `@storybook/addon-a11y` was set-up in `preview.js` which is not the [right way](https://www.npmjs.com/package/@storybook/addon-a11y) to set it up along with using `@storybook/addon-a11y/register`, so removed the usage from `preview.js` and added the correct syntax here.

PR #251, opened 2021-08-11
Add share verse action

osamasayed:
### Summary
This PR adds the ability to share the verse's url.

### Test Plan
1. Click on the verse's actions menu.
2. Click on share. The current url should be copied to the clipboard and a message indicating that should appear.
3. The message should return to default after 3 seconds.

### Screenshots

| Before | After |
| ------ | ------ |
| Screen Shot 2021-08-12 at 03 43 34 | Screen Shot 2021-08-12 at 03 41 55 |
|  | Screen Shot 2021-08-12 at 03 41 59 |

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -1,0 +1,7 @@
+/* eslint-disable import/prefer-default-export */
+export const getCurrentPath = () => {
+  if (typeof window !== 'undefined') {
+    return window.location.href;
+  }
+  return '';
+};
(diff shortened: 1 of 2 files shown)


src/utils/url.ts
```diff
@@ -0,0 +1,13 @@
+export const getCurrentPath = () => {
+  if (typeof window !== 'undefined') {
+    return window.location.href;
+  }
+  return '';
+};
+
+export const getOrigin = () => {
```
tifa2UP, line 8: nit: rename to `getWindowOrigin`

tifa2UP (approved):

PR #257, opened 2021-08-12
Add ability to read a specific Quranic page

osamasayed:
### Summary
This PR adds the ability to read a specific Quranic page through the url `/page/{pageNumber}`.

### Screenshots

| Before | After |
| ------ | ------ |
|  | Screen Shot 2021-08-13 at 05 57 26 |

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -15,7 +15,7 @@ import { selectReadingPreference } from '../../redux/slices/QuranReader/readingP
 import PageView from './PageView';
 import TranslationView from './TranslationView';
 import { QuranReaderDataType, ReadingPreference } from './types';
-import { makeJuzVersesUrl, makeVersesUrl } from '../../utils/apiPaths';
+import { makeJuzVersesUrl, makePageVersesUrl, makeVersesUrl } from '../../utils/apiPaths';
 import { QuranReaderStyles, selectQuranReaderStyles } from '../../redux/slices/QuranReader/styles';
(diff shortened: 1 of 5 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -157,6 +158,14 @@ const getRequestKey = (
...
     });
   }
+  if (isQuranicPageData) {
+    return makePageVersesUrl(id, {
+      page,
+      wordFields: `verse_key, verse_id, page_number, location, ${quranReaderStyles.quranFont}`,
+      translations: selectedTranslations.join(', '),
+    });
+  }
+
```
tifa2UP, line 168: Why is there a specific call for Quranic page data? Is it not request translations in the payload?

src/utils/validator.ts
```diff
@@ -59,3 +59,21 @@ export const isValidJuzId = (juzId: string): boolean => {
...
+ * @returns {boolean}
+ */
+export const isValidPageId = (juzId: string): boolean => {
+  const juzIdNumber = Number(juzId);
+  // if it's not a numeric string or it's numeric but out of the range of chapter 1->604
+  if (Number.isNaN(juzIdNumber) || juzIdNumber > 604 || juzIdNumber < 1) {
+    return false;
+  }
+  return true;
+};
```
tifa2UP, line 79: this should be pageId, not JuzId. right?

osamasayed: Yes you are right, forgot to rename :)

PR #259, opened 2021-08-13
AudioPlayer: hide ui when no audiofile & add loading UI

muhajirdev:
### Summary
- hide ui when no audiofile
-  audioplayer opacity 50% when loading

### Test Plan
- you should see no audioplayer, before you click "play" button
- try to play the audio, and see the loading for a split second
- can't click "play" button when it is loading.

### Screenshots

Loading 
image


tifa2UP:
1. Change the initial state for audio visibility to Hidden. 
2. Delete the audio styles slice, instead create an object called `isMinimized` to check the minimized state. The visibility such as hidden or not hidden will be inferred on the component level based on the state.

tifa2UP (approved):

PR #260, opened 2021-08-13
persist AudioPlayer state, except isPlaying

muhajirdev:
### Summary
- persist the audio player state, all of it, except `isPlaying`

### Test Plan
- try to play the audio, then reload the browser
- all-state should be preserved. But the audio player is paused


tifa2UP (approved):
love this. Thank you!

tifa2UP:
Handle the case were you resume session after closing the tab. The current time in redux is correct, but the current time in the audio file is not.

muhajirdev:
will the current time issue in another PR

PR #269, opened 2021-08-15
play verse timestamp with new API

muhajirdev:
### Summary
- previously we call api to get timestamp api everytime we click "play" verse.
- now we don't call the API anymore

- refactored `getRequestKey` a little bit.

### Test Plan
- click the "play" button on the left side of each verse, in translation reading view. See the network request. there should be no new API call.

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -169,6 +169,7 @@ const getRequestKey = ({
   if (quranReaderDataType === QuranReaderDataType.Juz) {
     return makeJuzVersesUrl(id, {
       page,
+      reciter,
       translations: selectedTranslations.join(', '),
       ...getDefaultWordFields(quranReaderStyles.quranFont),
     });
@@ -176,6 +177,7 @@ const getRequestKey = ({
   if (quranReaderDataType === QuranReaderDataType.Page) {
     return makePageVersesUrl(id, {
       page,
+      reciter,
       translations: selectedTranslations.join(', '),
       ...getDefaultWordFields(quranReaderStyles.quranFont),
     });
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -12,7 +12,7 @@ import {
 import classNames from 'classnames';
 import { selectTafsirs, TafsirsSettings } from 'src/redux/slices/QuranReader/tafsirs';
 import { getDefaultWordFields } from 'src/utils/api';
-import { selectReciter } from 'src/redux/slices/AudioPlayer/state';
+import { selectIsUsingDefaultReciter, selectReciter } from 'src/redux/slices/AudioPlayer/state';
 import { selectReadingPreference } from '../../redux/slices/QuranReader/readingPreference';
 import PageView from './PageView';
 import TranslationView from './TranslationView';
@@ -57,6 +57,7 @@ const QuranReader = ({
   ) as TranslationsSettings;
   const { selectedTafsirs, isUsingDefaultTafsirs } = useSelector(selectTafsirs) as TafsirsSettings;
   const reciter = useSelector(selectReciter, shallowEqual);
+  const isUsingDefaultReciter = useSelector(selectIsUsingDefaultReciter);
   const { data, size, setSize, isValidating } = useSWRInfinite(
     (index) =>
       getRequestKey({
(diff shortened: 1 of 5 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -55,9 +56,10 @@ const QuranReader = ({
     selectTranslations,
   ) as TranslationsSettings;
   const { selectedTafsirs, isUsingDefaultTafsirs } = useSelector(selectTafsirs) as TafsirsSettings;
+  const reciter = useSelector(selectReciter, shallowEqual);
   const { data, size, setSize, isValidating } = useSWRInfinite(
     (index) =>
-      getRequestKey(
+      getRequestKey({
```
osamasayed, line 62: Me and @tifa2UP discussed this recently. There is no organizational benefit of converting this into an object vs having multiple params both ways `getRequestKey` is still bloated.

src/components/QuranReader/index.tsx
```diff
@@ -132,22 +135,35 @@ const QuranReader = ({
...
+  index,
+  quranReaderStyles,
+  quranReaderDataType,
+  selectedTranslations,
+  selectedTafsirs,
+  reciter,
+}: RequestKeyInput): string => {
   // if the response has only 1 verse it means we should set the page to that verse this will be combined with perPage which will be set to only 1.
   const page = isVerseData || isTafsirData ? initialData.verses[0].verseNumber : index + 1;
   if (quranReaderDataType === QuranReaderDataType.Juz) {
```
osamasayed, line 169: The `reciter` should be passed here too because the user can be reading a Juz in the translation mode and he might want to play a verse (that's to assume the Juz's API support that).

src/components/QuranReader/index.tsx
```diff
@@ -67,7 +69,8 @@ const QuranReader = ({
         isVerseData,
         isTafsirData,
         id,
-      ),
+        reciter: reciter.id,
+      }),
     verseFetcher,
     {
       initialData: isUsingDefaultTranslations && isUsingDefaultTafsirs ? initialData.verses : null, // initialData is set to null if the user changes/has changed the default translations/tafsirs so that we can prevent the UI from falling back to the default translations while fetching the verses with the translations/tafsirs the user had selected and we will show a loading indicator instead.
```
tifa2UP, line 76: You'll have to set initialData to null when we're not using the default Reciter. will explain why over a call iA

src/components/Verse/PlayVerseAudioButton.tsx
```diff
@@ -1,20 +1,21 @@
...
+import Verse from 'types/Verse';
 
 import PlayIcon from '../../../public/icons/play-circle-outline.svg';
 import Button from '../dls/Button/Button';
 
 interface Props {
-  verseKey: string;
+  verse: Verse;
 }
 const PlayVerseAudioButton = (props: Props) => {
```
tifa2UP, line 12: A few comments here. 

1) I missed this in the previous review. We follow the pattern of destructuring props inside of the component parameters. E.g. 

```
const PlayVerseAudioButton = ({verse}: PlayVerseAudioButtonProps) => {

}
```

2) Rename the Props interface to be more specific (See comment above) 

3) If this button only cares about timestamps, we should be passing all of the verse object. It'll make our data flow bloated. For e.g. If we want the button to play audio starting from a certain time. It should take a *single* prop called `playFrom` in seconds. And then dispatch that value to the store.

src/components/Verse/PlayVerseAudioButton.tsx
```diff
@@ -1,20 +1,21 @@
...
   const dispatch = useDispatch();
   return (
     <Button
       icon={<PlayIcon />}
       onClick={() => {
-        dispatch(playVerse(props.verseKey));
+        dispatch(playVerse(props.verse));
```
tifa2UP, line 18: This comment applies to the rest of the PR. I don't think that we should still be using the playVerse functionality. Probably we should deprecate it. 

What we should do instead is that we directly set the current time and simplify all the other logic.

types/Verse.ts
```diff
@@ -31,6 +31,13 @@ interface Verse {
   translations?: Translation[];
   tafsirs?: Tafsir[];
   audio?: AudioResponse;
+  timestamps?: {
+    verseKey: string;
+    timestampFrom: number;
+    timestampTo: number;
+    duration: number;
+    segments: [number[]];
+  };
```
tifa2UP, line 40: Thank you!

src/redux/slices/AudioPlayer/state.ts
```diff
@@ -71,21 +72,20 @@ export const loadAndPlayAudioFile = createAsyncThunk<void, number>(
  * @param {number} verseKey example 1:1 -> al-fatihah verse 1
  *
  */
-export const playVerse = createAsyncThunk<void, string>(
+export const playVerse = createAsyncThunk<void, Verse>(
```
tifa2UP, line 75: refactor `playVerse` to `playFrom` that takes: (timestamp, chapterNumber, reciter). 

If the chapterNumber or Reciter do not match, fetch the new audio file.

If it does, set current time. 

This will be the function we use to play a specific word as well, so it's not specific to verses.

src/components/Verse/PlayVerseAudioButton.tsx
```diff
@@ -1,20 +1,21 @@
...
 import PlayIcon from '../../../public/icons/play-circle-outline.svg';
 import Button from '../dls/Button/Button';
 
-interface Props {
-  verseKey: string;
+interface PlayVerseAudioProps {
+  verse: Verse;
 }
-const PlayVerseAudioButton = (props: Props) => {
+const PlayVerseAudioButton = (props: PlayVerseAudioProps) => {
```
tifa2UP, line 12: the component should only take (timestamp, chapterNumber, reciter) as props

src/components/QuranReader/index.tsx
tifa2UP: Chatted with Mujajir offline. The benefits are making it an object is that: 

1. The order of the parameter doesn't matter
2. you don't have to explicitly pass nulls.

I think that this makes more sense. Let's leave it that way @osamasayed

tifa2UP (approved):

PR #275, opened 2021-08-17
Improve Quran Reader

osamasayed:
### Summary
This PR solves a few issues with the Quran Reader as following:

1. The mobile layout was broken.
2. On Desktop the alignment of the text was not correct which resulted in some lines showing less in width than the rest.
3. There were no page numbers.
4. There was no header for a Surah when in reading mode.
5. There was no way to load more data by scroll if the initial content is not big enough to reach the fixed scrolling threshold (2000 px) so in this PR a load more button was introduced.
6. Verse number was not showing correctly when choosing IndoPak font.

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2021-08-17 at 16 06 54|Screen Shot 2021-08-17 at 16 06 39|
|Screen Shot 2021-08-17 at 16 08 40|Screen Shot 2021-08-17 at 16 08 33|
|Screen Shot 2021-08-17 at 16 10 11|Screen Shot 2021-08-17 at 16 09 39|
|Screen Shot 2021-08-17 at 16 14 36|Screen Shot 2021-08-17 at 16 14 45|


src/components/QuranReader/QuranReader.module.scss
```diff
@@ -30,3 +31,15 @@
   display: flex;
   justify-content: center;
 }
+
+.loadMoreContainer {
+  text-align: center;
+  button {
+    background: none !important;
```
tifa2UP, line 38: strongly advise against using `!important`

We can extend the button component with props that give us background none, etc.

tifa2UP (approved):

PR #289, opened 2021-08-22
add scss color variable : success error warning alert

muhajirdev:


muhajirdev:
prepareing the PR for button component DLS

muhajirdev:
# Specs

Sizes:
- small
- normal (default, no need to write)
- large

shape:
- square (border radius 5px)
- circle (border radius 100%)

prefix suffix 
- prefix (support prefix icon)
- suffix (support suffix icon)
- both can be used at the same time

on LTR prefix will be on the left
on RTL prefix will be on the right

align
- start
- grow

so` start1` and `grow` only determine the position of the `children`. The prefix will always be on the left, and the suffix always on the right.
NEED TO CLARIFY: What about RTL?

type
- primary (default, no need to write this)
- secondary
- success
- error
- warning
- alert

variant
- shadow -> text white / black. with backgroundColor following the `type`
- ghost -> no background color, with color following the `type`

loading : boolean
 - when we have both `prefix` and `loading`-> the icon will be replaced with " circular loading icon" animation
 - when we have we don't have prefix. The `children` will be replaced with "..." animation ( see playground section in vercel design)


disabled
button with `disabled`
cursor `not-allowed`
color grey ish

width: number
custom width
text will be truncated if the we have `width` and text is too long

onClick

## Default styles
- width is not determined by default. It will just follow the default browser behavior

## Using with Icon
```
<Button>
   <SettingsIcon />
<Button>
```
## Using with text and Icon
```
<Button prefix={SettingsIcon />>
   some text here
</Button>
```

## Current Usage
- size
  XSmall = 'xsmall',
  Small = 'small',
  Medium = 'medium',
  Large = 'large',
- desktopSize
- onClick

tifa2UP (approved):

PR #302, opened 2021-08-25
[WIP] Modal DLS

muhajirdev:


src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+    transform: "translate(-50%, -50%) scale(1)";
+  }
+}
+
+.content {
+  background-color: white;
+  border-radius: 6px;
+  overflow: hidden;
+  box-shadow: hsl(206 22% 7% / 35%) 0px 10px 38px -10px,
+    hsl(206 22% 7% / 20%) 0px 10px 20px -15px;
```
tifa2UP, line 35: fyi, this will be derived for a token I'm currently working on

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 90vw;
+  max-width: 450px;
+  max-height: 85vh;
```
tifa2UP, line 42: these should be sass variables declared at the top of the file.

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+
+  @media (prefers-reduced-motion: no-preference) {
+    animation: contentShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
```
tifa2UP, line 45: can we make this a token?

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+    will-change: transform;
+  }
+
+  &:focus {
+    outline: none;
+  }
+}
+
+.body {
+  padding: 25px;
```
tifa2UP, line 55: no hardcoded padding

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+  line-height: var(--line-height-jumbo);
+}
+.subtitle {
+  text-align: center;
+  font-size: var(--font-size-normal);
+  font-weight: var(--font-weight-normal);
+}
+
+.description {
+  margin: 10px 0 20px;
```
tifa2UP, line 71: ditto

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
+@keyframes overlayShow {
```
osamasayed, line 1: There is an existing keyframes util inside `_utility.scss` which we should use to support different browsers.

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+  100% {
+    opacity: 1;
+  }
+}
+
+.overlay {
+  background-color: rgba($color: #000000, $alpha: 0.1);
+  position: fixed;
+  inset: 0;
+  @media (prefers-reduced-motion: no-preference) {
```
osamasayed, line 14: This should go to `_utility.scss`

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+    animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+  }
+}
+
+@keyframes contentShow {
```
osamasayed, line 19: ditto: use the util inside `_utility.scss`

src/components/dls/ModalNew/Modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+  border-radius: 6px;
+  overflow: hidden;
+  box-shadow: hsl(206 22% 7% / 35%) 0px 10px 38px -10px,
+    hsl(206 22% 7% / 20%) 0px 10px 20px -15px;
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 90vw;
+  max-width: 450px;
```
osamasayed, line 41: We should use multiples of `--spacing-{scale}` css variables

src/components/dls/ModalNew/Modal.module.scss
osamasayed, line 71: We should use multiples of `--spacing-{scale}` css variables

src/components/dls/ModalNew/Modal.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import * as DialogPrimitive from '@radix-ui/react-dialog';
+import styles from './Modal.module.scss';
+
+const Modal = ({ children, trigger }) => (
+  <DialogPrimitive.Root>
+    <DialogPrimitive.Overlay className={styles.overlay} />
+    {trigger}
```
osamasayed, line 8: Can we change this to have `<DialogPrimitive.Trigger>{trigger}</DialogPrimitive.Trigger>` instead of having to use `Modal.Trigger` as part of the prop every time `Modal` is being used?

src/components/dls/ModalNew/Modal.tsx
```diff
@@ -0,0 +1,39 @@
...
+const Body = ({ children }) => <div className={styles.body}>{children}</div>;
+const Title = ({ children }) => (
+  <DialogPrimitive.Title className={styles.title}>{children}</DialogPrimitive.Title>
+);
+const Subtitle = ({ children }) => (
+  <DialogPrimitive.Description className={styles.subtitle}>{children}</DialogPrimitive.Description>
+);
+
+const Actions = ({ children }) => <div className={styles.actionsContainer}>{children}</div>;
+const Action = ({ children }) => (
```
osamasayed, line 28: Shouldn't an action have a callback function to perform the action itself? or how are we going to perform the action?

src/components/dls/ModalNew/Modal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  <DialogPrimitive.Close className={styles.action}>{children}</DialogPrimitive.Close>
```
osamasayed, line 29: Is close the only action we will perform? I think action should have multiple types and a callback to actually perform the action.

src/components/dls/ModalNew/Modal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  </DialogPrimitive.Root>
+);
+
+const Content = React.forwardRef<HTMLDivElement, any>(({ children, ...props }, forwardedRef) => (
+  <DialogPrimitive.Content {...props} ref={forwardedRef} className={styles.content}>
+    {children}
+  </DialogPrimitive.Content>
+));
+
+const Body = ({ children }) => <div className={styles.body}>{children}</div>;
```
osamasayed, line 19: I think we should split the body into header and the body itself. So we add another section called `Modal.Header` that will include `Title` and `Subtitle` and the body will only include the content itself.

muhajirdev: i see, agree

src/components/dls/ModalNew/Modal.tsx
muhajirdev: yes, I am still thinking a little bit about this. 
But one of the idea is that ideally, we should be able to use `<Button>` as trigger, so we can use buttons's props. such as `type` `variant`, `shape` etc. For example

```
<Modal trigger={<Button type={ButtonType.Secondary} variant={ButtonVariant.Ghost} />}>
```

I'll probably try to play a little bit if it's possible to do that.

do you have any comment/idea about this?

src/components/dls/ModalNew/Modal.tsx
muhajirdev: good point, 
yes we should pass callback here. This way we'll be able to do many other stuff with this button.

muhajirdev: by `types`, do you mean we'll have type like "submit" will be a green button, and then "cancel" will be a gray button?

src/components/dls/ModalNew/Modal.tsx
osamasayed: I think we are forced to use `DialogPrimitive.Trigger` for any type. So we can introduce 2 new props for example:
1. `triggerType` which can be an enum called `TriggerTypes` to have `Button`, `Default` or others values  and set to `TriggerTypes.Default` by default.
2. `triggerProps` which is optional and can be an object and in the case of a `Button`, we can pass something like `{type:ButtonType.Secondary, variant:ButtonVariant.Ghost}`.
and we use it like this:

```
<DialogPrimitive.Trigger>
{triggerType === TriggerTypes.Button ? <Button {...triggerProps}>{trigger}</Button>: <div {...triggerProps}>{trigger}</div>}
</DialogPrimitive.Trigger>
```

What do you think?

src/components/dls/ModalNew/Modal.tsx
osamasayed: You can have a look at [Antd Modal's footer](https://ant.design/components/modal/#components-modal-demo-footer). in Radix there is only `Close` action technically. So we should link it to a prop like `closable` and we have a section called footer which can be anything like how Antd does it which can include a button or a text or anything.

PR #303, opened 2021-08-25
Upgrade Combobox DLS

osamasayed:


src/components/dls/Forms/Combobox/Combobox.module.scss
```diff
@@ -0,0 +1,179 @@
...
+    outline: none;
+  }
+}
+
+.comboboxContainer {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  max-width: 100%;
+  width: calc(8 * var(--spacing-mega));
```
tifa2UP, line 101: you can define this as a variable if you'd like (e.g. `$combo-box-width: 3 rem`) . Up to you.

src/components/dls/Forms/Combobox/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+  useEffect,
+  RefObject,
+  useCallback,
+  useRef,
+} from 'react';
+import classNames from 'classnames';
+import useOutsideClickDetector from 'src/hooks/useOutsideClickDetector';
+import useKeyPressedDetector from 'src/hooks/useKeyPressedDetector';
+import useScroll from '../../../../hooks/useScrollToElement';
+import CaretIcon from '../../../../../public/icons/caret-down.svg';
```
tifa2UP, line 18: no need to address it in this PR, I think that you can do `public/icons/...` directly since we configured the root in typescript

PR #332, opened 2021-08-29
Add multiselect support to Combobox

osamasayed:
### Summary
This PR adds multi-select support to the current Combobox component.

### Screenshots

| Before | After |
| ------ | ------ |
||Screen Shot 2021-08-29 at 18 42 52|
||Screen Shot 2021-08-29 at 18 41 42|
||Screen Shot 2021-08-29 at 18 42 22|


src/components/dls/Forms/Combobox/ComboboxItem.tsx
```diff
@@ -47,23 +47,24 @@ const ComboboxItem: React.FC<Props> = ({
           styles.itemContainer,
           { [styles.disabledItem]: disabled },
           { [styles.enabledItem]: !disabled },
+          { [styles.checkedItem]: checked },
         )}
       >
         <input
-          type="radio"
+          type="checkbox"
```
tifa2UP, line 54: should this be conditional radio/conditional based on the multiselect prop?

src/hooks/useFocusElement.ts
```diff
@@ -0,0 +1,24 @@
+import { useRef, RefObject } from 'react';
+
+/**
+ * A hook that focus a referenced element.
+ *
+ * @param {FocusOptions} options
+ * @returns {[() => void, RefObject<T>]}
+ */
+export const useScrollToElement = <T extends HTMLElement>(
```
tifa2UP, line 9: Is the function used to scroll to Element to focus an element? 

nit: focuses*

src/components/dls/Forms/Combobox/ComboboxItem.tsx
osamasayed: This is done intentionally as before, there was an issue even with single select which is once you select you cannot un-select due it being a radio and not a checkbox.

src/hooks/useFocusElement.ts
osamasayed: My bad. This needs renaming

tifa2UP (approved):

PR #365, opened 2021-09-03
DLS: Separator

muhajirdev:
### Summary
- Add separator dls

Need to discuss some approach about inline sytling

### DIFF
--- a/src/components/dls/Separator/Separator.module.scss
+++ b/src/components/dls/Separator/Separator.module.scss
@@ -1,0 +1,15 @@
+$separator-size: 1px;
+
+.base {
+  background-color: var(--color-borders-hairline);
+
+  &[data-orientation="horizontal"] {
+    height: $separator-size;
+    width: 100%;
+  }
+
+  &[data-orientation="vertical"] {
+    height: 100%;
+    width: $separator-size;
+  }
+}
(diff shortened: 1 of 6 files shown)


src/components/dls/Separator/Separator.module.scss
```diff
@@ -0,0 +1,15 @@
...
+.base {
+  background-color: var(--color-borders-hairline);
+
+  &[data-orientation="horizontal"] {
+    height: $separator-size;
+    width: 100%;
+  }
+
+  &[data-orientation="vertical"] {
+    height: 100%;
```
tifa2UP, line 12: height: 100% doesn't actually take 100% height. It takes 100% of the width. That's a limitation of the % usage in css. 

There's another css property for the container height, not sure what it was

src/components/dls/Separator/Separator.tsx
```diff
@@ -0,0 +1,18 @@
...
+import * as SeparatorPrimitive from '@radix-ui/react-separator';
+import styles from './Separator.module.scss';
+
+export enum SeparatorOrientation {
+  Horizontal = 'horizontal',
+  Vertical = 'vertical',
+}
+
+type SeparatorProps = {
+  orientation: SeparatorOrientation;
```
tifa2UP, line 11: Orientation should be optional and default to horizontal.

We'll just have another prop called vertical, when it's passed the component becomes a vertical divider

src/components/dls/Separator/Separator.module.scss
muhajirdev: would love to learn more about this.

tifa2UP (approved):

PR #399, opened 2021-09-08
Add minimizable context menu on scroll

osamasayed:
### Summary
This PR adds a hook to detect the scrolling direction and apply that to the `ContextMenu` inside of the QuranReader.


muhajirdev:
I had some use cases with scroll detection previously.

If I am not, on scroll is more expensive compared to intersection observer

https://itnext.io/1v1-scroll-listener-vs-intersection-observers-469a26ab9eb6

Some libraries that may be useful for this scroll related activities:
- https://github.com/russellgoldenberg/scrollama
- react-scrollama
- react-intersection-observer

muhajirdev:
I think yeah, Intersection Observer is less applicable in this context. Even tho we can technically put the intersection observer into the document / body itself. But it is less straightforward compared to using scroll. And also not sure if it's more performant or not.

Is this context I think it makes more sense to just debounce the on scroll. So it doesn't fires event that many times.

src/hooks/useScrollDirection.ts
```diff
@@ -0,0 +1,46 @@
...
+ * A hook that detects scrolling and determines whether
+ * the user is scrolling up and down and passes it to the
+ * callback.
+ *
+ * @param {direction: ScrollDirection, position: number) => void} onDirectionChange
+ */
+const useScrollDirection = (
+  onDirectionChange: (direction: ScrollDirection, position: number) => void,
+) => {
+  const [, setLastYPosition] = useState<number>(0);
```
muhajirdev, line 18: also,
disclaimer: I don't really have crystal clear picture about the thing I will say. 😁

Does it make sense to use `useRef` here? 

```
const lastYPosition = useRef(0);

// accessing the position 
lastYPosition.current

// set lastYPosition
lastYPosition.current = newYPosition
```

I remember I need to use `useRef` a couple times in the past to deal with heave animation

Essentially, both `useRef` and `useState` can store a value. But the difference is with `useState`, everytime the state is updated. It will trigger `re rendering` with `useRef` it doesn't trigger re rendering. Which is what we want in this context. 

We don't want changing y position trigger re render, we only want to call `onDirectionChange` callback when new direction is detected.

CMIIW. Because I think I don't have as much as context as you for this code 😄 

Reference 
- https://blog.logrocket.com/usestate-vs-useref/

osamasayed: Thanks a lot @muhajirdev for the suggestion. It seems using `useRef` in this case works better than `useState`. Will update the code accordingly.

tifa2UP (approved):

PR #400, opened 2021-09-08
Show translation and tafsir langauges

osamasayed:
### Summary
This PR adds the language of the translation and the Tafsir selectors.

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2021-09-08 at 09 46 44|Screen Shot 2021-09-08 at 09 46 34|
|Screen Shot 2021-09-08 at 09 47 03|Screen Shot 2021-09-08 at 09 46 55|
|Screen Shot 2021-09-08 at 09 47 27|Screen Shot 2021-09-08 at 09 47 20|

### DIFF
--- a/src/utils/input.ts
+++ b/src/utils/input.ts
@@ -1,5 +1,4 @@
 import capitalize from 'lodash/capitalize';
-import { DropdownItem } from 'src/components/dls/Forms/Combobox/ComboboxItem';
 import { RadioItem } from 'src/components/dls/Forms/RadioGroup/RadioGroup';
 import { SelectOption } from 'src/components/dls/Forms/Select';
 import AvailableTranslation from 'types/AvailableTranslation';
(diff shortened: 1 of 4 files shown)


muhajirdev:
looks likt the vercel build failed @osamasayed

src/utils/input.ts
```diff
@@ -1,6 +1,8 @@
 import capitalize from 'lodash/capitalize';
 import { RadioItem } from 'src/components/dls/Forms/RadioGroup/RadioGroup';
 import { SelectOption } from 'src/components/dls/Forms/Select';
+import AvailableTranslation from 'types/AvailableTranslation';
```
muhajirdev, line 4: code looks good to me overall mA.

btw what is the difference between `AvailableTranslation` and `Translation`? looks like we have both in `types/`

osamasayed: `AvailableTranslation` is the list of all supported translations. `Translation` refers to a verse/word translation.

muhajirdev: I see.

out of the scope of this PR.
- shall we make “Translation” more specific? Like VerseTranslation?
- Or is it possible to merge them together, not sure how the structure of the backend api now.


IMO there’s a room for improvement in these naming

src/utils/input.ts
```diff
@@ -26,3 +29,31 @@ export const generateRadioItems = (items: string[]): RadioItem[] =>
...
+export const getTranslatedLabelWithLanguage = (item: TafsirInfo | AvailableTranslation): string =>
+  `${capitalize(item.languageName)} - ${item.translatedName.name}`;
+
+/**
+ * Convert a TafsirInfo or AvailableTranslation items into combobox items.
+ *
+ * @param {TafsirInfo[] | AvailableTranslation[]} items
+ * @returns  {DropdownItem[]}
+ */
+export const itemsToComboboxItems = (
```
tifa2UP, line 48: Have a mixed opinion here for 2 reasons: 

1) the name is misleading, initially I wanted to suggest naming the function "arrayToComboboxItems" but this function is not a generic one, it's only for tafsirs and available translations. Which leads me to 2)

2) Since this function is only for tafsirs and available translations schema, imo it shouldn't be a util. It's ok to have the code duplicated in both files.

osamasayed: Okay no strong opinion here, just wanted to save duplicating the data but can revert no probs.

PR #419, opened 2021-09-13
Add remove recent search query button

osamasayed:
### Summary
This PR adds the ability to remove a recent search query.

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2021-09-13 at 10 52 54|Screen Shot 2021-09-13 at 10 52 47|

### DIFF
--- a/src/components/Navbar/SearchDrawer/PreInput/SearchQuerySuggestion/SearchQuerySuggestion.module.scss
+++ b/src/components/Navbar/SearchDrawer/PreInput/SearchQuerySuggestion/SearchQuerySuggestion.module.scss
@@ -18,7 +18,7 @@
 .removeButton {
   background: none;
   border: none;
-  color: var(--color-primary-medium);
+  color: var(--color-text-default);
   font-size: var(--font-size-small);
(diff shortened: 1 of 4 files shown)


src/components/Navbar/SearchDrawer/PreInput/SearchQuerySuggestion/SearchQuerySuggestion.module.scss
```diff
@@ -12,3 +14,10 @@
     background-color: var(--color-background-alternative-deep);
   }
 }
+
+.removeIcon {
```
muhajirdev, line 18: we're using `.removeIcon` class instead of `<Button>` component because the Button component doesn't support height and width `spacing-small` ? is that the case?

osamasayed: Yes when using Button even for small size it takes too much space also the hovering effect is off

tifa2UP (approved):

PR #421, opened 2021-09-13
add `RootState` typing for redux 

muhajirdev:
TSIA


muhajirdev:
removed `usedTypedSelector`, as we always use `select*` function. For example `selectIsUsingDefaultTranslations`

And we never use inline selector like this
```javascript 
useSelector( (state) => state.translations.isUsingDefaultTranslations; ) 
```

package.json
```diff
@@ -55,7 +55,8 @@
     "react-infinite-scroller": "^1.2.4",
     "react-redux": "^7.2.4",
     "redux-persist": "^6.0.0",
-    "swr": "^1.0.1"
+    "swr": "^1.0.1",
+    "typesafe-actions": "^5.1.0"
```
tifa2UP, line 59: this should be a dev dependency?

muhajirdev: correct

PR #422, opened 2021-09-13
Alert audio mistmatch on click "play" in audio player

muhajirdev:
TSIA


src/utils/api.ts
```diff
@@ -4,7 +4,7 @@ import { stringify } from 'qs';
 export const ITEMS_PER_PAGE = 10;
 const DEFAULT_MUSHAF = 4; // King Fahad Quran Complex 15 Lines Hafs text.
 
-const STAGING_API_HOST = 'https://staging.quran.com/api/qdc';
+const STAGING_API_HOST = 'https://api.quran.com/api/qdc';
```
muhajirdev, line 7: This is just for testing, will revert this back to staging when we're about to merge the PR

src/components/AudioPlayer/PlayPauseButton.tsx
```diff
@@ -1,24 +1,47 @@
...
 import { triggerPauseAudio, triggerPlayAudio } from './EventTriggers';
+import SurahAudioMismatchModal from './SurahAudioMismatchModal';
 
 const PlayPauseButton = () => {
+  const dispatch = useDispatch();
+
   const { isPlaying } = useSelector(selectAudioPlayerState, shallowEqual);
-  const audioFileStatus = useSelector(selectAudioFileStatus);
-  const isLoading = audioFileStatus === AudioFileStatus.Loading;
+  const isLoading = useSelector(selectAudioFileStatus) === AudioFileStatus.Loading;
```
osamasayed, line 25: should we add `shallowEqual` here? if this PR doesn't go before the `RootState` one.

src/components/AudioPlayer/PlayPauseButton.tsx
```diff
@@ -1,24 +1,47 @@
...
+
+  const audioFile = useSelector(selectAudioFile);
```
osamasayed, line 27: ditto

src/hooks/useChapterId.ts
```diff
@@ -12,7 +12,7 @@ import { getChapterIdsForJuz, getChapterIdsForPage } from 'src/utils/chapter';
  * - /page/2 => ["2"]
  * - /chapter/1 => ["1"]
  */
-const useChaptersIdByUrlPath = (): string[] => {
+const useChaptersIdsByUrlPath = (): string[] => {
```
osamasayed, line 15: I think this should be `useChapterIdsByUrlPath`

src/components/dls/Modal/Modal.tsx
```diff
@@ -43,7 +70,7 @@ const Subtitle = ({ children }) => (
 const Footer = ({ children }) => <div className={styles.footer}>{children}</div>;
 type ActionProps = {
   children: React.ReactNode;
-  onClick?: () => void;
+  onClick?: (e) => void;
```
osamasayed, line 73: Should we add the type of the event here?

src/components/dls/Modal/Modal.tsx
```diff
@@ -9,8 +9,16 @@ type ModalProps = {
...
+  shouldStopPropagation?: boolean;
 };
-const Modal = ({ children, trigger, open, onClickOutside, triggerClassName }: ModalProps) => (
+const Modal = ({
+  children,
+  trigger,
+  open,
+  onClickOutside,
+  triggerClassName,
+  shouldStopPropagation = true,
```
osamasayed, line 20: since we set `shouldStopPropagation` to `true`, it will affect any component that is using `Modal`. Would that affect the functionality of the other components or all good?

src/components/AudioPlayer/PlayPauseButton.tsx
muhajirdev: yep I think this PR will go before RootState. 

For audioFileStatus I think this is okay, it will transpiled into primitive (tested with typescript repl.it) CMIIW

src/components/AudioPlayer/PlayPauseButton.tsx
muhajirdev: yes, thanks for the comment.

src/hooks/useChapterId.ts
muhajirdev: ✅

src/components/dls/Modal/Modal.tsx
muhajirdev: ✅

src/components/dls/Modal/Modal.tsx
muhajirdev: tested with other modal (advanced copy).

So the only effect is the `onClick` event will not bubble up to the parent of React component
- in the context of this PR, it will not bubble up to audioplayer. So it won't trigger audio expand
- in Advanced copy context, the onClick will not bubble up to the popover menu.

So, all good I think

tifa2UP: It sounds to me that shouldStopPropagation is not part of the modal scope. It should be handled in the audio player. What do you think?

src/utils/api.ts
tifa2UP: Sounds good, let's revert this before merging iA.

tifa2UP (changes requested):
Noticed some bugs in the juz and page routes. Pinged on discord

muhajirdev:
fixed the bug

osamasayed (approved):

PR #425, opened 2021-09-14
Reduce Lodash bundle size and add bundle analyzer

osamasayed:
### Summary

[Webpack Tree-shaking](https://webpack.js.org/guides/tree-shaking/) is the process of removing unused (dead) code during the bundling process which results in a smaller bundle size. 

To analyze the bundle size and investigate if tree-shaking happens and which dependency/code takes the largest space in a bundle, [`webpack-bundle-analyzer`](https://github.com/webpack-contrib/webpack-bundle-analyzer) plugin can be used. A Next.js version of the plugin called [`@next/bundle-analyzer`](https://github.com/vercel/next.js/tree/canary/packages/next-bundle-analyzer) which supports using the analyzer for Next.js apps was used. 

After analyzing the client bundle size, turns out that out of all the dependencies , [`Lodash`](https://lodash.com) was contributing the most to the bundle size: 

| Size | Description |
| ------ | ------ |
|75.98 KB|parsed|
|25.69 KB|zipped|

Before 1

The reason for this was that we were importing the entire library using the following syntax: 

`import _ from 'lodash';` 

while in most cases we were only using 1 function out of the whole library inside our components e.g. 

`{_.capitalize(chapter.revelationPlace)}`

`import _ from 'lodash';`is not tree-shakable because to tree-shake, Webpack relies on the static structure of ES2015 module syntax, i.e. `import` and `export`. which means how we import the package is what matters not how we are using it. So even though we were using only 1 function but since we are importing the whole library, Webpack wasn't doing tree-shaking since we were telling it "hey, we need the entire package". In general there were 3 ways to over-come this and reduce the bundle size:

1. Using [`lodash/babel-plugin-lodash`](https://github.com/lodash/babel-plugin-lodash) babel transformer which would cherry-pick Lodash modules automatically so would transform:
```
import _ from 'lodash'
...
<p>{_.capitalize(chapter.revelationPlace)}</p>
...
```
into:
```
import _capitalize from 'lodash/capitalize'
...
<p>{_capitalize(chapter.revelationPlace)}</p>
...
```
2. Using destructing importing syntax e.g. 
`import { capitalize } from 'lodash';` but that would require adding [Webpack optimization](https://webpack.js.org/configuration/optimization/#optimizationusedexports) for tree-shaking to work. Otherwise it will still load the entire library.
3. Directly importing Lodash modules on demand e.g. 
`import capitalize from 'lodash/capitalize';`

In this PR I took the 3rd route as it doesn't require any extra configs and it's the most readable way of importing which reduced Loadash's bundle size by ~75% down to:

| Before | After | Description |
| ------ | ------ | ------ |
|75.98 KB|18.3 KB|parsed|
|25.69 KB|5.61 KB|zipped|

Resources:
- https://developers.google.com/web/fundamentals/performance/optimizing-javascript/tree-shaking
- https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/
- https://stackoverflow.com/questions/58741044/why-webpack-doesnt-tree-shake-the-lodash-when-using-import-as
- https://stackoverflow.com/questions/43479464/how-to-import-a-single-lodash-function
- https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

### Screenshots

| Before | After | Description |
| ------ | ------ | ------ |
| Before 1 | After 1
|Before 2|After 2 |
|Before 3|After 3||
|Before 4|After 4|On development environment, the size of `_app` and `index` bundles have gone down|


src/redux/slices/QuranReader/bookmarks.ts
```diff
@@ -39,6 +38,6 @@ export const { toggleVerseBookmark } = bookmarksSlice.actions;
 export const selectBookmarks = (state) => state.bookmarks;
 export const selectOrderedBookmarkedVerses = (state) =>
   // sort the bookmarked verses by the order they appear in the Mushaf.
-  _(state.bookmarks.bookmarkedVerses).toPairs().sortBy(0).fromPairs().value();
+  Object.fromEntries(Object.entries(state.bookmarks.bookmarkedVerses).sort());
```
muhajirdev, line 41: nice, without lodash haha 👍

muhajirdev:
Looks good to me.

Yeah, looks like the`en.json` adds a bit more size to bundle. I think we can reduce the file size. The are things we don't really need in the json I think. Like 

```
"revelationPlace": "makkah",
"revelationOrder": 5,
```

package.json
```diff
@@ -16,14 +16,16 @@
     "generate-sitemap": "node sitemap-generator.js",
     "test": "jest",
     "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "analyze": "ANALYZE=true yarn build"
```
tifa2UP, line 20: nit: ANALYZE_BUNDLE.

package.json
```diff
@@ -16,14 +16,16 @@
...
   },
   "jest-runner-eslint": {
     "cliOptions": {
       "ignorePath": "./.gitignore"
     }
   },
   "dependencies": {
+    "@next/bundle-analyzer": "^11.1.2",
```
tifa2UP, line 28: This can't be a devDependcy, right?

osamasayed: No we cannot it needs to go with `dependencies`

PR #435, opened 2021-09-15
Upgrade Radix UI Primitives

osamasayed:
### Summary
Upgrade `radix-ui` primitives to the [latest version](https://www.radix-ui.com/docs/primitives/overview/releases) which contain some breaking changes.


src/components/dls/Modal/Modal.tsx
```diff
@@ -15,10 +15,10 @@ const Modal = ({ children, trigger, open, onClickOutside, triggerClassName }: Mo
     <DialogPrimitive.Overlay className={styles.overlay} />
     {trigger && (
       <DialogPrimitive.Trigger
-        as="div"
+        asChild
```
muhajirdev, line 18: interesting, so with `asChild` it will not render this div? it will only render the child?

tifa2UP (approved):

PR #442, opened 2021-09-15
Misc design changes

tifa2UP:
### Summary
Miscellaneous design changes based on various feedback. 
- Fix the clear button not appearing on dark mode
- Add a container utility that has 100% width on mobile, varying length based on desktop width.
- Use the container utility for the Quran reader. (this resolves the feedback request to reduce the side gaps on mobile) 
- Reduce several paddings in the Quran reader as requested by Aqeel.
- Fix a bug in the surah header where the Q icon used to appear above the surah name text (modified the z-index) 
- Add padding under the bismillah SVG in the surah header

### Known Issues
The reading mode currently doesn't have any padding in the surah header. Working on resolving it.

### Screenshots

| Before | After |
| ------ | ------ |
| 2021-09-16 at 00 20 41 | 2021-09-16 at 00 22 28 |

### DIFF
--- a/src/components/QuranReader/QuranReader.module.scss
+++ b/src/components/QuranReader/QuranReader.module.scss
@@ -1,8 +1,12 @@
 @use "src/styles/constants";
 @use "src/styles/breakpoints";
+@use "src/styles/utility";
+
+$quran-reader-padding-mobile: 4rem;
+$quran-reader-padding-tablet: 5rem;
 
 .infiniteScroll {
-  max-width: 95%;
+  @include utility.container();
   margin: 0 auto;
 }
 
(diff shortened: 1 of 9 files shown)


src/components/QuranReader/QuranReader.module.scss
```diff
@@ -1,8 +1,12 @@
 @use "src/styles/constants";
 @use "src/styles/breakpoints";
+@use "src/styles/utility";
+
+$quran-reader-padding-mobile: 4rem;
+$quran-reader-padding-tablet: 5rem;
 
 .infiniteScroll {
-  max-width: 95%;
+  @include utility.container();
```
muhajirdev, line 9: nit: does `container` needs `()`?, since it doesn't accept params.
in other utility, such as `mobileM` we don't have `()`. 

I am not super familiar with scss btw, so maybe I am missing something.

osamasayed: +1

tifa2UP: Fixed. Thank you Muhajir!

PR #446, opened 2021-09-16
Fix storybook deployment + add  storybook deploy preview url in PR's comment

muhajirdev:
# Summary

### Fix storybook deployment fail

  - We deploy storybook by pushing the storybook build artifact into `gh-pages` branch in this repository
  - Previously the storybook deployment failed because, we run github action twice `on: [push, pull request]`.
  - So sometime there's a race condition, between the 2 actions. Where both is trying to execute `Commit updates`, both action tried to commit and push to same branch at the same time. (see `build-storybook.yml` for the details
  - So github rejected one the `push`
    
```
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
```
 
 - now, we only run the action once `on: [pull request]`

### Adding storybook URL for the PR

- Create a PR comment `Storybook preview : *In-Progress`, when PR is created
- Then update the comment into `Storybook preview: [storybook-url]` when the storybook deployment is finished

PR test : #452 

### Known issues
- There's some code duplication (`find-comment`). 
I decided to do just duplicate code, after a few iterations few with "without duplication" alternative (by using github actions' output and yaml `merge` feature.  The other approaches is less familiar and adding complexity to the code.  

The amount of duplication is still manageable, and they sit close to each other. I think this is okay.

### DIFF
--- a/.github/workflows/build-storybook.yml
+++ b/.github/workflows/build-storybook.yml
@@ -62,7 +62,7 @@ jobs:
           body-includes: "Storybook preview:"
       - name: Save comment id
         id: comment-id
-        run: echo "::set-output name=comment-id::446"
+        run: echo "::set-output name=comment-id::${{ steps.find-comment.outputs.comment-id }}"
       - name: Create comment
         if: steps.find-comment.outputs.comment-id == '' # only create new comment, if no previous storybook comment found
         uses: peter-evans/create-or-update-comment@v1
--- a/.github/workflows/build-storybook.yml
+++ b/.github/workflows/build-storybook.yml
@@ -74,9 +74,8 @@ jobs:
 
   post-storybook-url:
     runs-on: ubuntu-latest
-    needs: create-storybook-comment
+    needs: [create-storybook-comment, deploy-storybook]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-storybook.yml
```diff
@@ -47,3 +47,42 @@ jobs:
           GIT_AUTHOR_NAME: ${{ github.actor }}
           GIT_COMMITTER_EMAIL: ${{ github.actor }}@users.noreply.github.com
           GIT_COMMITTER_NAME: ${{ github.actor }}
+
+  create-storybook-comment:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Find Comment
```
osamasayed, line 54: Can we try to put `find-comment` in a block and re-use it since we are duplicating the code. You can have a look at https://stackoverflow.com/questions/8466223/reuse-a-block-of-code-in-yaml

muhajirdev: thanks for the suggestion @osamasayed 👍 

we found a way to share the output between jobs, so use that now.

PR #451, opened 2021-09-16
Fix issues related to advanced copy

muhajirdev:
### Summary
- Call the API ahead of time, so the data is already ready when we click "copy". This solve the copying problem on iOS
- move some of the function into separate file `advanced-copy-helper`
- fix UI issue on dark theme
- fix can't download the txt file on advanced copy
- update API to use `mushaf` instead of `fields` (this might not work in next.quran.com with api.quran.com until the API is fixed)

## UI issues
Before

image

After
image


muhajirdev:
# The copying issues

#### The problem
This issue only exist in safari, both mobile and desktop. The problem is safari only allow copying during the user gesture / click. When we call backend there's about +- 1 second delay between the "click" event and the actually copy. 

image
image
image

There's a library that implement a fallback for this constraint. They do it like this. So when we click copy, they will show this browser prompt. Then we have to press cmd + c by ourself. Which is a pretty bad UX
image

### The possible solution

Fetch the data ahead of time, so the data is ready when we click "copy". So, no delay between "pointerdown" event and calling "clipboardCopy". 

The cons of this approach are:
- if the user doesn't actually copy the verse. That was a waste of API call
- when user select "range of verses". We do another API call, which means the first API call was a wasted call.
one side note: when user click "current verse" again. We don't do additional API call because it is already cached by `useSWR`

The pros:
- When user click "copy", there is no delay between the click and the response. Which is good for UX if we take the `100ms rule` into account 
image

#### In the big picture,

We are actually already doing this kind of data loading. / "prefetching". If we go to next.quran.com, and open `network` tab in dev tools. We'll see the number of network request grows as we scroll down the page.

image

The reason is, nextjs is actually prefetching the data for us. So once we click one of the page, it feels the page opend "**instantly**". Because the data is already prefetched.

So technically we already have a lot of "wasted fetch". It's a tradeoff for the UX.

---
Side note
- There's actually even an effort to predict what user will click : [guess.js](https://guess-js.github.io/)
  https://www.youtube.com/watch?v=0jB4YWgAxUo

src/components/Verse/AdvancedCopy/VerseAdvancedCopy.tsx
```diff
@@ -183,56 +153,24 @@ const VerseAdvancedCopy: React.FC<Props> = ({ verse, children }) => {
...
+      shouldCopyText,
+      translations,
+      verseKey: verse.verseKey,
+    })
+      .then((blob) => {
+        setIsLoadingData(false);
+        setObjectUrl(window.URL.createObjectURL(blob));
+      })
+      .catch((error) => {
+        setCustomMessage(error);
```
osamasayed, line 171: I don't think we need to be showing a technical error to the user.

src/components/Verse/AdvancedCopy/utils/copyText.ts
```diff
@@ -0,0 +1,71 @@
...
+  const getTextBlobPromise = () =>
+    getTextToCopy({
+      rangeEndVerse,
+      rangeStartVerse,
+      shouldCopyFootnotes,
+      shouldCopyText,
+      showRangeOfVerses,
+      translations,
+      verseKey,
+    }).then((text) => new Blob([text], { type: 'text/plain' }));
```
osamasayed, line 40: We need to move this to a global util function that can be re-used. e.g. `textToBlob`

src/components/Verse/AdvancedCopy/utils/copyText.ts
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * Safari needs `navigator.clipboard.write` to be called immediately without waiting
+   * Chrome can wait
+   *
+   * So, for safari we call it navigator.clipboard immediately and give it a promise
+   * for chrome we wait for the promise to resolve, the wen call navigator.clipboard.write
+   */
+
+  try {
+    copyOnSafari(textBlobPromise);
```
osamasayed, line 56: We need to move this to a global util function that can be re-used.

osamasayed: Also we should rename `copyOnSafari` since the way we are using is not only Safari specific but rather supports multiple browsers

src/components/Verse/AdvancedCopy/utils/copyText.ts
```diff
@@ -0,0 +1,71 @@
...
+    return textBlobPromise;
+  } catch (e) {
+    const blob = await textBlobPromise;
```
osamasayed, line 59: I wonder why we need to do this since [`Navigator.clipboard`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) should be working on Chrome too according to the Browser compatibility table of that API. Have you tried what is recommened [here](https://stackoverflow.com/questions/57278923/chrome-76-copy-content-to-clipboard-using-navigator)  or [`Clipboard.writeText()`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) instead of having to convert the text into a blob then write it?

osamasayed (approved):

PR #455, opened 2021-09-16
Add ease-in and ease-out transitions for the drawer

tifa2UP:
### Summary
This improves the drawer animation feel: https://www.joshwcomeau.com/animation/css-transitions/


muhajirdev:
Looks, good 👍 

btw, we in Modal/_constant.scss we have an easing variable (taken from vercel). I don’t know if we want to standardize the easing or not.

PR #465, opened 2021-09-19
Add Spinner for Infinite scroll

osamasayed:
### Summary
This PR adds a loading spinner instead of text when the user scrolls to the end of the Quran Reader and there are more verses to be fetched.

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2021-09-19 at 14 30 10|Screen Shot 2021-09-19 at 14 29 20|

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -130,9 +130,13 @@ const QuranReader = ({
             loadMore={loadMore}
             loader={
               <div className={styles.loadMoreContainer} key={0}>
-                <button type="button" onClick={loadMore} disabled={isValidating}>
-                  {isValidating ? <Spinner size={SpinnerSize.Large} isCentered /> : 'Load More ...'}
-                </button>
+                {isValidating ? (
+                  <Spinner size={SpinnerSize.Large} />
+                ) : (
+                  <button type="button" onClick={loadMore} disabled={isValidating}>
+                    Load More...
+                  </button>
+                )}
               </div>
             }
           >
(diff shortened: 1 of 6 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -133,7 +131,7 @@ const QuranReader = ({
             loader={
               <div className={styles.loadMoreContainer} key={0}>
                 <button type="button" onClick={loadMore} disabled={isValidating}>
```
muhajirdev, line 133: quick question, a bit unrelated to the changes in this PR.

is there a reason we make this a button?

src/components/dls/Spinner/Spinner.module.scss
```diff
@@ -16,6 +16,10 @@
   box-sizing: border-box;
 }
 
+.centered {
```
muhajirdev, line 19: another quick question 😁.
what if we make this styling on the parent component. Is it more complicated in this context? or shall we just make it centered by default?

I don't have strong opinion on this

src/components/QuranReader/index.tsx
tifa2UP: +1, I'd imagine that the spinner would not be interactive.

tifa2UP (approved):

src/components/dls/Spinner/Spinner.tsx
```diff
@@ -9,14 +9,16 @@ export enum SpinnerSize {
 }
 type SpinnerProps = {
   size?: SpinnerSize;
+  isCentered?: boolean;
 };
 
-const Spinner = ({ size = SpinnerSize.Medium }: SpinnerProps) => (
+const Spinner = ({ size = SpinnerSize.Medium, isCentered = false }: SpinnerProps) => (
```
tifa2UP, line 15: it should be centered by default, no?

tifa2UP (approved):

tifa2UP (approved):

PR #471, opened 2021-09-20
Listen to currentTime and update the highlightedLocation in redux

muhajirdev:
TSIA


src/utils/apiPaths.ts
```diff
@@ -45,8 +45,9 @@ export const makeLanguagesUrl = (language: string): string =>
  * @returns {string}
  */
 export const makeRecitersUrl = (): string => makeUrl('/audio/reciters');
-export const makeAudioFilesUrl = (reciterId: number, chapter: number) =>
-  makeUrl(`/audio/reciters/${reciterId}`, { chapter });
+
+export const makeAudioFilesUrl = (reciterId: number, chapter: number, segments: boolean) =>
```
muhajirdev, line 49: So we are actually reusing the same API path.

We might need to refactor the data flow to get audioFile and timestamps data. So they can be fetched at once.

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useRef } from 'react';
```
muhajirdev, line 1: This component can probably just become a hook, since it doesn't really do any rendering

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
```diff
@@ -0,0 +1,57 @@
...
+    getChapterAudioFile(reciterId, chapterId, true),
+  );
+
+  const lastHighlightedLocation = useRef(defaultHighlightedLocation);
+
+  // eslint-disable-next-line consistent-return
+  useEffect(() => {
+    if (!data) return;
+
+    const highlightedLocation = getHighlightedLocation(
```
muhajirdev, line 42: I think this still has room for performance improvement, we can do something like this

```javascript
// eslint-disable-next-line consistent-return
  useEffect(() => {
    if (!data) return;
    const currentTimeInMs = currentTime * 1000;

    let currentHighlightedVerse;
    if (
      lastHighlightedVerse.current &&
      currentTimeInMs >= lastHighlightedVerse.current.timestampFrom &&
      currentTimeInMs <= lastHighlightedVerse.current.timestampTo
    ) {
      currentHighlightedVerse = lastHighlightedVerse.current;
    } else {
      currentHighlightedVerse = data.verseTimings.find(
        (verse) => currentTimeInMs >= verse.timestampFrom && currentTimeInMs <= verse.timestampTo,
      );
      lastHighlightedVerse.current = currentHighlightedVerse;
    }

    const highlightedWordLocation = getHighlightedWordLocation(
      currentTimeInMs,
      currentHighlightedVerse,
    );
```

muhajirdev: so, it only do verseTimings.find when the condition doesn't match.

muhajirdev: we can also have `lastHighlightedWordLocation`

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
```diff
@@ -0,0 +1,57 @@
...
+    );
+
+    // only update redux value if current highlight value does not equal prev value
+    if (!shallowEqual(highlightedLocation, lastHighlightedLocation.current)) {
+      dispatch(setHighlightedLocation(highlightedLocation));
+      lastHighlightedLocation.current = highlightedLocation;
+    }
+  }, [data, dispatch, currentTime]);
+
+  return null;
```
tifa2UP, line 54: no need to return null here

src/components/AudioPlayer/getHighlightedLocation.ts
```diff
@@ -0,0 +1,41 @@
...
+    (verse) => currentTime >= verse.timestampFrom && currentTime <= verse.timestampTo,
+  );
+
+  if (!selectedVerse) return defaultHighlightedLocation;
+  const selectedWord = selectedVerse.segments.find((segment) => {
+    const [, timestampFrom, timestampTo] = segment; // the structure of the segment is: [wordLocation, timestampFrom, timestampTo]
+    return currentTime >= timestampFrom && currentTime <= timestampTo;
+  });
+
+  if (!selectedWord) return defaultHighlightedLocation;
```
tifa2UP, line 30: this is what needs to change, handle the case where the verse is highlighted but there's no word to highlight.

src/api.ts
```diff
@@ -86,13 +86,21 @@ export const getAvailableReciters = async (): Promise<RecitersResponse> =>
...
+ * additionally you can pass `segment: true` to get the timestamps
+ * for each verse and words
  *
  * @param {number} reciterId
  * @param {number} chapter the id of the chapter
  */
 
-export const getAudioFile = async (reciterId: number, chapter: number): Promise<AudioFile> => {
-  const res = await fetcher<AudioFilesResponse>(makeAudioFilesUrl(reciterId, chapter));
+export const getChapterAudioFile = async (
```
tifa2UP, line 96: getChapterAudioData?

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
tifa2UP: +1, I actually have a similar proposal. Mentioning it in the loom.

src/api.ts
muhajirdev: +1 , yes. `Data` is actually more correct. Since we're not actually fetching the `File`

If we agree on this naming, we should also update other naming that mention `audioFile`, since they are also referring to the same value. For example `types/AudioFile` -> `types/AudioData`.

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
muhajirdev: I think React will be angry if we don't return null here?

image

muhajirdev:
image

src/redux/slices/QuranReader/highlightedLocation.ts
```diff
@@ -0,0 +1,38 @@
...
+export type HighlightedLocationState = {
+  chapter: number | null;
+  verse: number | null;
+  word: number | null;
+};
+
+export const initialState: HighlightedLocationState = {
+  chapter: null,
+  verse: null,
+  word: null,
```
tifa2UP, line 14: highlightedChapter: 
highlightedVerse:
highlightedWord: 

—-
const {highlihgtedChapter} = useSelector(...)

src/components/AudioPlayer/HighlightedLocationUpdater.tsx
```diff
@@ -0,0 +1,146 @@
...
+  reciterId,
+  chapterId,
+  currentTime,
+}: AudioTimestampsHighlightListenerProps) => {
+  const dispatch = useDispatch();
+  const { data: audioFileData } = useSWRImmutable(
+    makeChapterAudioFiles(reciterId, chapterId, true),
+    () => getChapterAudioFile(reciterId, chapterId, true),
+  );
+  const currentTimeInMs = currentTime * 1000;
```
tifa2UP, line 39: fix undefined bug

tifa2UP (approved):

PR #486, opened 2021-09-23
Update highlighting logic

muhajirdev:


src/components/AudioPlayer/QuranReaderHighlightDispatcher.tsx
```diff
@@ -104,15 +107,18 @@ const useMemoizedHighlightedWordLocation = (
   const lastHighlightedWordLocation = useRef<Segment>(null);
   if (!currentHighlightedVerseTiming) return null;
 
+  // do not highlight verse when currentTime is 0
```
tifa2UP, line 110: // Do not highlight the verse when the audio hasn't started playing
if (currentTime === 0) return defaultState/initialState;

src/components/AudioPlayer/QuranReaderHighlightDispatcher.tsx
```diff
@@ -146,4 +152,20 @@ const getHighlightedLocation = (
...
+ * - timestampFrom = 10, timestampTo = 20, currentTime = 20 should return false
+ *
+ * @param {number} currentTime
+ * @param {number} timestampFrom
+ * @param {number} timestampTo
+ * @returns {boolean} isWithinRange
+ */
+const isCurrentTimeInRange = (currentTime: number, timestampFrom: number, timestampTo: number) =>
+  currentTime >= timestampFrom && currentTime < timestampTo;
+
```
tifa2UP, line 170: IMO, this is not a big enough code to be a function, it's probably more readable inline. 

2021-09-23 at 08 47 49@2x

Not a strong opinion though, I'll leave the decision to you :)

src/components/AudioPlayer/QuranReaderHighlightDispatcher.tsx
tifa2UP: Also, let's do this check in the top part of the file.

This way we can short circuit the unnecessary code.

src/components/AudioPlayer/QuranReaderHighlightDispatcher.tsx
muhajirdev: yeah, this function is actually very simple.

But since this is a function, we add docs and give examples. Which helps to make the coder is easier to understand IMO.

If we duplicate the docs in 4 places, it's a bit less clean.

So I will keep this function for now

src/components/AudioPlayer/QuranReaderHighlightDispatcher.tsx
muhajirdev: 👍  updated the docs to be more clear.

and this function needs to return `null | VerseTiming`. The final result will be `defaultState` when this function returns null 👍

tifa2UP (approved):

PR #489, opened 2021-09-23
TranslationViewCell verse play/pause button

muhajirdev:
### Summary
Make the play button in TranslationViewCell render conditionally (play or pause). Based on highlighedLocation data

## Notes
- first commit is simpler but without "loading support"
- second commit is with loading support but contains more changes


src/redux/slices/AudioPlayer/state.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
muhajirdev, line 1: need to breakdown this file into separate line, but will do it if this PR is approved.

tifa2UP (approved):

PR #493, opened 2021-09-25
Fix broken hrefs in the link component

tifa2UP:
### Summary
Our link components didn't work before. Passing an href (like in the homepage welcome message) caused the href on the anchor tag to be null making it uninteractive.

### DIFF
--- a/src/components/dls/Link/Link.tsx
+++ b/src/components/dls/Link/Link.tsx
@@ -23,7 +23,7 @@ type LinkProps = {
 const Link: React.FC<LinkProps> = ({ href, children, newTab = false, variant, download }) => (
   <Wrapper shouldWrap={!download} wrapper={(node) => <NextLink href={href}>{node}</NextLink>}>
     <a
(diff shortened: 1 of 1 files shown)


src/components/dls/Link/Link.tsx
```diff
@@ -23,7 +23,7 @@ type LinkProps = {
 const Link: React.FC<LinkProps> = ({ href, children, newTab = false, variant, download }) => (
   <Wrapper shouldWrap={!download} wrapper={(node) => <NextLink href={href}>{node}</NextLink>}>
     <a
-      href={download ? href : undefined}
```
muhajirdev, line 26: hmmm, interesting. Nextjs Link wants `href` inside the `a`. 
If this is the case maybe we can 
- just do `href={href}` directly 
- or use `passHref` in `NextLink` not sure if this one will works or not

tifa2UP: gonna do href={href} this is what makes sense to me as well

PR #498, opened 2021-09-25
New Audio player follow-up design polish

tifa2UP:
### Summary
* Reduce the height of the player on desktop and mobile (currently uses negative margin, the time and button should be combined into a single component) 
* Use total time instead of remaining time

### Mobile Screenshots

| Before | After |
| ------ | ------ |
| 2021-09-25 at 16 49 25 | 2021-09-25 at 16 49 38 |

### Desktop Screenshots

| Before | After |
| ------ | ------ |
| 2021-09-25 at 16 48 43 | 2021-09-25 at 16 47 52 |

### DIFF
--- a/src/components/AudioPlayer/AudioPlayerSlider.tsx
+++ b/src/components/AudioPlayer/AudioPlayerSlider.tsx
@@ -7,12 +7,11 @@ import { triggerSetCurrentTime } from './EventTriggers';
 import useAudioPlayerCurrentTime from './hooks/useCurrentTime';
 
 import Slider from 'src/components/dls/Slider';
-import { secondsFormatter, milliSecondsToSeconds } from 'src/utils/datetime';
+import { secondsFormatter } from 'src/utils/datetime';
 
 const NUMBER_OF_STEPS = 100;
 
 type SliderProps = {
-  audioDuration: number;
   isMobileMinimizedForScrolling: boolean;
   audioPlayerElRef: React.MutableRefObject<HTMLAudioElement>;
 };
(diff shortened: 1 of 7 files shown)


src/components/AudioPlayer/AudioPlayerSlider.tsx
```diff
@@ -67,7 +70,7 @@ const AudioPlayerSlider = ({
           max={NUMBER_OF_STEPS}
         />
       </div>
-      <span className={styles.remainingTime}>{secondsFormatter(remainingTime)}</span>
+      <span className={styles.remainingTime}>{secondsFormatter(audioDuration)}</span>
```
osamasayed, line 73: This causes the total time to be very bloated since `audioDuration` is in milliseconds and inside 
`secondsFormatter` we multiple it by 1000 again assuming that it's in seconds (since `currentTime` is in seconds). We have a couple of options here:

1. Keep using `audioDuration` but add a param inside `secondsFormatter` called `isInSeconds` (maybe also rename it to `timeFormatter` or something) to indicate whether the time that was passed to the function is in milliseconds or seconds format and when it's in milliseconds, the following line:

`let formattedTime = new Date(seconds * 1000).toISOString().substr(12, 7);`

should be replaced by:

`let formattedTime = new Date(isInSeconds? time * 1000: time ).toISOString().substr(12, 7);`

2. Use `audioDurationMilliSeconds` instead of `audioDuration` because `audioDurationMilliSeconds` is in seconds already. (Either way`audioDurationMilliSeconds` is a typo that I fixed in the smooth slider PR that wasn't merged and should be renamed to `audioDurationSeconds` because the value has been converted to seconds already.)

Option 2 is slightly worse because we have the value converted already from milliseconds to seconds then inside `secondsFormatter` we will convert it back to milliseconds.

PR #503, opened 2021-09-27
Audio repeat UI

muhajirdev:
### Summary
Prepare the UI for Audio Repeat feature

image

# Notes
- when clicking "repeatVerse" from translationview cell, the model will adapt to the current page's chapterId, instead of current playing audio's chapter id.
- when user is clicking `repeat verse` from there, they probably want to interact with that chapter.


src/components/AudioPlayer/PlaybackControls.tsx
```diff
@@ -1,32 +1,30 @@
...
+import {
+  AudioFileStatus,
+  selectAudioFile,
+  selectAudioFileStatus,
+} from 'src/redux/slices/AudioPlayer/state';
 import { withStopPropagation } from 'src/utils/event';
 
 const PlaybackControls = () => {
   const audioFileStatus = useSelector(selectAudioFileStatus);
+  const audioFile = useSelector(selectAudioFile);
```
tifa2UP, line 22: this check should happen inside of the repeat button to determine if it's going to be disabled.

src/components/AudioPlayer/RepeatButton.tsx
```diff
@@ -0,0 +1,131 @@
+import { useMemo, useState, useEffect } from 'react';
+
+import RepeatIcon from '../../../public/icons/ic_repeat_24px 1.svg';
```
tifa2UP, line 3: nit, this was like this from before, can we rename the svg to repeat.svg

src/components/AudioPlayer/RepeatButton.tsx
```diff
@@ -0,0 +1,131 @@
...
+      name: chapterVersesKey,
+      value: chapterVersesKey,
+      label: chapterVersesKey,
+    }));
+
+    return initialState;
+  }, [chapterId]);
+
+  // eslint-disable-next-line prefer-destructuring
+  const firstVersesRangeItems = rangeVersesItems[0];
```
tifa2UP, line 39: not sure what range items are, what does the variable represent.

src/components/AudioPlayer/RepeatAudioModal/RepeatAudioModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+  // reset repeatVerseRange's `to` and `from`, when chapter changed
+  useEffect(() => {
+    setRepeatVerseRange((prevRepeatVerseRange) => ({
+      ...prevRepeatVerseRange,
+      from: firstVerseKeyInThisChapter,
+      to: lastVerseKeyInThisChapter,
+    }));
+  }, [chapterId, firstVerseKeyInThisChapter, lastVerseKeyInThisChapter]);
+
+  const onClickPlay = () => {
```
tifa2UP, line 69: onPlayClick

src/components/AudioPlayer/RepeatAudioModal/RepeatAudioModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+    onClose();
+  };
+  const onClickCancel = () => {
```
tifa2UP, line 72: onCancelClick

src/components/AudioPlayer/RepeatAudioModal/RepeatAudioModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+    onClose();
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClickOutside={onClose}>
```
tifa2UP, line 77: outsideClick is technically not following the convention. 

https://stackoverflow.com/questions/56051358/function-name-onbuttonclick-or-onclickbutton/56053062

src/components/AudioPlayer/RepeatAudioModal/SelectType.tsx
```diff
@@ -0,0 +1,79 @@
...
+    label: 'Single Verse',
+  },
+  {
+    value: RepeatType.Range,
+    id: RepeatType.Range,
+    label: 'Range of verses',
+  },
+];
+
+const SelectType = ({
```
tifa2UP, line 26: Type generally indicates a TS Type, we should use a different name for the component.

src/components/Verse/AdvancedCopy/SelectorContainer.tsx
```diff
@@ -38,6 +38,7 @@ const SelectorContainer: React.FC<SelectorProps> = ({ type, value, dropdownItems
     <Combobox
       id={type}
       value={value}
+      clearable={false}
```
tifa2UP, line 41: tested this, and it seems that clearable allows clearning (?) by typing back space. Out of scope of this PR, but fyi.

src/components/AudioPlayer/RepeatAudioModal/RepeatAudioModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+            minValue={1}
+            onChange={(val) => setVerseRepetition({ ...verseRepetition, repeatRange: val })}
+            suffix="times"
+          />
+          <RepeatSetting
+            label="Delay between verse"
+            value={delayMultiplierBetweenVerse}
+            minValue={0}
+            onChange={(val) => setDelayMultiplierBetweenVerse(val)}
+            suffix="times"
```
naveed-ahmad, line 114: This should be seconds, step 1. A dropdown would be better here IMO.

tifa2UP (approved):

src/components/AudioPlayer/RepeatAudioModal/RepeatAudioModal.tsx
muhajirdev: i see,
we planned to make the delay relative to the duration of previous ayah. So it will be 1x,2x,3x previous ayah.
maybe the suffix should `x prev ayah` ?

what do you think @naveed-ahmad

PR #512, opened 2021-09-30
create badge dls

muhajirdev:


src/components/dls/Badge/Badge.tsx
```diff
@@ -0,0 +1,19 @@
+import classNames from 'classnames';
+
+import styles from './Badge.module.scss';
+
+type BadeProps = {
+  content: React.ReactNode;
+  children: React.ReactNode;
+};
+
+const Badge = ({ content, children }: BadeProps) => {
```
muhajirdev, line 10: Right now the bade is very simple, only support position bottom right. 

Do we have other use case? we can also support other colors and other positions

src/components/AudioPlayer/PlaybackControls.module.scss
```diff
@@ -5,7 +5,7 @@
   justify-content: center;
   align-items: center;
 }
-.container > button + button {
+.container > * + * {
```
muhajirdev, line 8: I know this is not specific. 
But this is very common util and pretty handy.

In fact we can probably turn this into a util that's integrated with our design system, with `--flow-space` css variable. 

References:
- https://every-layout.dev/layouts/stack/
- https://piccalil.li/quick-tip/flow-utility/

src/components/dls/Badge/Badge.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
```
tifa2UP, line 11: Can we add a storybook for this component?

src/components/AudioPlayer/PlaybackControls.module.scss
tifa2UP: I do not actually agree. We can do this to achieve the same result with higher specificity.

```
.container > .stackItem + .stackItem {
	...
}
```

You'll just have to add a className for items you want this rule to apply to.

tifa2UP (approved):

PR #516, opened 2021-09-30
Improve Search page UI

osamasayed:
### Summary
This PR improves the UI of the search page.

### Screenshots

| Before | After |
| ------ | ------ |
|D-B|D-A|
|M-B|M-A|
||Screen Shot 2021-09-30 at 21 32 42|
||Screen Shot 2021-09-30 at 21 32 46|
||Screen Shot 2021-09-30 at 21 32 55|


src/components/Search/Filters/Filter.module.scss
```diff
@@ -0,0 +1,15 @@
+@use "src/styles/breakpoints";
+
+.dropdownLabel {
+  font-weight: var(--font-weight-bold);
+  font-size: var(--font-size-small);
+}
+
+.comboboxItems {
+  div[role="dialog"] {
+    width: 82%;
```
tifa2UP, line 10: not sure if there were hardcoded from before but probably shouldn't be hardcoded.

tifa2UP (approved):

PR #519, opened 2021-10-01
Prevent additional API call when clicking "play" button on TranslationView Cell

muhajirdev:
### Summary

Clicking the play button here, now won't trigger additional API calls to fetch the timestamp data. We will use the verse's timestamp data that comes from verses API

image


src/components/QuranReader/TranslationView/TranslationViewCell.tsx
```diff
@@ -64,7 +64,10 @@ const TranslationViewCell = ({ verse }: TranslationViewCellProps) => {
           </div>
```
tifa2UP, line 64: In addition to this change, can we modify the api call to not get the segment data. We no longer need it.

muhajirdev: Will create another PR for this 👍

tifa2UP (approved):

PR #527, opened 2021-10-03
Add Drawer Component

osamasayed:
### Summary
This PR adds a parent `Drawer` component that will wrap the settings, search and navigations drawers to unify how the UI looks like for all of them and remove duplicated code.


src/components/Navbar/Drawer/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const getActionCreator = (type: DrawerType) => {
+  if (type === DrawerType.Navigation) {
+    return setIsNavigationDrawerOpen.type;
+  }
+  if (type === DrawerType.Settings) {
+    return setIsSettingsDrawerOpen.type;
+  }
+  return setIsSearchDrawerOpen.type;
+};
```
muhajirdev, line 64: quick question 

is it possible to make the Drawer component not aware of the `type`. So it can be more "polymorphism" / decoupled from detail

tifa2UP (approved):

PR #530, opened 2021-10-04
create global text color and background css

muhajirdev:
### Summary
Status: testing

### Screenshots

with global css + local color removed
image

without global css + local color removed
image


src/styles/global.scss
```diff
@@ -0,0 +1,4 @@
+body {
+  background: var(--color-background-default);
+  color: var(--color-text-default);
+}
```
tifa2UP, line 4: Need to do two extra things: 
1. move our global theme settings out of ThemeProvider.module.scss to global.scss and delete that file. 
2. Remove usages of color: var(--color-text-default); and background: var(--color-background-default); since they're now redundant in most components

src/styles/global.scss
```diff
@@ -10,6 +10,12 @@ body {
   ::-moz-selection {
     background: var(--color-highlight);
   }
+
+  overscroll-behavior-y: none;
+  font-family: "ProximaVara", "Helvetica Neue", Helvetica, Arial;
+  line-height: 1;
```
tifa2UP: part of reset

tifa2UP (approved):

PR #531, opened 2021-10-04
fix: audio data bug

muhajirdev:
### Summary

There's a bug when we rename audioFile -> audioData. The backend is still returning filed `audio_files`. This PR fixed the issue


muhajirdev:
merging this, since this is a critical fix

PR #541, opened 2021-10-07
Add command bar

osamasayed:
### Summary
This PR adds a global command bar that is accessible from any page. The command bar can be opened either by clicking `cmd+k` or by clicking on the trigger in the homepage. Currently it can be used to navigate to certain Juz, Page, Hizb, Surah or Verse.

### Screenshots
Screen Shot 2021-10-07 at 14 24 51
Screen Shot 2021-10-07 at 14 25 01
Screen Shot 2021-10-07 at 14 25 14
Screen Shot 2021-10-07 at 14 25 30
Screen Shot 2021-10-07 at 14 25 41

### DIFF
--- a/src/components/CommandBar/CommandBarTrigger/index.tsx
+++ b/src/components/CommandBar/CommandBarTrigger/index.tsx
@@ -1,0 +1,26 @@
+import React, { useCallback } from 'react';
+
+import { useDispatch } from 'react-redux';
+
+import IconSearch from '../../../../public/icons/search.svg';
+
+import styles from './CommandBarTrigger.module.scss';
+
+import KeyboardInput from 'src/components/dls/KeyboardInput';
+import { toggleIsOpen } from 'src/redux/slices/commandBar';
+
+const CommandBarTrigger: React.FC = () => {
+  const dispatch = useDispatch();
+  const onClick = useCallback(() => {
+    dispatch({ type: toggleIsOpen.type });
+  }, [dispatch]);
+  return (
+    <button className={styles.container} type="button" onClick={onClick}>
+      <IconSearch />
+      <span className={styles.text}>Quick access to anything</span>
+      <KeyboardInput meta keyboardKey="K" />
+    </button>
+  );
+};
+
+export default CommandBarTrigger;
--- a/src/components/CommandBar/CommandsList/index.tsx
+++ b/src/components/CommandBar/CommandsList/index.tsx
@@ -7,6 +7,7 @@ import { useRouter } from 'next/router';
 import { useHotkeys } from 'react-hotkeys-hook';
 import { useDispatch } from 'react-redux';
 
+import CloseIcon from '../../../../public/icons/close.svg';
 import NavigateIcon from '../../../../public/icons/east.svg';
 
 import styles from './CommandList.module.scss';
@@ -103,7 +104,7 @@ const CommandsList: React.FC<Props> = ({ navigationItems, isPreInput = false }) 
     );
   }
 
-  const onItemHover = (itemIndex: number) => {
+  const onItemSelected = (itemIndex: number) => {
     setSelectedItemIndex(itemIndex);
   };
 
@@ -124,7 +125,7 @@ const CommandsList: React.FC<Props> = ({ navigationItems, isPreInput = false }) 
                 [styles.selected]: selectedItemIndex === index,
               })}
               onClick={() => navigateToItem(navigationItem)}
-              onMouseOver={() => onItemHover(index)}
+              onMouseOver={() => onItemSelected(index)}
             >
               <div className={styles.commandItemBody}>
                 <span className={styles.commandPrefix}>
@@ -139,7 +140,7 @@ const CommandsList: React.FC<Props> = ({ navigationItems, isPreInput = false }) 
                     size={ButtonSize.Small}
                     onClick={(e) => onRemoveItemClicked(e, navigationItem.key)}
                   >
-                    X
+                    <CloseIcon />
                   </Button>
                 ) : (
                   <KeyboardInput keyboardKey="enter" />
--- a/src/components/CommandBar/CommandsList/index.tsx
+++ b/src/components/CommandBar/CommandsList/index.tsx
@@ -1,0 +1,156 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/mouse-events-have-key-events */
+import React, { MouseEvent, useState, useCallback, RefObject } from 'react';
+
+import classNames from 'classnames';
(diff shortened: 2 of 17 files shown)


src/components/CommandBar/CommandsList/index.tsx
```diff
@@ -0,0 +1,156 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
tifa2UP, line 1: prefer doing this on top of statements, not on top of file. to know exactly what we're disabling

src/components/CommandBar/CommandBarTrigger/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+const CommandBarTrigger: React.FC = () => {
+  const dispatch = useDispatch();
+  const onClick = useCallback(() => {
+    dispatch({ type: toggleIsOpen.type });
+  }, [dispatch]);
+  return (
+    <button className={styles.container} type="button" onClick={onClick}>
+      <IconSearch />
+      <span className={styles.text}>Quick access to anything</span>
+      <KeyboardInput meta keyboardKey="K" />
```
tifa2UP, line 21: discussed this offline, meta should show control on windows

src/components/CommandBar/CommandsList/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const router = useRouter();
+  const dispatch = useDispatch();
+  const onUpKeyClicked = useCallback(
+    (event: KeyboardEvent) => {
+      // prevent the input's cursor to jump to the end of the input value.
+      event.preventDefault();
+      setSelectedItemIndex((prevSelectedItemIndex) => prevSelectedItemIndex - 1);
+      scrollToSelectedItem();
+    },
+    [scrollToSelectedItem],
```
tifa2UP, line 44: want to double check that this on triggered when the command bar is open

src/components/CommandBar/CommandsList/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  if (!numberOfItems) {
+    return (
+      <p className={styles.noResult}>
+        {isPreInput ? 'No recent navigations' : 'No navigation results'}{' '}
+      </p>
+    );
+  }
+
+  const onItemHover = (itemIndex: number) => {
```
tifa2UP, line 106: nit: should we call this highlight/select instead of hover? hover usually indicates mouse hovering

src/components/CommandBar/CommandsList/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+                {navigationItem.name}
+              </div>
+              <div className={styles.keyboardInputContainer}>
+                {isPreInput ? (
+                  <Button
+                    variant={ButtonVariant.Ghost}
+                    size={ButtonSize.Small}
+                    onClick={(e) => onRemoveItemClicked(e, navigationItem.key)}
+                  >
+                    X
```
tifa2UP, line 142: haha, the everlasting hack. Let's use an x icon here. Muhajir recently added a small cross icon from radix ui

src/components/CommandBar/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+const CommandBar: React.FC = () => {
+  const dispatch = useDispatch();
+  const isOpen = useSelector(selectCommandBarIsOpen, shallowEqual);
+  const toggleShowCommandBar = useCallback(() => {
+    dispatch({ type: toggleIsOpen.type });
+  }, [dispatch]);
+  const closeCommandBar = useCallback(() => {
+    dispatch({ type: setIsOpen.type, payload: false });
+  }, [dispatch]);
+  useHotkeys('cmd+k', toggleShowCommandBar, [dispatch]);
```
tifa2UP, line 20: discussed offline

src/components/DataFetcher/index.tsx
```diff
@@ -56,4 +56,4 @@ const DataFetcher: React.FC<Props> = ({ queryKey, render, initialData }: Props):
   return render(data);
 };
 
-export default DataFetcher;
+export default memo(DataFetcher);
```
tifa2UP, line 59: nice

src/components/CommandBar/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  useHotkeys('Escape', closeCommandBar, { enabled: isOpen, enableOnTags: ['INPUT'] }, [dispatch]);
+
+  return (
+    <Modal isOpen={isOpen} onClickOutside={closeCommandBar}>
+      <CommandBarBody />
```
tifa2UP, line 25: can we render this conditionally only when the modal is open (not sure if that's handled as part of the modal code). This way we can make our dom footprint smaller.

src/components/GlobalKeyboardListeners.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import CommandBar from 'src/components/CommandBar';
+
+const GlobalKeyboardListeners: React.FC = () => {
+  return (
+    <>
+      <CommandBar />
```
tifa2UP, line 8: hmmm, a few thoughts here:

1. It seems to me that the commandbar component is not a keyboard listener. A keyboard listener would just dispatch an action to set the commandbar to be open but wouldn't have the actual component code.
2.  Was going to suggest rendering it conditionally but better do it inside of the component body, left the comment above
3. If we apply 1. should CommandBar be under app directly, similar to the feedback widget

src/components/dls/KeyboardInput/KeyboardInput.module.scss
```diff
@@ -12,4 +14,7 @@
   > span + span {
     margin-left: var(--spacing-xsmall);
   }
+  @include breakpoints.smallerThanTablet {
+    display: none;
+  }
 }
```
tifa2UP, line 20: nice

src/components/CommandBar/CommandsList/index.tsx
osamasayed: It still shows the error when using for `next-line` so had to resort to this.

src/components/CommandBar/index.tsx
osamasayed: It's handled as part of the Modal. Anything that is inside the `Modal` is ever mounted in the DOM only when the `Modal` is open.

src/redux/store.ts
```diff
@@ -30,7 +31,7 @@ import theme from './slices/theme';
 
 const persistConfig = {
   key: 'root',
-  version: 7,
+  version: 9,
```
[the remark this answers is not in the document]
tifa2UP: yup

tifa2UP (approved):

PR #548, opened 2021-10-08
audio player: next and prev verse

muhajirdev:
### Summary
TSIA

This feature still has corner cases:
- when we set repeat each verse to 2 -> when we go to prev verse -> it will not repeat the verse 2 times, instead, it will only repeat 1 time.

- when user set repeatFrom and repeatTo verse 1:3 - 1:3. And their repeatEachVerse 1 and repeatRange 3. When user click "next ayah" should it repeat the verse or just stop the verse?

Will do a follow up PR to fix this issue

### DIFF
--- a/src/components/AudioPlayer/SeekButton.tsx
+++ b/src/components/AudioPlayer/SeekButton.tsx
@@ -12,6 +12,7 @@ import { getChapterAudioData } from 'src/api';
 import Button, { ButtonShape, ButtonVariant } from 'src/components/dls/Button/Button';
 import {
   finishRepeatEachVerseProgress,
+  resetRepeatEachVerseProgress,
   selectAudioData,
   selectIsInRepeatMode,
   selectReciter,
@@ -55,7 +56,8 @@ const SeekButton = ({ type, isLoading }: SeekButtonProps) => {
     if (isInRepeatMode) {
       // when in repeatMode, finish the repeat progress for current ayah
       // otherwise the AudioRepeatManager will replay the current Ayah when we set the new timestamp
-      dispatch(finishRepeatEachVerseProgress());
+      if (type === SeekButtonType.NextAyah) dispatch(finishRepeatEachVerseProgress());
+      else dispatch(resetRepeatEachVerseProgress());
     }
     const newVerse = type === SeekButtonType.PrevAyah ? highlightedVerse - 1 : highlightedVerse + 1;
     const verseKey = makeVerseKey(highlightedChapter, newVerse);
--- a/src/redux/slices/AudioPlayer/state.ts
+++ b/src/redux/slices/AudioPlayer/state.ts
@@ -236,6 +236,13 @@ export const audioPlayerStateSlice = createSlice({
         repeatEachVerse: state.repeatSettings.repeatEachVerse,
       },
     }),
+    resetRepeatEachVerseProgress: (state) => ({
+      ...state,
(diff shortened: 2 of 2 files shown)


src/components/AudioPlayer/SeekButton.tsx
```diff
@@ -1,30 +1,76 @@
...
+  NextAyah = 'nextAyah',
+  PrevAyah = 'prevAyah',
 }
 
 type SeekButtonProps = {
   type: SeekButtonType;
   isLoading: boolean;
 };
 const SeekButton = ({ type, isLoading }: SeekButtonProps) => {
+  const { highlightedChapter, highlightedVerse } = useSelector(selectHighlightedLocation);
```
tifa2UP, line 30: We shouldn't use highlighted here, we should use the more generic hook. No? I think that we called it activeVerse

src/components/AudioPlayer/SeekButton.tsx
```diff
@@ -1,30 +1,76 @@
...
+    reciter.id && audioData?.chapterId
+      ? makeChapterAudioDataUrl(reciter.id, audioData?.chapterId, true)
+      : null, // only fetch when reciterId and chapterId is available
+    () => getChapterAudioData(reciter.id, audioData?.chapterId, true),
+  );
+
+  const verseTimingData = chapterAudioData?.verseTimings || [];
+
+  const onSeek = () => {
+    const verse = type === SeekButtonType.PrevAyah ? highlightedVerse - 1 : highlightedVerse + 1;
```
tifa2UP, line 48: verse -> newVerse?

src/components/AudioPlayer/SeekButton.tsx
```diff
@@ -1,30 +1,76 @@
...
+
+  // disable the button if loading, or chapterAudioData not available, or highlighted verse not available
+  // or currently playing first verse or last verse
+  const isDisabled =
+    isLoading ||
+    !highlightedChapter ||
+    !highlightedVerse ||
+    !chapterAudioData ||
+    (type === SeekButtonType.PrevAyah && highlightedVerse <= 1) ||
+    (type === SeekButtonType.NextAyah && highlightedVerse >= chapterData?.versesCount);
```
tifa2UP, line 63: nice

src/components/AudioPlayer/SeekButton.tsx
```diff
@@ -1,30 +1,76 @@
...
+      tooltip={type === SeekButtonType.PrevAyah ? 'Previous Ayah' : 'Next Ayah'}
       variant={ButtonVariant.Ghost}
       shape={ButtonShape.Circle}
-      disabled={isLoading}
-      onClick={withStopPropagation(() => triggerSeek(type === SeekButtonType.Rewind ? -10 : 10))}
+      disabled={isDisabled}
+      onClick={onSeek}
     >
-      {type === SeekButtonType.Rewind ? <ReplayIcon /> : <ForwardIcon />}
+      {type === SeekButtonType.PrevAyah ? <PrevIcon /> : <NextIcon />}
```
tifa2UP, line 73: 2 comments on this: 

1. Can we use this icons instead? I believe the ones you used indicate playing the previous/next track
image

2. Can we use icons that use the same size as the other audio player icons? the ones on the deployment appear a bit bigger

tifa2UP (approved):

PR #551, opened 2021-10-09
Create Switch DLS

muhajirdev:
### Summary
TSIA

image

It doesn't really look like the reference (vercel), mainly because of the sizing. (e.g we don't have 14px font size token)

API and design is derived from https://vercel.com/design/switch

### DIFF
--- a/src/components/dls/Switch/Switch.module.scss
+++ b/src/components/dls/Switch/Switch.module.scss
@@ -30,3 +30,13 @@
   cursor: not-allowed;
   opacity: var(--opacity-50);
 }
+
+// sizing
+.itemLarge {
+  padding: var(--spacing-small) var(--spacing-small);
+}
+
+.itemSmall {
+  font-size: var(--font-size-small);
+  padding: var(--spacing-xxsmall) var(--spacing-small);
+}
--- a/src/components/dls/Switch/Switch.module.scss
+++ b/src/components/dls/Switch/Switch.module.scss
@@ -1,10 +1,15 @@
 .container {
-  display: flex;
+  display: grid;
+  grid-template-columns: repeat(
+    auto-fit,
+    minmax(0, 1fr)
+  ); // use grid to prevent layout shift when `font-family: bold` on item selected https://css-tricks.com/bold-on-hover-without-the-layout-shift/
 }
 
 .item {
   cursor: pointer;
   background-color: var(--color-background-default);
+  color: var(--color-text-default);
   border: 1px solid var(--color-borders-hairline);
 }
 .item:first-child {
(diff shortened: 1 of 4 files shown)


src/components/dls/Switch/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type Item = {
+  name: string;
+  value: string;
+  width: number;
+  disabled?: boolean;
+};
+export enum SwitchSize {
+  Small = 'small',
+  Normal = 'normal',
```
osamasayed, line 15: I think "Normal" should be renamed to "Medium" and also inside the css class names for the sake of consistency with other components when it comes to sizing

src/components/dls/Switch/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+          disabled={item.disabled}
+          type="button"
+          className={classNames(styles.item, selected === item.value && styles.itemSelected, {
+            [styles.itemLarge]: size === SwitchSize.Large,
+            [styles.itemNormal]: size === SwitchSize.Normal,
+            [styles.itemSmall]: size === SwitchSize.Small,
+          })}
+          key={item.value}
+          onClick={() => onSelect(item.value)}
+          style={{ width: item.width }}
```
osamasayed, line 39: I don't think we should have a flexible width value here. The size is what should control the width

src/components/dls/Switch/Switch.tsx
```diff
@@ -0,0 +1,48 @@
+import classNames from 'classnames';
+
+import styles from './Switch.module.scss';
+
+// reference: https://vercel.com/design/switch
+
+type Item = {
+  name: string;
+  value: string;
+  width: number;
```
osamasayed, line 10: we shouldn't use `width`

src/components/dls/Switch/Switch.tsx
muhajirdev: agree 👍

tifa2UP: ditto

src/components/dls/Switch/Switch.tsx
tifa2UP: yeah, let's not make width a prop. We can have a `fullWidth` option that takes 100% of the container width.

tifa2UP (approved):

tifa2UP (approved):

src/components/dls/Switch/Switch.module.scss
```diff
@@ -0,0 +1,45 @@
+.container {
```
tifa2UP, line 1: switch doesn't appear to be working on dark mode

PR #558, opened 2021-10-09
Fix API root path

osamasayed:
### Summary
This PR removes an extra `/` added at the end of the root path which resulted in the generated API path being in the following format:

`https://staging.quran.com/api/qdc//resources/translations?language=en`

instead of

`https://staging.quran.com/api/qdc/resources/translations?language=en`

Screen Shot 2021-10-10 at 06 28 23


muhajirdev:
Looks goood 👍

PR #559, opened 2021-10-10
Remove 171 as default Tafsir

osamasayed:
### Summary
This PR removes tafsir 171 as a default tafsir since it's no longer a tafsir but a translation.


muhajirdev:
Looks good to me

tifa2UP (approved):

PR #561, opened 2021-10-10
add audioplayer control speed

muhajirdev:
image

Currently the speed is not persisted across session, will persist it in the next PR


src/components/Navbar/SettingsDrawer/AudioSection.tsx
```diff
@@ -78,11 +86,24 @@ const AudioSection = () => {
...
+            </Section.Row>
           </Section>
         )}
       />
     </div>
   );
 };
 
+const DEFAULT_PLAYBACK_RATE = '1';
+const playbackRates = generateSelectOptions(['0.5', '1', '1.5', '2']);
```
tifa2UP, line 107: Nice! Can we use these values? Taken from youtube.

Can we also display "Normal" when the value is 1? I think that it'll be clearer for less technical users

image

src/components/Navbar/SettingsDrawer/AudioSection.tsx
```diff
@@ -37,6 +39,7 @@ const AudioSection = () => {
   const dispatch = useDispatch();
   const selectedReciter = useSelector(selectReciter, shallowEqual);
   const enableAutoScrolling = useSelector(selectEnableAutoScrolling);
+  const [playbackRate, setPlayBackRateState] = useState(DEFAULT_PLAYBACK_RATE);
 
   // given the reciterId, get the full reciter object.
```
tifa2UP, line 44: Nice! this all looks good. The only change is that we want to save and persist the playback speed in redux. 

Users who customize the playback speed would usually like to maintain that speed cross session.

PR #562, opened 2021-10-10
implement reading preference switcher

muhajirdev:
Status: still doing testing

image
image
image
image


src/components/QuranReader/ReadingPreferenceSwitcher.module.scss
```diff
@@ -0,0 +1,4 @@
+.container {
+  max-width: calc(20 * var(--spacing-large));
```
tifa2UP, line 2: 1. this should be a const
2. we should do a css min between this value and 90%. because we don't want it to occupy the full width on small devices.

image

tifa2UP (approved):

PR #564, opened 2021-10-11
create helper tooltip component

muhajirdev:
### Summary
Helper tooltip + 1 implementation example


src/components/dls/HelperTooltip/HelperTooltip.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import styles from './HelperTooltip.module.scss';
+
+import Tooltip from 'src/components/dls/Tooltip';
+
+interface HelperTooltipProps {
+  children: React.ReactNode;
+}
+
+const HelperTooltip = ({ children }: HelperTooltipProps) => {
```
tifa2UP, line 11: Looks good overall mA, two things here: 

- can we use an icon with a grey fill similar to canny.io/pricing
- it looks like the icon is not vertically centered (looking at the continue reading example). Can we add the property to the style of the component?

PR #569, opened 2021-10-14
Update QuranReader intersectionObserver

osamasayed:
### Summary
This PR updates the logic of using `IntersectionObserver` to detect which Ayah is being read. Before, We were creating a new `IntersectionObserver` instance every time `VerseText`/`ChapterHeader` component mounts. This seems to affect performance. An alternative way this PR introduces is that we attach a global observer inside `window` object and each `VerseText`/`ChapterHeader` would ask that observer to observe when it lies within the observer's window. This seems to have positive gains in terms of performance as shown below. The below performance snapshots were taking when scrolling from the beginning of Surah Yaseen to Ayah 27. 

| Before | After |
| ------ | ------ |
|Screen Shot 2021-10-14 at 11 35 48|Screen Shot 2021-10-14 at 11 35 54|

Also having a single `IntersectionObserver` seems to reduce total blocking time (still needs to be improved but not related to Intersection observers).

| Before | After |
| ------ | ------ |
|Screen Shot 2021-10-14 at 12 22 59|Screen Shot 2021-10-14 at 12 23 07|

This PR also adjusts the Observer window to handle all screens by making the top boundary of the Observer's window an absolute value instead of it being relative to the viewport because it results in different behavior on different screens.

| Before | After |
| ------ | ------ |
|Screen Shot 2021-10-14 at 12 30 27|Screen Shot 2021-10-14 at 12 29 32|
|Screen Shot 2021-10-14 at 12 30 17|Screen Shot 2021-10-14 at 12 29 49|


src/components/QuranReader/Loading.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <>
+      <div
+        className={classNames(containerClassName, {
+          [visibleSideBarClassName]: isSideBarVisible,
+        })}
+      >
+        <Spinner size={SpinnerSize.Large} isCentered />
+      </div>
+      <Notes />
```
tifa2UP, line 29: Is notes meant to be under loading?

src/components/QuranReader/observer.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Get the observer options based on the reading preference.
+ *
+ * @param {boolean} isReadingPreference
+ * @returns {{rootMargin: string, threshold: number | number[]}}
+ */
+export const getOptions = (
+  isReadingPreference: boolean,
+): { rootMargin: string; threshold: number | number[] } => ({
+  rootMargin: isReadingPreference ? READING_MODE_ROOT_MARGIN : DEFAULT_ROOT_MARGIN,
+  threshold: OBSERVER_THRESHOLD,
```
tifa2UP, line 34: Are we using the 50px threshold with this implementation?

src/components/Verse/VerseText.tsx
```diff
@@ -21,31 +20,10 @@ type VerseTextProps = {
   isHighlighted?: boolean;
 };
 
-const READING_MODE_ROOT_MARGIN = '-10% 0px -85% 0px';
-const DEFAULT_ROOT_MARGIN = '-23% 0px -72% 0px';
-const OBSERVER_THRESHOLD = 0.01;
-
+const OBSERVER_NAME = 'quranReaderObserver';
```
tifa2UP, line 23: should this be imported instead of redefined? and maybe it should be ID instead of name

`QURAN_READER_OBSERVER_ID`

src/components/QuranReader/Loading.tsx
osamasayed: It was there from before. The idea is that if the Notes are open and the QuranReader is loading because let's say the user has just changed his Translations settings, the Notes should still be visible.

src/components/QuranReader/observer.ts
osamasayed: The 50px threshold is used inside `GlobalScrollListener` and both will not be affected by each other.

tifa2UP (approved):

PR #576, opened 2021-10-15
new design: navbar

muhajirdev:
TSIA

image
image


src/components/Navbar/NavbarBody/index.tsx
```diff
@@ -49,13 +48,16 @@ const NavbarBody: React.FC = () => {
...
-                <IconQ />
-              </Button>
-            </a>
-          </Link>
+          <Button
+            href="/"
+            shape={ButtonShape.Circle}
+            variant={ButtonVariant.Ghost}
+            className={styles.logoWrapper}
+            size={ButtonSize.Large}
```
muhajirdev, line 56: - using button size large, following the design (Q icon is bigger than the menu icon). 
- But there is a side effect, try hovering on menu icon and Q icon. 
  - The sizing of the  background color will bit slightly different

PR #589, opened 2021-10-16
[WIP] Improve web vitals metrics

osamasayed:


src/pages/_app.tsx
```diff
@@ -20,12 +20,36 @@ import 'src/styles/fonts.scss';
...
+        <link
+          rel="preload"
+          as="font"
+          type="font/woff2"
+          href="/fonts/ProximaVara/proxima_vara_bold.woff2"
+          crossOrigin="anonymous"
+        />
         <link rel="manifest" href="/manifest.json" />
         <link rel="preconnect" href={API_HOST} />
+        <link rel="preconnect" href={VERCEL_INSIGHTS_URL} />
```
tifa2UP, line 52: dt that we need to connect to it

src/styles/fonts.scss
```diff
@@ -56,7 +56,7 @@
   src: url("/fonts/ProximaVara/proxima_vara_regular.woff2"),
     url("/fonts/ProximaVara/proxima_vara_regular.woff");
   font-weight: normal;
-  font-display: swap;
+  font-display: optional;
```
tifa2UP, line 59: confirm that optional is optimal for us

osamasayed: `optional` is an optimization that prevents Layout shift. So basically there would be a window (100ms) for Proxima font to load. During this window the text doesn't show and if the font finishes loading within this window, the text shows in Proxima font, if not, it will show the fallback font https://web.dev/preload-optional-fonts/

PR #595, opened 2021-10-17
add padding to drawer

muhajirdev:
### Summary
TSIA

### DIFF
--- a/src/components/Navbar/Drawer/Drawer.module.scss
+++ b/src/components/Navbar/Drawer/Drawer.module.scss
@@ -44,7 +44,10 @@
 
 .bodyContainer {
   margin-top: constants.$navbar-height;
-  padding: var(--spacing-small);
+  padding-top: var(--spacing-small);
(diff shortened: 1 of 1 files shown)


src/components/Navbar/Drawer/Drawer.module.scss
```diff
@@ -44,7 +44,10 @@
 
 .bodyContainer {
   margin-top: constants.$navbar-height;
-  padding: var(--spacing-small);
+  padding-top: var(--spacing-small);
+  padding-bottom: calc(2 * var(--spacing-large));
+  padding-left: var(--spacing-small);
+  padding-right: var(--spacing-small);
```
muhajirdev, line 50: using -top, -bottom, -left, and -right is more verbose. I find it be more clear instead using one liner `paddding: var(--spacing-small) var(--spacing-small) calc(2 * var(--spacing-large)) var(--spacing-small)`;

PR #717, opened 2021-11-08
Fix font of footnote text and translation name

naveed-ahmad:
### Summary
TSIA

### Screenshots

| Before | After |
| ------ | ------ |
| b|after|


src/components/QuranReader/TranslationView/TranslationText/FootnoteText.tsx
```diff
@@ -4,34 +4,43 @@ import React, { MouseEvent } from 'react';
...
 import classNames from 'classnames';
 import useTranslation from 'next-translate/useTranslation';
+import { useRouter } from 'next/router';
 
 import CloseIcon from '../../../../../public/icons/close.svg';
+import {
+  getLanguageDirectionById,
+  getLanguageFontById,
+  findLanguageIdByLocale,
+} from '../../../../utils/locale';
```
osamasayed, line 14: the path should be `src/utils/locale` instead

src/components/QuranReader/TranslationView/TranslationText/FootnoteText.tsx
```diff
@@ -4,34 +4,43 @@ import React, { MouseEvent } from 'react';
...
 const FootnoteText: React.FC<FootnoteTextProps> = ({
-  text,
+  footnote,
   onCloseClicked,
   onTextClicked,
   isLoading,
-  isStaticContent = false,
 }) => {
-  const direction = useDirection();
+  const { locale } = useRouter();
```
osamasayed, line 36: you can access `lang` inside `useTranslation` instead of using `useRouter` in addition to `useTranslation`

src/components/QuranReader/TranslationView/TranslationText/index.tsx
```diff
@@ -8,6 +8,8 @@ import React, { MouseEvent, useState } from 'react';
 import classNames from 'classnames';
 import useTranslation from 'next-translate/useTranslation';
 
+import { getLanguageDirectionById, getLanguageFontById } from '../../../../utils/locale';
```
osamasayed, line 11: the path should be `src/utils/locale` instead

src/utils/locale.ts
```diff
@@ -62,6 +147,39 @@ export const isRTLLocale = (locale: string): boolean => RTL_LOCALES.includes(loc
 export const getDir = (locale: string): Direction =>
   isRTLLocale(locale) ? Direction.RTL : Direction.LTR;
 
+/**
+ * Get direction of language by language id
+ *
+ * @param {number} languageId
+ * @returns {string}
+ */
+export const getLanguageDirectionById = (languageId: number): string => {
```
osamasayed, line 156: instead of having `getLanguageFontById` and `getLanguageFontById` together and calling them twice to just access two different keys why don't we combine them in 1 method e.g. `getLanguageDataById` or something similar

PR #750, opened 2021-11-13
Added Skeleton components for sidemenus and homepage

mahmoudfelfel:
### Summary
Replaced `Spinner` component with Skeleton for homepage's side menus and Chapters list.

### Test Plan
The skeletons disappear once their component loads, so to test it you mostly have to throttle the network to have enough time to see the skeleton loader.
I tested it on different screen sizes and RTL.

**Chapters list skeleton:**
https://user-images.githubusercontent.com/3989750/141656066-44e282a5-bc07-4138-8194-62de8aaefe3b.mp4

**Settings Menu Skeleton**
https://user-images.githubusercontent.com/3989750/141656281-72d36b72-6a12-47d8-9c11-8d88330191ca.mp4

**Nav Menu Skeleton:**
https://user-images.githubusercontent.com/3989750/141656168-e52c09ea-2c76-492c-b340-b298780252b6.mp4

### DIFF
--- a/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
+++ b/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
@@ -1,0 +1,16 @@
+.skeletonContainer {
+  display: flex;
+  flex-direction: column;
+}
+
+.navRow {
+  width: auto;
+  height: 40px;
+  margin: 8px 10px 4px;
+}
+
+.blockRow {
+    width: auto;
+    height: 200px;
+    margin: 8px 10px 4px;
+}
--- a/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
+++ b/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
@@ -5,12 +5,15 @@
 
 .navRow {
   width: auto;
-  height: 40px;
-  margin: 8px 10px 4px;
+  height: var(--spacing-mega);
+  padding: var(--spacing-xxsmall) 0;
+  margin: var(--spacing-xxsmall) var(--spacing-xsmall);
+  margin-inline-start: var(--spacing-small);
 }
 
 .blockRow {
-    width: auto;
-    height: 200px;
-    margin: 8px 10px 4px;
+  width: auto;
+  height: calc(var(--spacing-mega) * 6);
+  margin: var(--spacing-xxsmall) var(--spacing-xsmall);
+  margin-inline-start: var(--spacing-small);
 }
--- a/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
+++ b/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
@@ -1,6 +1,6 @@
-import Skeleton from "../../dls/Skeleton/Skeleton";
+import Skeleton from '../../dls/Skeleton/Skeleton';
 
-import styles from "./NavigationDrawerBodySkeleton.module.scss";
+import styles from './NavigationDrawerBodySkeleton.module.scss';
 
 const NAV_ROW_COUNT = 6;
 const rowsArr = Array(NAV_ROW_COUNT).fill(null);
@@ -8,12 +8,7 @@ const rowsArr = Array(NAV_ROW_COUNT).fill(null);
 const renderLinesSkeleton = (index) => {
   return rowsArr.map((k, i) => (
     // eslint-disable-next-line react/no-array-index-key
-    <Skeleton
-      key={`skeleton_${index}_${i}`}
-      isActive
-      isSquared
-      className={styles.navRow}
-    />
+    <Skeleton key={`skeleton_${index}_${i}`} isActive isSquared className={styles.navRow} />
   ));
 };
 
--- a/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
+++ b/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
@@ -1,0 +1,24 @@
+import Skeleton from '../../dls/Skeleton/Skeleton';
+
+import styles from './NavigationDrawerBodySkeleton.module.scss';
+
+const NAV_ROW_COUNT = 6;
+const rowsArr = Array(NAV_ROW_COUNT).fill(null);
+
+const NavigationDrawerBodySkeleton = () => {
+  return (
+    <span className={styles.skeletonContainer}>
+      {rowsArr.map((k, i) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Skeleton key={`skeleton_${i}`} isActive isSquared className={styles.navRow} />
+      ))}
+      <Skeleton isActive isSquared className={styles.blockRow} />
+      {rowsArr.map((k, i) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Skeleton key={`skeleton_${i}`} isActive isSquared className={styles.navRow} />
+      ))}
+    </span>
+  );
+};
+
+export default NavigationDrawerBodySkeleton;
--- a/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
+++ b/src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
@@ -1,22 +1,28 @@
-import Skeleton from '../../dls/Skeleton/Skeleton';
+import Skeleton from "../../dls/Skeleton/Skeleton";
 
-import styles from './NavigationDrawerBodySkeleton.module.scss';
+import styles from "./NavigationDrawerBodySkeleton.module.scss";
 
 const NAV_ROW_COUNT = 6;
 const rowsArr = Array(NAV_ROW_COUNT).fill(null);
 
+const renderLinesSkeleton = (index) => {
+  return rowsArr.map((k, i) => (
+    // eslint-disable-next-line react/no-array-index-key
+    <Skeleton
+      key={`skeleton_${index}_${i}`}
+      isActive
+      isSquared
+      className={styles.navRow}
+    />
+  ));
+};
+
 const NavigationDrawerBodySkeleton = () => {
   return (
     <span className={styles.skeletonContainer}>
-      {rowsArr.map((k, i) => (
-        // eslint-disable-next-line react/no-array-index-key
-        <Skeleton key={`skeleton_${i}`} isActive isSquared className={styles.navRow} />
-      ))}
+      {renderLinesSkeleton(1)}
       <Skeleton isActive isSquared className={styles.blockRow} />
-      {rowsArr.map((k, i) => (
-        // eslint-disable-next-line react/no-array-index-key
-        <Skeleton key={`skeleton_${i}`} isActive isSquared className={styles.navRow} />
-      ))}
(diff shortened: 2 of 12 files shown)


src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
```diff
@@ -0,0 +1,16 @@
+.skeletonContainer {
+  display: flex;
+  flex-direction: column;
+}
+
+.navRow {
+  width: auto;
+  height: 40px;
+  margin: 8px 10px 4px;
```
tifa2UP, line 9: A couple of things: 
1. We don't use hardcoded values for the height, we use the spacing tokens from the theme file. 
2. The margin should use logical css instead to properly support rtl layouts.

src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.module.scss
```diff
@@ -0,0 +1,16 @@
...
+}
+
+.blockRow {
+    width: auto;
+    height: 200px;
+    margin: 8px 10px 4px;
```
tifa2UP, line 15: ditto

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  flex-direction: row;
+  align-items: center;
+}
+
+.inputRow {
+  margin-bottom: 18px;
+  padding-bottom: 18px;
+  border-bottom: 1px solid;
+  [data-theme="dark"] & {
+    border-color: rgba(255, 255, 255, 0.13);
```
tifa2UP, line 29: ditto, should use a theme token

src/components/dls/Skeleton/Skeleton.module.scss
```diff
@@ -23,18 +14,16 @@ $animationDuration: 8s;
...
-  content: "";
+
+  [data-theme="dark"] & {
+    background-color: rgba(255, 255, 255, 0.13);
+  }
+  [data-theme="light"] & {
+    background-color: rgba(0, 0, 0, 0.11);
+  }
+  animation: 1.5s ease-in-out 0.5s infinite normal none running loading;
+
```
tifa2UP, line 25: Can you post a before/after for this change? Since it touches our dls component.

src/components/chapters/ChapterAndJuzListSkeleton.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const CHAPTERS_COUNT = 114;
+const chaptersArr = Array(CHAPTERS_COUNT).fill(null);
+
+const ChapterAndJuzListSkeleton = () => {
+  return (
+    <span className={styles.skeletonContainer}>
+      <div className={styles.tabSkeleton}>
+        <Skeleton isActive isSquared className={styles.tabSkeletonItem} />
+        <Skeleton isActive isSquared style={{ width: '100px' }} />
```
tifa2UP, line 13: no inline styles

src/components/dls/Skeleton/Skeleton.module.scss
mahmoudfelfel: **before:**

https://user-images.githubusercontent.com/3989750/141662811-b4290f5f-b3f9-41b2-bf3a-b71e63916321.mov


https://user-images.githubusercontent.com/3989750/141662812-56a57b84-d9f4-4f70-abe4-6dc22cbcd9a3.mov

**after:**

https://user-images.githubusercontent.com/3989750/141662825-32ccdab6-a7be-4397-8d0b-104b2ccbd03f.mov


https://user-images.githubusercontent.com/3989750/141662826-227c174b-131c-4a29-93a7-84f8bcba63ff.mov

src/components/Navbar/NavigationDrawer/NavigationDrawerBodySkeleton.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const NavigationDrawerBodySkeleton = () => {
+  return (
+    <span className={styles.skeletonContainer}>
+      {rowsArr.map((k, i) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Skeleton key={`skeleton_${i}`} isActive isSquared className={styles.navRow} />
+      ))}
+      <Skeleton isActive isSquared className={styles.blockRow} />
+      {rowsArr.map((k, i) => (
```
osamasayed, line 16: 2 things:
1. the `.map` code bit is duplicated so it makes sense to define it once and re-use it.
2. Because the code is running twice and the index is being used so for each element in the first loop there will be a corresponding element in the 2nd loop with the same key e.g. `skeleton_0` will exist twice, `skeleton_1` will exists twice and so on.

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.module.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.rowSkeleton {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+}
+
+.inputRow {
+  margin-bottom: var(--spacing-large);
```
osamasayed, line 25: Do you have `style-lint` installed? because `margin-bottom/up` is a physical value and we are meant to use [logical values](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties). `yarn lint:scss` should have run on committing which would have auto-fixed this but for some reason it seems it was not run so you might want to manually run `yarn lint:scss` and push again.

src/components/chapters/ChapterAndJuzList.tsx
```diff
@@ -3,27 +3,18 @@ import React, { useState, useMemo } from 'react';
...
-import dynamic from 'next/dynamic';
 
 import CaretDownIcon from '../../../public/icons/caret-down.svg';
 import Link from '../dls/Link/Link';
-import Spinner from '../dls/Spinner/Spinner';
 import SurahPreviewRow from '../dls/SurahPreview/SurahPreviewRow';
 import Tabs from '../dls/Tabs/Tabs';
 
 import styles from './ChapterAndJuzList.module.scss';
+import JuzView from './JuzView';
```
osamasayed, line 13: Is there a reason why `dynamic` was removed? we have added it to lazy load this component instead of it being part of the initial payload which makes it larger and slower to parse

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.tsx
```diff
@@ -0,0 +1,27 @@
+import Skeleton from '../../dls/Skeleton/Skeleton';
+
+import styles from './SettingsBodySkeleton.module.scss';
+
+const SETTINGS_ROW_COUNT = 4;
+const rowsArr = Array(SETTINGS_ROW_COUNT).fill(null);
```
osamasayed, line 6: I feel this code is being duplicated in multiple places and It would be better to move it to a reusable Skeletons generator

src/components/dls/Skeleton/Skeleton.module.scss
```diff
@@ -23,18 +14,16 @@ $animationDuration: 8s;
...
-    var(--shade-1),
-    var(--shade-2),
-    var(--shade-1),
-    var(--shade-0)
-  );
-  background-size: 400% 100%; // increase background size's width to 4x to make background animate-able
-  animation: loading $animationDuration ease-in-out infinite;
-
-  content: "";
+
```
osamasayed, line 17: There is a missing case that the user is choosing `System` theme and his system's theme is dark. in this case the skeletons do not show

Screen Shot 2021-11-19 at 19 17 41

mahmoudfelfel: PR https://github.com/quran/quran.com-frontend-next/pull/781 has a fix for this, we can merge it first then I will rebase this PR.

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.module.scss
mahmoudfelfel: dono, but will run it manually and push.
Just curious, I understand the value of logical margins for right and left, but what difference it will make for up/bottom?

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.tsx
mahmoudfelfel: hmm, it is already two lines of code and only required in some of the skeletons based on the component, are you sure abstracting it is needed?

src/components/dls/Skeleton/Skeleton.module.scss
```diff
@@ -23,18 +14,16 @@ $animationDuration: 8s;
...
+  [data-theme="dark"] & {
```
osamasayed, line 18: same as the other PR, we should use `styles/_theme.scss`

src/components/Navbar/SettingsDrawer/SettingsBodySkeleton.tsx
osamasayed: So what I mean is not just this line but the the body of the `SettingsBodySkeleton` and other Skeletons too because I feel all of them share most of the code and what changes is just the className + the number of skeletons generated. But not very strong opinion

src/components/dls/Skeleton/Skeleton.module.scss
```diff
@@ -1,15 +1,7 @@
+@use "src/styles/theme";
 $baseSize: var(--spacing-mega);
 $animationDuration: 8s;
 
-@keyframes loading {
```
muhajirdev, line 4: sorry, I am I didn't follow this PR since the beginning.

but is there a reason we're removing this loading `background-position` animation ?

src/components/dls/Skeleton/Skeleton.module.scss
```diff
@@ -47,6 +38,25 @@ $animationDuration: 8s;
...
+  0% {
+    opacity: 1;
+  }
+  50% {
+    opacity: 0.4;
+  }
+  100% {
+    opacity: 1;
+  }
+}
```
muhajirdev, line 51: I think the postcss in nextjs will handle this for us? or

src/components/Navbar/Drawer/Drawer.module.scss
```diff
@@ -2,7 +2,7 @@
 @use "src/styles/constants";
 
 .container {
-  background: var(--color-background-elevated);
```
muhajirdev, line 5: We'll probably lose some details if we use `faded`

with `faded`
image


with `elevated`

image

src/components/dls/Skeleton/Skeleton.module.scss
mahmoudfelfel: The animation wasn't showing up in dark mode, I could've fixed that but I found Material UI use this approach and it looked better, I commented above with videos to show the difference. wdyt?

src/components/dls/Skeleton/Skeleton.module.scss
mahmoudfelfel: yes, I think it should. will remove it.

src/components/Navbar/Drawer/Drawer.module.scss
mahmoudfelfel: the current loading animation colors didn't look good with the new bg:
image

hmm, will try to change the skeleton color to look good while keeping the same bg.

osamasayed (approved):

PR #784, opened 2021-11-17
Use language specific font for wbw translation

naveed-ahmad:
### Summary
TSIA

### Screenshots

| Before | After |
| ------ | ------ |

### DIFF
--- a/src/components/dls/QuranWord/QuranWord.tsx
+++ b/src/components/dls/QuranWord/QuranWord.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 import React, { ReactNode, useState, useMemo, useCallback } from 'react';
 
 import classNames from 'classnames';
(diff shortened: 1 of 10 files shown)


src/components/dls/QuranWord/QuranWord.tsx
```diff
@@ -64,18 +67,14 @@ const QuranWord = ({
...
-    selectWordByWordByWordPreferences,
+    selectWordByWordPreferences,
     shallowEqual,
   );
   const readingPreference = useSelector(selectReadingPreference);
   const showTooltipFor = useSelector(selectShowTooltipFor, areArraysEqual);
 
-  // creating wordLocation instead of using `word.location` because
-  // the value of `word.location` is `1:3:5-7`, but we want `1:3:5`
-  const wordLocation = makeWordLocation(word.verseKey, word.position);
```
osamasayed, line 75: @muhajirdev is removing this safe?

muhajirdev: I think @naveed-ahmad updated the API. so `word.location` is already using the expected format now, so we can remove `wordLoation`

naveed-ahmad:
PR https://github.com/quran/quran.com-frontend-next/pull/832 will solve some of these issues. And now implementation would be much simpler. Closing it

PR #790, opened 2021-11-19
Change word translation language query

naveed-ahmad:
### Summary
TSIA

### Screenshots

| Before | After |
| ------ | ------ |


src/components/QuranReader/api.ts
```diff
@@ -72,7 +72,7 @@ export const getRequestKey = ({
   }
```
osamasayed, line 72: We should change for `makeJuzVersesUrl` and `makePageVersesUrl` also

PR #791, opened 2021-11-19
Fixes #782 | translation font size change not reflecting

### LINKED ISSUE #782: Translation font-size change needs  page refresh
I can resize arabic text, but translation font-size needs to page refresh. Does not work like arabixc font-size.

image

AadilMehrajBhat:
### Summary
The translations font size change isn't reflecting while changing the font size from the settings drawer

### Screenshots

| Before | After |
| ------ | ------ |

### DIFF
--- a/src/components/QuranReader/utils/memoization.ts
+++ b/src/components/QuranReader/utils/memoization.ts
@@ -1,5 +1,3 @@
-import isEqual from 'lodash/isEqual';
-
 import QuranReaderStyles from 'src/redux/types/QuranReaderStyles';
 import Verse from 'types/Verse';
 import Word from 'types/Word';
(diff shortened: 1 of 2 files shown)


muhajirdev:
awesome, this seemss to fix the issue.

src/components/QuranReader/utils/memoization.ts
```diff
@@ -30,7 +32,7 @@ export const verseFontChanged = (
   prevWords: Word[],
   nextWords: Word[],
 ): boolean =>
-  prevQuranStyles.mushafLines !== nextQuranStyles.mushafLines ||
+  !isEqual(prevQuranStyles, nextQuranStyles) ||
```
osamasayed, line 35: using `isEqual` is a great solution but I have 2 comments:

1. We try to stay away from using packages as much as possible if the same can be easily achieved manually.
2. Comparing the entire `quranStyles` object would result in un-necessary re-renders even if the changes are not related to the translation view e.g. changing the Tafsir font while the user is viewing translation will cause every word in the translation view to re-render. You can check the profiler snapshot below for the case of changing the Tafsir font, while viewing the translations, all verses re-rendered:

Screen Shot 2021-11-21 at 09 00 07

so I suggest targeting only the translation font `prev` and `next` and doing the comparison inside `TranslationViewCell.areVersesEqual` because the translation font change is only related to the translation view and not the reading view. JazakAllah Khairan

osamasayed (changes requested):

src/components/QuranReader/utils/memoization.ts
AadilMehrajBhat: @osamasayed Jazakallahu khair for review the PR, I have update the PR as suggested. I have added a separate helper `verseTranslationFontChanged` similar to `verseFontChanged` for comparing the translation font style changes. Do let me know WDYT? 

Shukran

osamasayed (approved):

src/components/QuranReader/utils/memoization.ts
osamasayed: Looks great to me and good to merge! thanks a lot!

PR #792, opened 2021-11-19
Use Mushaf#11 for tajweed images

naveed-ahmad:
### Summary
- Fix tajweed word highlighting 
- Moved tajweed word to dls
- Use a better tajweed image
- Replace alt with Uthmani text, it is better for SEO, and users will see the real text if any image fails to load 

### Screenshots

| Before | After |
| ------ | ------ |


src/components/dls/QuranWord/QuranWord.tsx
```diff
@@ -7,11 +7,11 @@ import useSWRImmutable from 'swr/immutable';
 import GlyphWord from './GlyphWord';
 import onQuranWordClick from './onQuranWordClick';
 import styles from './QuranWord.module.scss';
+import TajweedWord from './TajweedWordImage';
```
osamasayed, line 10: `TajweedWordImage` will have to be renamed when we implement the Tajweed text instead of images so I would rather leave it as it is.

osamasayed (approved):

Issue #798, opened 2021-11-20
"Select language" does not have enough width


muhajirdev:
![image](https://user-images.githubusercontent.com/12745166/142731882-31dcbc19-a319-4456-94ab-f59c41344081.png)

- This should show "Indonesia" instead of "Indone"

iyunusov:
I am working on this.
Question arises with this issue and there're two simple ways I know for fixing it:
1. By setting max-width & min-width (following this again if there's some super long option name, it won't still appear fully, since we'll be still limiting the width at some point).
2. By calculated the box width depending on the option length and then inlining the css.

The first option seems to be more cleaner to me, since languages usually aren't very long (say not more than 20 letters)...

muhajirdev:
I don't have a strong opinion. option 1 looks good 👍

circlePulse:
Shouldn't this be closed?

iyunusov:
> Shouldn't this be closed?

PR still has questions.

PR #809, opened 2021-11-21
Select language does not have enough width #798 bug fixed

iyunusov:
### Summary
Width of Navbar Selectbox changed to 160px. 

### Screenshots

Screenshot 2021-11-22 at 01 36 44
Screenshot 2021-11-22 at 01 37 18

### DIFF
--- a/src/components/Navbar/LanguageSelector.module.scss
+++ b/src/components/Navbar/LanguageSelector.module.scss
@@ -1,8 +1,8 @@
 @use "src/styles/utility";
 @use "src/styles/breakpoints";
 
-$select-width: calc(6.3 * var(--spacing-small));
-$select-width-mobile: calc(7 * var(--spacing-small));
+$select-width: calc(10 * var(--spacing-medium));
+$select-width-mobile: calc(8 * var(--spacing-medium));
 
 .container {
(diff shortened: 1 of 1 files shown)


src/components/Navbar/LanguageSelector.module.scss
```diff
@@ -1,8 +1,8 @@
 @use "src/styles/utility";
 @use "src/styles/breakpoints";
 
-$select-width: calc(6.3 * var(--spacing-small));
```
muhajirdev, line 4: awesome,

in PR #800 , we already make the `bahasa melayu` shorter. I think we can reduce the `select-width` here. Maybe 6.5 spacing medium is a good width? What do you think @iyunusov ?

iyunusov: @muhajirdev We can reduce it to any size we want. I have made a little bit of research, seems the longest language name to exist is less than or around 20 characters, so with the current setting we are more on the safer side in case in future more languages will be added to the list...  On the other end, Navbar has only few items and there's plenty of empty space after selectbox, we could change `select-width` to `8` instead of `10`, but again any name longer than ex. Portugues which currently is 9 character name, would not appear fully.

PR #810, opened 2021-11-22
Toast status

mahmoudfelfel:
### Summary
Added different statuses for toast.
Screen Shot 2021-11-22 at 3 01 39 AM
Screen Shot 2021-11-22 at 3 01 51 AM

### DIFF
--- a/src/components/dls/Toast/Toast.tsx
+++ b/src/components/dls/Toast/Toast.tsx
@@ -11,6 +11,12 @@ import styles from './Toast.module.scss';
 
 import Button, { ButtonSize, ButtonType } from 'src/components/dls/Button/Button';
 
+enum ToastStatus {
+  Success = 'success',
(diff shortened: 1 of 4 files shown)


src/components/dls/Toast/Toast.tsx
```diff
@@ -18,6 +19,7 @@ type Action = {
 type Options = {
   preserve?: boolean;
   actions?: Action[];
+  status?: 'success' | 'error' | 'warning';
```
muhajirdev, line 22: Awesome! @mahmoudfelfel 

The codebase is already using enums for this use case. Might be a good idea to also use `enum` here too.

maybe we do `ToastStatus`.

other than that, looks good!.

mahmoudfelfel: done, thanks for the suggestion

PR #815, opened 2021-11-23
Add Tarteel VS authorization

piraka9011:
### Summary

We've introduced authorization for the Tarteel VS moving forward.
Another approach to this is using cookies and reading them server side, but that's no different than just using query params.

See [this article](https://faqs.ably.com/is-it-secure-to-send-the-access_token-as-part-of-the-websocket-url-query-params).

### Test Plan

Tested w/ QDC site and Tarteel VS locally.

### Screenshots

N/A


piraka9011:
Testing a bit more and this PR does not handle the case where the socket refuses connections well.

PR #819, opened 2021-11-24
fix QuranWord story

naveed-ahmad:
### Summary
TSIA

### Screenshots

| Before | After |
| ------ | ------ |


osamasayed:
@naveed-ahmad Looks good to me except for QPC V1 and V2 don't seem to show properly.
Screen Shot 2021-11-25 at 15 58 54
Screen Shot 2021-11-25 at 15 58 51

PR #840, opened 2021-11-30
homepage : use svg logo

muhajirdev:


src/components/HomePage/HomePageHero.tsx
```diff
@@ -16,9 +14,8 @@ const HomePageHero = () => {
         <PrayerTimes />
         <div className={styles.innerContainer}>
           <div className={styles.imageContainer}>
-            <Image className={styles.homepageImage} src={quranKarimLogo} priority />
+            <AlQuranulKarimSVG />
```
osamasayed, line 17: Shouldn't we keep using `Image` component to set `priority` so it's visible as fast as possible to the user since it's the very first part that user looks at?

tifa2UP: This is actually faster than using an image component. Because it's rendered inline and served within the html page

osamasayed (approved):

PR #850, opened 2021-12-01
feat: select reciter from audio player

muhajirdev:
image

image


src/components/dls/PopoverMenu/PopoverMenu.tsx
```diff
@@ -44,6 +45,7 @@ PopoverMenu.Item = ({
   onClick,
   isDisabled,
   shouldCloseMenuAfterClick = false,
+  shouldFlipIconRTL = false,
```
osamasayed, line 48: nit: I think it should be renamed to `shouldFlipOnRTL` which is already used in multiple places

src/components/AudioPlayer/Buttons/SelectReciterMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const SelectReciterMenu = ({ onBack }) => {
+  const { lang, t } = useTranslation('common');
+  const selectedReciter = useSelector(selectReciter);
+  const dispatch = useDispatch();
+
+  const reciters = (
+    <DataFetcher
+      queryKey={makeRecitersUrl()}
+      render={(data: RecitersResponse) => {
```
osamasayed, line 27: the render prop should be memoized

src/components/AudioPlayer/Buttons/SelectReciterMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import useTranslation from 'next-translate/useTranslation';
+import { useDispatch, useSelector } from 'react-redux';
+
+import CheckIcon from '../../../../public/icons/check.svg';
+import ChevronLeftIcon from '../../../../public/icons/chevron-left.svg';
+
+import DataFetcher from 'src/components/DataFetcher';
+import PopoverMenu from 'src/components/dls/PopoverMenu/PopoverMenu';
+// import { playbackRates } from 'src/components/Navbar/SettingsDrawer/AudioSection';
```
osamasayed, line 11: I think we should remove the comments and un-used code unless you think we will need it in the future

osamasayed (approved):

PR #853, opened 2021-12-01
add spacer utils for indopak words that has waqf sign

naveed-ahmad:
### Summary
- Renamed `CharTypeName` to `CharType`
- Removed unused `IndopakWord` component 
- Fixed spacing for `Indopak` words that has waqf sign

### Screenshots

| | |
| ------ | ------ |
|Before|before|
|After|after|


types/Word.ts
```diff
@@ -26,6 +26,7 @@ interface Word {
   textIndopak?: string;
   highlight?: string | boolean;
   text?: string;
+  cssClass?: string;
```
osamasayed, line 29: is `cssClass` coming from BE?

src/components/dls/QuranWord/TextWord.tsx
```diff
@@ -10,6 +10,7 @@ import { CharType } from 'types/Word';
 type MadaniWordTextProps = {
   text: string;
   font: string;
+  cssClass?: string;
```
osamasayed, line 13: 1. `cssClass` should be renamed to `className` for consistency for example see `PopoverMenuItemProps` or `ButtonProps`.
2. I cannot think of why we need this prop because we can add the styles directly inside `styles.word` since `TextWord` component is only used once and by only `QuranWord` component

src/styles/global.scss
```diff
@@ -40,3 +40,14 @@ textarea {
 a {
   color: var(--color-text-default);
 }
+
+// these spacer utilities are used to add hard coded spaces to components
+// names of these spacer classes is coming from the api.
+// One sample usage is for Indopak words, some words need extra space at the end(see 2:2:4)
+// detecting these spaces in FE is hard, and will slow down the app
```
osamasayed, line 47: I am not in favor of having this as a global class since it's only used inside `QuranText`. what I suggest doing is moving those 2 classes to `TextWord.module.scss` and to use them, you can do something like this inside `TextWord`:

`[styles[cssClass]]: cssClass` like the other existingcases with using different fonts

src/components/dls/QuranWord/TextWord.tsx
```diff
@@ -23,9 +24,9 @@ const UTHMANI_HAFS_FONTS = decamelizeKeys({
   textIndopak: INDO_PAK,
 });
 
-const TextWord: React.FC<MadaniWordTextProps> = ({ text, font, charType }) => (
+const TextWord: React.FC<MadaniWordTextProps> = ({ text, font, charType, cssClass }) => (
   <span
-    className={classNames(styles.word, {
+    className={classNames(styles.word, cssClass, {
```
osamasayed, line 29: Check the comment below regarding this

src/components/dls/QuranWord/QuranWord.tsx
```diff
@@ -85,8 +85,10 @@ const QuranWord = ({
     wordText = <GlyphWord font={font} text={getGlyph(word, font)} pageNumber={word.pageNumber} />;
   } else if (font === QuranFont.Tajweed) {
     wordText = <TajweedWord path={word.text} alt={word.textUthmani} />;
-  } else if (word.charTypeName !== CharType.Pause) {
-    wordText = <TextWord font={font} text={word.text} charType={word.charTypeName} />;
+  } else {
+    wordText = (
```
osamasayed, line 89: is `CharType.Pause` handled in this case?

osamasayed:
This change seems to break the layout on mobile:

| Before | After |
| ------ | ------ |
|261328983_676548159994438_776048454004367759_n|260366379_1509997792713444_2311557381047434840_n|

src/components/dls/QuranWord/QuranWord.tsx
naveed-ahmad: We don't need `CharType.Pause` anymore. It is removed from the BE, pause marks are already merged with their previous word.

We actually can remove `word.charTypeName`, we now have only `word` and `end`( ayah number) char types. And sending `char_type_name` attribute for each word is bit overkill IMO. 

We can just add `isEnd` to `end` words, just like we're sending `highlighted` to words that should be highlighted. Let me know if you've a better solution @osamasayed

naveed-ahmad:
Closing this in fav of https://github.com/quran/quran.com-frontend-next/pull/954

PR #863, opened 2021-12-03
Sidebar navigation

muhajirdev:
- TODO
    - [x] Click outside mobile 
       - [ ]  still need to check performance
    - [x] localization
    - [ ] refactor codes
    - [x] check RTL support
    - Styling
        - [x] Searchbox
        - [x] Make sure things are okay on light mode and and dark mode
        - [x] adjust the spacing when the thing scroll 
        - [x] Make sure things are okay on mobile
        - [x] items styling 
    - Minor
        - [x] switcher issue when closing the pane
        - [ ] highlight current page
        - [ ] highlight current juz
        - [ ] split the css
    - Optimization
        - [x] Lazy load
        - [ ]  component naming
        - [x] Split the files
        - [ ] optimize scrolling behavior

- Known issues
   - Clicking toggle on mobile does work as expected

- Will do in another PR
  - Switch DLS update


src/components/dls/KeyboardInput/KeyboardInput.module.scss
```diff
@@ -1,9 +1,15 @@
 @use "src/styles/breakpoints";
+@use "src/styles/theme";
 
 .container {
   direction: ltr;
   color: var(--color-text-default);
-  border: 1px solid var(--color-borders-hairline);
+  @include theme.light {
+    border: 1px solid var(--color-borders-hairline);
```
osamasayed, line 8: Both light and dark seem to use the same value `border: 1px solid var(--color-borders-hairline);`

src/redux/slices/QuranReader/sidebarNavigation.ts
```diff
@@ -0,0 +1,59 @@
...
+  selectedNavigationItem: string;
+};
+
+export enum NavigationItem {
+  Surah = 'surah',
+  Juz = 'juz',
+  Page = 'page',
+}
+
+export const navigationItems = [
```
osamasayed, line 16: I don't think this belongs here. I think this should be moved to `src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx`. Also the value of `name` field should be localized

src/redux/slices/QuranReader/sidebarNavigation.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const sidebarNavigationSlice = createSlice({
+  name: 'sidebarNavigation',
+  initialState,
+  reducers: {
+    setIsVisible: (state: SidebarNavigation, action: PayloadAction<boolean>) => ({
+      ...state,
+      isVisible: action.payload,
+    }),
+    selectNavigationItem: (state: SidebarNavigation, action: PayloadAction<string>) => ({
```
osamasayed, line 44: nit: `PayloadAction<string>` should be `PayloadAction<NavigationItem>`

src/utils/page.ts
```diff
@@ -46,3 +46,12 @@ export const isLastPage = (
   }
   return pageNumber === mushafTotalPages;
 };
+
+/**
+ * Return array of page id
+ *
+ * @returns {number[]}
+ */
+export const getPageIds = (mushaf: Mushaf = Mushaf.KFGQPCHAFS) => {
```
osamasayed, line 55: Rename to `getMushafPages` or `getPageIdsByMushaf`

src/components/QuranReader/SidebarNavigation/JuzSelection.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import useTranslation from 'next-translate/useTranslation';
+
+import styles from './SidebarNavigation.module.scss';
+
+import Link from 'src/components/dls/Link/Link';
+import { getJuzIds } from 'src/utils/juz';
+import { getJuzNavigationUrl } from 'src/utils/navigation';
+
+const JuzSelection = () => {
```
osamasayed, line 11: I feel the code is almost the same as `PageSelection`. Maybe we should have a component called `SidebarNavigationSelection` that accepts `SelectionType` and that handles both cases in-case we want to change the logic we don't need to change in 2 places

src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  navigationItems,
+  selectIsSidebarNavigationVisible,
+  selectNavigationItem,
+  selectSelectedNavigationItem,
+  NavigationItem,
+  setIsVisible,
+} from 'src/redux/slices/QuranReader/sidebarNavigation';
+
+const JuzSelection = dynamic(() => import('./JuzSelection'));
```
osamasayed, line 27: Should have a loading/skeleton indicator

src/components/QuranReader/SidebarNavigation/SurahList.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <div className={styles.surahListContainer}>
+      <input
+        className={styles.searchInput}
+        value={searchQuery}
+        onChange={(e) => setSearchQuery(e.target.value)}
+        placeholder={t('sidebar.search-surah')}
+      />
+      <div className={styles.list}>
+        {filteredChapters.map((chapter) => (
+          <Link href={`/${chapter.id}`}>
```
osamasayed, line 49: nit: use `getSurahNavigationUrl`

src/components/QuranReader/SidebarNavigation/SurahList.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const filteredSurah = fuse.search(searchQuery).map(({ item }) => item);
+  return filteredSurah as Chapter[];
+};
+
+const SurahList = () => {
+  const chapterIds = useChapterIdsByUrlPath();
+  const currentChapterId = chapterIds[0];
+  const { t } = useTranslation('common');
+
+  const chaptersData = getAllChaptersData();
```
osamasayed, line 29: Locale/lang needs to be passed to `getAllChaptersData` so that the Surah names will be localized. You can get the locale value from `const { t, lang } = useTranslation('common');`

src/components/QuranReader/SidebarNavigation/SurahList.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [searchQuery, setSearchQuery] = useState('');
+
+  const chapterDataArray = Object.entries(chaptersData).map(([id, chapter]) => ({
```
osamasayed, line 32: This should either be memoized or move to outside the functional component

src/components/QuranReader/SidebarNavigation/VerseList.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const VerseList = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const chapterIds = useChapterIdsByUrlPath();
+  const { t } = useTranslation('common');
+  if (!chapterIds || chapterIds.length === 0) return null;
+
+  const currentChapterId = chapterIds[0];
+
+  const verseKeys = generateChapterVersesKeys(currentChapterId);
```
osamasayed, line 20: This should me memoized because it will be re-run on every `searchQuery` change

osamasayed:
When it's RTL locale, the scroll bar because on the left side which is a bit weird experience in my opinion

Screen Shot 2021-12-07 at 21 16 46

src/hooks/useOutsideClickDetector.ts
```diff
@@ -11,9 +15,13 @@ const useOutsideClickDetector = (
   ref: React.RefObject<HTMLElement>,
   onClickOutsideDetected: () => void,
   enableDetection: boolean,
+  options?: Options,
 ) => {
   useEffect(() => {
     const handleClickOutside = (event) => {
+      // if maxWidth is specified and clientWidth is greater than maxWidth, do not trigger `onClickOutsideDetected`
+      if (options?.maxWidth && document.documentElement.clientWidth > options.maxWidth) return;
```
osamasayed, line 23: I have mixed feelings regarding this. I think it would be a better approach to leverage `enableDetection` instead of adding `options` because the logic of checking the width doesn't belong to the hook in my opinion so we can move this logic to `src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx` and instead of setting the value of `enableDetection` to `true`, we set the value to `document.documentElement.clientWidth > 768`

src/components/QuranReader/SidebarNavigation/ScrollableSelection.tsx
```diff
@@ -0,0 +1,32 @@
...
+        className={styles.searchInput}
+        value={searchQuery}
+        onChange={(e) => setSearchQuery(e.target.value)}
+        placeholder={searchPlaceholder}
+      />
+      <div className={styles.list}>
+        {items
+          .filter((item) => item.toString().startsWith(searchQuery))
+          .map((item) => (
+            <Link href={getPageNavigationUrl(item)}>
```
osamasayed, line 23: this needs to be dynamic as `juz` will have a different route than `page`

src/components/dls/KeyboardInput/KeyboardInput.module.scss
```diff
@@ -1,4 +1,5 @@
 @use "src/styles/breakpoints";
+@use "src/styles/theme";
```
osamasayed, line 2: nit: this is not used, we should remove it

src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+        {selectedNavigationItem === NavigationItem.Surah && <SurahSelection />}
+        {selectedNavigationItem === NavigationItem.Juz && <JuzSelection />}
+        {selectedNavigationItem === NavigationItem.Page && <PageSelection />}
+      </div>
+    </div>
+  );
+};
+
+const TABLET_WIDTH = 768;
+const isMobile = () => {
```
osamasayed, line 114: Nothing actionable now but if we needed this logic again in the future, we should move it to a util.

src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+  selectIsSidebarNavigationVisible,
+  selectNavigationItem,
+  selectSelectedNavigationItem,
+  NavigationItem,
+  setIsVisible,
+} from 'src/redux/slices/QuranReader/sidebarNavigation';
+
+const Loading = () => (
+  <div className={styles.loadingContainer}>
+    <Spinner />
```
osamasayed, line 29: No strong opinion but I feel a Skeleton would be better UI-wise

osamasayed (approved):

PR #865, opened 2021-12-04
Fix drawer issue on safari mobile

Mirodil:
### Summary
Fix drawer issue on safari mobile

### Test Plan
Test on mobile devices

### Screenshots

| Before | After |
| ------ | ------ |
| simulator_screenshot_9391A349-221C-426A-8CCF-002380AD89F0 | simulator_screenshot_30CAAFF5-987A-4068-819D-ADE8BD00AD75 |

### DIFF
--- a/src/components/Navbar/Drawer/Drawer.module.scss
+++ b/src/components/Navbar/Drawer/Drawer.module.scss
@@ -4,7 +4,6 @@
 .container {
   background: var(--color-background-elevated);
   position: fixed;
-  height: 100vh;
   width: 100%;
   inset-block: 0;
   z-index: var(--z-index-header);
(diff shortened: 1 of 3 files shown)


Mirodil:
fix hero image on safari mobile

### Screenshots
| Before | After |
| ------ | ------ |
| simulator_screenshot_922BA21E-C06C-46CB-B69F-01C53342F342 | simulator_screenshot_41DC468A-4616-47F0-8E2F-9D4E95B9777E |

muhajirdev:
Salam @Mirodil
thank you so much for your first contribution. May Allah reward you.

src/components/Navbar/Drawer/Drawer.module.scss
```diff
@@ -13,17 +12,19 @@
...
   overflow-x: hidden;
   overscroll-behavior-y: contain;
+  top: 0;
+  bottom: 0;
   @include breakpoints.tablet {
     width: constants.$side-menu-desktop-width;
   }
   &.left {
-    inset-inline-start: -100%;
+    left: -100%;
```
muhajirdev, line 21: We're using css logical properties btw, so `left` will automatically becoems `right` on RTL https://css-tricks.com/css-logical-properties/

Mirodil:
@muhajirdev It does not work on iPhone XP with iOS version 13. It does not work on iOS 14.x also.

src/components/Navbar/Drawer/Drawer.module.scss
Mirodil: I got it but those props does not supported by most iOS devices. Which makes those props useless unless it would be supported

Mirodil:
I am going to close PR as there is not interest on this fix.
Thank you.

PR #837, opened 2021-11-28
Localized numbers, and dates

mahmoudfelfel:
### Summary
Added localization for numbers and dates.
Note: Plurals are still not very accurate in some places due to the stitching we are doing between strings and numbers but we can fix that in another PR.

Closes #831 

### Screenshots
Screen Shot 2021-11-29 at 4 00 46 AM
Screen Shot 2021-11-29 at 1 16 58 AM
Screen Shot 2021-11-29 at 1 17 28 AM
Screen Shot 2021-11-29 at 1 36 36 AM
Screen Shot 2021-11-29 at 2 15 22 AM
Screen Shot 2021-11-29 at 2 19 36 AM
Screen Shot 2021-11-29 at 2 23 42 AM
Screen Shot 2021-11-28 at 11 30 25 PM
Screen Shot 2021-11-28 at 11 31 57 PM
Screen Shot 2021-11-28 at 11 36 40 PM
Screen Shot 2021-11-28 at 11 36 55 PM
Screen Shot 2021-11-28 at 11 37 10 PM

PR #876, opened 2021-12-08
fix chapter header number

muhajirdev:
image
image


muhajirdev:
closing this PR, will be fixed in #837

PR #890, opened 2021-12-10
fix: font sizing issues

muhajirdev:


src/components/QuranReader/TafsirView/TafsirView.module.scss
```diff
@@ -1,13 +1,17 @@
...
 
 .container {
-  max-width: 80%;
+  max-width: 90%;
   margin-block-start: var(--spacing-small);
   margin-block-end: var(--spacing-small);
   margin-inline-start: auto;
   margin-inline-end: auto;
   border-bottom: 1px solid var(--color-primary-medium);
+  @include breakpoints.tablet {
```
muhajirdev, line 12: usually when there are > 100 characters per line. The text becomes less comfortable to read. I forgot where I read this. 

But you can see how medium.com do this. They are aimin +- 80 character per line.

So we're limiting the max-width here, the number of characters per line is < 100

src/components/QuranReader/ReadingView/Page.module.scss
```diff
@@ -6,10 +6,7 @@
   margin-block-start: var(--spacing-xsmall);
   margin-block-end: var(--spacing-mega);
   max-width: 100%;
-  min-width: 100%;
-  @include breakpoints.tablet {
-    min-width: calc(10 * var(--spacing-mega));
-  }
+  min-width: calc(9.5 * var(--spacing-mega));
```
tifa2UP, line 9: nit: move to const

src/components/QuranReader/TafsirView/TafsirView.module.scss
```diff
@@ -1,13 +1,17 @@
...
+    max-width: calc(23 * var(--spacing-mega));
```
tifa2UP, line 13: move to const

src/components/Verse/VerseText.module.scss
```diff
@@ -40,8 +40,20 @@
...
+@include utility.generate-font-sizes("code_v2");
+@include utility.generate-font-sizes("text_indopak");
+@include utility.generate-font-sizes("qpc_uthmani_hafs");
 
 .highlighted {
   background-color: var(--color-background-alternative-faded);
 }
+
+// make the text slightly bigger on translation or tafsir mode
+.tafsirOrTranslationMode {
```
tifa2UP, line 53: So the arabic font size in translation mode is basically 1.2x what's on the reading mode?

muhajirdev: correct, on mobile.

PR #906, opened 2021-12-12
improve navbar, context & audio player menu scrolling experience

muhajirdev:
Cannot provide screenshot.
you need to compare the experience directly on your phone

### DIFF
--- a/src/components/AudioPlayer/AudioPlayer.module.scss
+++ b/src/components/AudioPlayer/AudioPlayer.module.scss
@@ -24,6 +24,7 @@
   @include breakpoints.tablet {
     height: constants.$audio-player-default-desktop-height;
   }
+  will-change: transform;
 }
 
 .containerHidden {
--- a/src/components/AudioPlayer/AudioPlayer.module.scss
+++ b/src/components/AudioPlayer/AudioPlayer.module.scss
@@ -33,4 +33,5 @@
 
 .containerMinimized.containerDefault {
   transform: translate3d(0, constants.$audio-player-minimized-height, 0);
+  // https://ptgamr.github.io/2014-09-13-translate3d-vs-translate-performance/
 }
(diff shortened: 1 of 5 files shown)


src/components/AudioPlayer/AudioPlayer.module.scss
```diff
@@ -31,5 +31,5 @@
 }
 
 .containerMinimized.containerDefault {
-  height: constants.$audio-player-minimized-height;
+  transform: translate3d(0, constants.$audio-player-minimized-height, 0);
```
muhajirdev, line 34: based on some resource, it says translate3d is faster than `translate`.
but I will double check

src/components/AudioPlayer/AudioPlayerSlider.module.scss
```diff
@@ -51,7 +51,6 @@
   width: 100%;
   transform: translate(-50%, 0);
   position: fixed;
-  inset-block-end: constants.$audio-player-minimized-height;
```
muhajirdev, line 54: the slider was not working on next.quran.com. deleting this fixed the issue

src/components/AudioPlayer/AudioPlayer.module.scss
tifa2UP: Nice, we should add this comment in the code

src/components/AudioPlayer/AudioPlayer.module.scss
```diff
@@ -24,12 +24,13 @@
   @include breakpoints.tablet {
     height: constants.$audio-player-default-desktop-height;
   }
+  will-change: transform;
```
tifa2UP, line 27: There are a bunch of warning about will change: https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

I think that using it is still the right call given how frequently this animation happens.

tifa2UP (approved):

PR #918, opened 2021-12-14
design update: update secondary color on dark theme

muhajirdev:
This is just a temporary fix. I will do a more systematic update related to color token later.

Before
image

After
image

### DIFF
--- a/src/components/dls/Button/Button.module.scss
+++ b/src/components/dls/Button/Button.module.scss
@@ -313,7 +313,7 @@
 }
 
 // when button used as Popover's trigger, add style when Popover is open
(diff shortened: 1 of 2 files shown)


src/components/dls/Button/Button.module.scss
```diff
@@ -313,7 +313,7 @@
 }
 
 // when button used as Popover's trigger, add style when Popover is open
-[data-state="open"] .base {
+[data-state="open"] > button > .base {
```
muhajirdev, line 316: this fix the previous bug caused by #911 

The bug was :
- in repeat settings. the + button was highlighted.

PR #951, opened 2021-12-17
Adding  ﷺ after prophet ﷺ's name in multiple languages

faiqali1:
### Summary
Adding ﷺ after the prophet ﷺ's name in multiple languages including:
-  English
-  French
-  Indonesian
-  Italian
-  Malay
-  Dutch
-  Turkish


muhajirdev:
Halo @faiqali1 , thanks a lot for the contribution and the effort. May Allah reward you.

We're using a localization service called https://lokalise.com, this json will be updated via lokalise as you already submitted a PR to update the english version previously.

so, I am closing this PR for now.

faiqali1:
I see, so with service you only need to change the english version and the others are updated automatically?

PR #960, opened 2021-12-19
Adds some cosmetic improvements for the translation view.

sharabash:
### Summary
- Adds some aesthetic styling to the highlighted container
- Hides the ayah number glyph/character
- Removes the "Surah" prefix at the top
- Fixes the word gap occurring on the first two words

### Screenshots
Screen Shot 2021-12-19 at 7 48 02 AM
Screen Shot 2021-12-19 at 7 48 52 AM

### DIFF
--- a/src/components/QuranReader/TranslationView/TranslationViewCell.module.scss
+++ b/src/components/QuranReader/TranslationView/TranslationViewCell.module.scss
@@ -1,7 +1,12 @@
 @use "src/styles/breakpoints";
 
 .highlightedContainer {
-  background: var(--color-background-alternative-faded);
+  background: var(--gradient-background-highlighted-container);
+  background-color: var(--color-background-alternative-faded);
+  margin: 0 var(--negative-margin-medium);
+  padding-inline-start: var(--spacing-medium) !important;
+  padding-inline-end: var(--spacing-medium) !important;
+  border-radius: var(--spacing-xsmall);
 }
 
 .cellContainer {
--- a/src/components/QuranReader/TranslationView/TranslationViewCell.tsx
+++ b/src/components/QuranReader/TranslationView/TranslationViewCell.tsx
@@ -112,7 +112,7 @@ const TranslationViewCell: React.FC<TranslationViewCellProps> = ({
           </div>
         </div>
       </div>
-      <Separator />
+      <Separator isTranslationView />
     </div>
   );
 };
--- a/src/components/chapters/ChapterHeader/index.tsx
+++ b/src/components/chapters/ChapterHeader/index.tsx
@@ -53,7 +53,7 @@ const ChapterHeader: React.FC<Props> = ({ chapterId, pageNumber, hizbNumber }) =
         <div className={styles.left}>
           {!isMinimalLayout && <div className={styles.translatedName}>{translatedName}</div>}
           <div className={styles.transliteratedName}>
-            {t('surah')} <br /> {transliteratedName}
+            {transliteratedName}
           </div>
           <div className={styles.infoContainer}>
             <Button
--- a/src/components/dls/QuranWord/QuranWord.module.scss
+++ b/src/components/dls/QuranWord/QuranWord.module.scss
@@ -2,8 +2,8 @@
   display: inline-block;
   cursor: pointer;
 }
-.additionalWordGap + .additionalWordGap {
-  margin-inline-end: var(--spacing-xxsmall);
+.additionalWordGap:first-child, .additionalWordGap + .additionalWordGap {
+  margin-inline-end: var(--spacing-nano);
 }
 
 .tajweedWord {
--- a/src/components/dls/Separator/Separator.module.scss
+++ b/src/components/dls/Separator/Separator.module.scss
@@ -1,5 +1,22 @@
 $separator-size: 1px;
 
+.soft {
+  position: relative;
+  top: 1px;
+
+  &[data-orientation="horizontal"] {
+    background: var(--gradient-borders-horizontal-hairline);
+    height: $separator-size;
+    width: 100%;
+  }
+
+  &[data-orientation="vertical"] {
+    background: var(--gradient-borders-vertical-hairline);
+    height: 100%;
+    width: $separator-size;
+  }
+}
+
 .base {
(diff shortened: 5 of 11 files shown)


src/styles/theme.scss
```diff
@@ -5,6 +5,7 @@
 
 /** System theme (auto-detects OS light/dark mode), other themes extend this one */
 :root {
+  --spacing-nano: 0.125rem;
```
osamasayed, line 8: We did extensive research to come up with the upper and lower values of the spacing. What we usually do if we want to go below `micro` spacing (very rare cases) is to use [`calc`](https://developer.mozilla.org/en-US/docs/Web/CSS/calc()) e.g. `calc(0.5 * var(--spacing-micro))`

src/components/chapters/ChapterHeader/index.tsx
```diff
@@ -52,9 +52,7 @@ const ChapterHeader: React.FC<Props> = ({ chapterId, pageNumber, hizbNumber }) =
       <div className={styles.container}>
         <div className={styles.left}>
           {!isMinimalLayout && <div className={styles.translatedName}>{translatedName}</div>}
-          <div className={styles.transliteratedName}>
-            {t('surah')} <br /> {transliteratedName}
-          </div>
+          <div className={styles.transliteratedName}>{transliteratedName}</div>
```
osamasayed, line 55: Are we meant to remove the `Surah` wording?

src/components/QuranReader/TranslationView/TranslationViewCell.tsx
```diff
@@ -112,7 +112,7 @@ const TranslationViewCell: React.FC<TranslationViewCellProps> = ({
           </div>
         </div>
       </div>
-      <Separator />
+      <Separator isTranslationView />
```
osamasayed, line 115: Since `Separator` is a general component and not necessarily related to the QuranReader or Translation view, we shouldn't have this param. What we can do is to add an optional `className` prop that we can pass the value of the className we want to be applied to the Seperator e.g. <Seperator className={styles.faded} and we will consume the prop `className` inside the `Seperator` component. Something like this:

```
<SeparatorPrimitive.Separator
    orientation={isVertical ? 'vertical' : 'horizontal'}
    className={classNames(styles.base,{[className]: className})}
  />
  ```
  
  You can check `Skeleton` component for reference.

File: src/components/dls/QuranWord/QuranWord.module.scss [hunk 4]
```diff
@@ -2,8 +2,8 @@
   display: inline-block;
   cursor: pointer;
 }
-.additionalWordGap + .additionalWordGap {
-  margin-inline-end: var(--spacing-xxsmall);
+.additionalWordGap:first-child, .additionalWordGap + .additionalWordGap {
+  margin-inline-end: var(--spacing-nano);
```
osamasayed, line 6: Please see my other comment about using `nano`

src/components/QuranReader/TranslationView/TranslationViewCell.module.scss
```diff
@@ -1,7 +1,26 @@
...
+  }
+
+  [data-theme="light"] & {
+    --color-gradient-background-light: 60, 120, 180;
+    background: radial-gradient(ellipse farthest-corner at -360% -360%, rgba(var(--color-gradient-background-light), 0.45) 50%, rgba(var(--color-gradient-transition), 0) 90%);
+  }
+
+  background-color: var(--color-background-alternative-faded);
+  margin: 0 calc(var(--spacing-medium) * -1);
+  border-radius: var(--spacing-xsmall);
```
osamasayed, line 18: We have pre-set tokens for `border-radius`
Screen Shot 2021-12-20 at 18 51 27

src/components/dls/Separator/Separator.module.scss
```diff
@@ -1,5 +1,44 @@
 $separator-size: 1px;
 
+.faded {
```
osamasayed, line 3: Changing the separator styling will result in visual changes to any place we are using this component:

Screen Shot 2021-12-20 at 18 54 19

these values should be passed from the parent component wanting this change as the value of `className` prop. Please see my comment above on `TranslationViewCell.tsx`.

src/components/QuranReader/TranslationView/TranslationViewCell.module.scss
```diff
@@ -1,7 +1,26 @@
 @use "src/styles/breakpoints";
 
 .highlightedContainer {
-  background: var(--color-background-alternative-faded);
+  --color-gradient-transition: 196, 196, 196;
+
+  [data-theme="dark"] & {
```
osamasayed, line 6: We have a partial called `_theme.scss` that support dark and light themes. To use this, please refer to `Skeleton.module.scss`lines 19 and 22.

Screen Shot 2021-12-20 at 18 57 24

src/components/QuranReader/TranslationView/TranslationViewCell.module.scss
```diff
@@ -1,7 +1,26 @@
...
+    --color-gradient-background-dark: 150, 125, 225;
+    background: radial-gradient(ellipse farthest-corner at -360% -360%, rgba(var(--color-gradient-background-dark), 0.25) 50%, rgba(var(--color-gradient-transition), 0) 90%);
```
osamasayed, line 8: We already have pre-defined tokens for background colors for each theme:

Screen Shot 2021-12-20 at 18 58 24

sharabash:
Thanks for the feedback Osama. I will correct all these things and give them back to you again for another review.

About the 'Surah' prefix, I don't like it. The right side simply says the surah name in arabic, so why should the left side say "surah ...." instead of just the name? And then of course if you are on Surat al-Anfal it should say Surat al-Anfal, not Surah al-Anfal, and the same goes for any other surah which is a proper noun. Since the translation is in english, should we say "Chapter of al-Anfal" ? No, and for non-Muslims visiting the site it's just an extra thing that they don't understand immediately.

It's just taking up real estate and looks un-aesthetic with it's <br/> line break, it doesn't add any value

Make sense?
cc @tifa2UP about my opinion

PR #964, opened 2021-12-20
Create EmbeddableContent dls

muhajirdev:
Create EmbeddableContent dls that will be used for tafsirView
image

image


src/components/dls/Button/Button.module.scss
```diff
@@ -313,7 +313,8 @@
 }
 
 // when button used as Popover's trigger, add style when Popover is open
-[data-state="open"] > button > .base {
-  background-color: var(--color-primary-medium);
-  @include utility.lighten-background-color;
-}
+// TODO: removing this for now, buggy. Need to a better implementation
```
tifa2UP, line 316: Not sure what's happening here

src/components/dls/EmbeddableContent/EmbeddableContent.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+const EmbeddableContent = ({ children, isOpen, onClose }: EmbeddableContentProps) => {
+  return (
+    <Dialog.Root open={isOpen}>
+      <Dialog.Portal>
+        <Dialog.Overlay className={styles.overlay}>
+          <Dialog.Content className={styles.content} onInteractOutside={onClose}>
+            <Dialog.Close className={styles.closeIcon}>
+              <Button variant={ButtonVariant.Ghost} shape={ButtonShape.Circle} onClick={onClose}>
+                <CloseIcon />
```
tifa2UP, line 21: Close button should be optional, it should be visible by default

tifa2UP (approved):

src/components/dls/Button/Button.module.scss
muhajirdev: image

this is related to this nice to have effect (see the globe icon), when the popover menu is open.

will need to do a follow up PR to fix this.

PR #974, opened 2021-12-23
sort reciter by name

muhajirdev:
image

image

### DIFF
--- a/src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
+++ b/src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
@@ -59,22 +59,28 @@ const SettingsReciter = () => {
             : data.reciters;
           return (
             <div>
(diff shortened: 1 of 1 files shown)


src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
```diff
@@ -59,22 +59,28 @@ const SettingsReciter = () => {
...
-                    onChange={(e) => {
-                      onSelectedReciterChange(e.target.value, data.reciters);
-                    }}
-                  />
-                  <span>{reciter.name}</span>
-                  <span className={styles.recitationStyle}>{reciter.style.name}</span>
-                </label>
-              ))}
+              {filteredReciters
+                .sort((a, b) => (a.name > b.name ? 1 : -1))
```
tifa2UP, line 63: This will run on every render, might be cleaner to do do the sorting and mapping in a useMemo

PR #982, opened 2021-12-23
[Logging] Add logging to homepage

osamasayed:


src/components/CommandBar/index.tsx
```diff
@@ -21,27 +22,45 @@ const CommandBarBody = dynamic(() => import('./CommandBarBody'), {
...
+};
+
 const CommandBar: React.FC = () => {
   const dispatch = useDispatch();
   const isOpen = useSelector(selectCommandBarIsOpen, shallowEqual);
   const toggleShowCommandBar = useCallback(
     (event: KeyboardEvent) => {
+      getPressedShortcut(event);
+      // eslint-disable-next-line i18next/no-literal-string
+      logEvent(`command_bar_${isOpen ? 'close' : 'open'}_${getPressedShortcut(event)}`);
```
tifa2UP, line 42: nit: imo this is the case where the pressed shortcut should be meta data, not in the name. cmd vs. ctrl isn't important to know on a top level, we probably care more about the aggregate keyboard triggers. lmk if this makes sense

tifa2UP (approved):

PR #985, opened 2021-12-23
Update Radix Dialog package

osamasayed:


tifa2UP (approved):

src/components/CommandBar/CommandBarBase/CommandBarBase.module.scss
```diff
@@ -99,4 +100,5 @@ $min-height: calc(9 * var(--spacing-mega));
       animation: overlayHide var(--transition-fast) ease;
     }
   }
+  z-index: var(--z-index-toast);
```
tifa2UP, line 103: this should be called something else. like z-index-modal-overlay or z-index-overlay

src/components/dls/Modal/Modal.module.scss
```diff
@@ -26,6 +26,7 @@
   }
   position: fixed;
   inset: 0;
+  z-index: var(--z-index-toast);
```
tifa2UP, line 29: ditto, this shouldn't be a toast cause it's not a toast

PR #990, opened 2021-12-24
[Logging] Settings drawer Reciters

osamasayed:


src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
```diff
@@ -40,6 +44,8 @@ const SettingsReciter = () => {
   const onSelectedReciterChange = (reciterId: string, reciters: Reciter[]) => {
     if (!reciterId) return;
     const reciter = reciters.find((r) => r.id === Number(reciterId));
+    logValueChange('reciter', selectedReciter.id, reciter.id);
+    logItemSelectionChange('reciter', reciter.id);
```
tifa2UP, line 48: 1. since this is a single check, maybe logItemSelectionChange is suffice
2. maybe call the variable 'selected_reciter'?

PR #995, opened 2021-12-24
Fix bismillah font family

muhajirdev:


src/components/dls/Bismillah/Bismillah.module.scss
```diff
@@ -1,6 +1,7 @@
+$bismillah-font-size: 3rem;
```
muhajirdev, line 1: when we use the bismillah font family, 
the font size becomes, smaller. So increase the font size a bit more here.

it's actually bigger then the one that we have next.quran.com now, but I think this one looks a bit better. just IMO

PR #1008, opened 2021-12-27
use language specific font for tafsir content

naveed-ahmad:
### Summary

Use profer font for tafsir content, we've already defined fonts for each language. Let's use them.

### Screenshots

| Before | After |
| ------ | ------ |


muhajirdev (approved):
tested on my end

PR #1009, opened 2021-12-27
50% width was breaking the layout on small screens

naveed-ahmad:
### Screenshots

| Before | After |
| ------ | ------ |


osamasayed:
Screen Shot 2021-12-28 at 10 32 10
It breaks on RTL layout

PR #1028, opened 2021-12-29
[Logging] unsupported logical css properties

osamasayed:
### Summary
This PR adds logging when the user's browser doesn't support [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties).

### DIFF
--- a/src/utils/css.ts
+++ b/src/utils/css.ts
@@ -1,7 +1,7 @@
 import { logEvent } from './eventLogger';
 
 /* eslint-disable import/prefer-default-export */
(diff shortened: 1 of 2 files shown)


src/utils/css.ts
```diff
@@ -0,0 +1,11 @@
+import { logEvent } from './eventLogger';
+
+/* eslint-disable import/prefer-default-export */
+export const checkForCSSLogicalUnsupported = () => {
```
tifa2UP, line 4: logUnsupoortedLogicalCSS

PR #1058, opened 2022-01-01
Added theme switcher in navbar

Wasehdev:
In the navbar, there was no option to switch the theme for users, So have added theme switcher in navbar

Screenshots

 **Before**
image

 **After**
image


osamasayed:
Alsalamu Alikum, the team has thought about it and we think that putting a theme switcher in the navbar makes the navbar too tight on mobile particularly the ones with small screens like the image below of an iPhone 4. Jazak Allah Khairan for your contribution.

Screen Shot 2022-08-23 at 1 51 39 PM

Wasehdev:
Very grateful for considering. I've the same thought now.

PR #1059, opened 2022-01-02
feat: sort translations

muhajirdev:
## Before
image

## After
image


Ayman24x7:
Looking great but two things I noticed

image

PR #1071, opened 2022-01-05
reading view skeleton

muhajirdev:


src/components/QuranReader/ReadingView/ReadingPreviewSkeleton.module.scss
```diff
@@ -0,0 +1,17 @@
+@use "src/styles/breakpoints";
+@use "src/styles/utility";
+@use "src/styles/theme";
+
+.skeleton {
+  width: 100%;
+  min-width: calc(9.5 * var(--spacing-mega));
+  max-width: calc(15 * var(--font-size));
```
muhajirdev, line 8: will need to do follow up PR about the max-width when we implement the windowing

src/components/QuranReader/ReadingView/ReadingViewSkeleton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { shallowEqual, useSelector } from 'react-redux';
+
+import styles from './ReadingPreviewSkeleton.module.scss';
+
+import Skeleton from 'src/components/dls/Skeleton/Skeleton';
+import { selectQuranReaderStyles } from 'src/redux/slices/QuranReader/styles';
+import QuranReaderStyles from 'src/redux/types/QuranReaderStyles';
+
+const ReadingPreviewSkeleton = () => {
+  const numberOfLines = 15;
```
muhajirdev, line 12: will need to do follow up PR related to how many skeleton we want to render

src/components/QuranReader/index.tsx
```diff
@@ -140,6 +141,20 @@ const QuranReader = ({
   };
   const hasMore = size < getPageLimit(isVerseData, initialData);
 
+  let loader;
+  // if (readingPreference === ReadingPreference.Translation) {
+  //   loader = <TranslationViewSkeleton />;
+  // } else
+  if (readingPreference === ReadingPreference.Reading) {
```
tifa2UP, line 148: Needs to be reverted

src/components/QuranReader/ReadingView/ReadingViewSkeleton.module.scss
```diff
@@ -0,0 +1,17 @@
+@use "src/styles/breakpoints";
```
tifa2UP, line 1: reuse styles from the actual component

src/components/QuranReader/ReadingView/ReadingViewSkeleton.tsx
tifa2UP: The skeleton always shows a full page. Number of lines should read from redux

src/components/QuranReader/ReadingView/ReadingViewSkeleton.module.scss
muhajirdev: we only have one class here `.skeleton`. It's a bit hard to reuse the styling here since
1)  we dont use the actualy verse component
2) the verse component has a lot of whitespace. It will increase the skeleton size. bad perceived size.
image

tifa2UP (approved):

PR #1116, opened 2022-01-19
Remove unused Tarteel search response fields no longer sent by api

mmmoussa:
### Summary
In the Tarteel search backend we had some deprecated fields included in the search result matches. We have now fully removed those fields from the response. Those fields were already unused on quran.com, but the type definition needs to be updated.

### Test Plan
Inspected the code to understand whether any of these fields were being used. Will test with the updated backend response before merging. Opening now for early review.


mmmoussa:
@tifa2UP please TAL

PR #1131, opened 2022-01-22
Move fonts to CDN

naveed-ahmad:
### Summary
TSIA


src/components/Fonts/FontPreLoader.tsx
```diff
@@ -2,17 +2,15 @@ import React from 'react';
...
 const LOCALE_PRELOADED_FONTS = {
-  [DEFAULT_LOCALE]: [
-    { type: 'font/woff2', location: 'ProximaVara/proxima_vara_regular.woff2' },
-    { type: 'font/woff2', location: 'ProximaVara/proxima_vara_semibold.woff2' },
-    { type: 'font/woff2', location: 'ProximaVara/proxima_vara_bold.woff2' },
-  ],
-  ar: [{ type: 'font/woff2', location: 'urdu/NotoNaskhArabic-Regular.woff' }],
-  bn: [{ type: 'font/ttf', location: 'NotoSerifBengali/NotoSerifBengali-Regular.ttf' }],
-  ur: [{ type: 'font/woff2', location: 'urdu/NotoNaskhArabic-Regular.woff' }],
+  [DEFAULT_LOCALE]: [{ type: 'font/woff2', location: 'lang/ProximaVara/ProximaVara.woff2' }],
```
osamasayed, line 10: We were using proxima_vara_semibold and proxima_vara_bold intentionally for non-default `font-weight` values.

osamasayed, line 10: We were using `proxima_vara_semibold` and `proxima_vara_bold` intentionally for non-default `font-weight` values.

src/styles/global.scss
```diff
@@ -14,17 +14,43 @@ body {
   }
 
   overscroll-behavior-y: none;
-  font-family: "ProximaVara", "Helvetica Neue", Helvetica, Arial;
+  font-family: "Helvetica Neue", Helvetica, Arial;
```
osamasayed, line 17: Why are we we removing `ProximaVara` from here here?

src/styles/fonts.scss
```diff
@@ -1,96 +1,102 @@
...
+// Fallback for browsers that don't support variable fonts
 @font-face {
-  font-family: "NotoSerifBengali";
-  src: url("/fonts/NotoSerifBengali/NotoSerifBengali-Regular.ttf");
+  font-family: "ProximaVara";
+  src: local("Proxima Vara"),
+    url("#{$font-cdn}/lang/ProximaVara/ProximaVara.woff2");
+  font-weight: 100 900;
+  /* default 400 */
+  font-stretch: 50% 100%;
```
osamasayed, line 99: Why do we need this?

osamasayed:
@naveed-ahmad I suggest we remove the font files from the repo since we are not longer using them.

naveed-ahmad:
@osamasayed https://github.com/quran/quran.com-frontend-next/pull/1139

src/components/Fonts/FontPreLoader.tsx
osamasayed: It seems the new font being used is less in size that the 3 files we used to use collectively:

Before:
Screen Shot 2022-01-25 at 09 39 32

After:
Screen Shot 2022-01-25 at 09 39 27

osamasayed (approved):

PR #1141, opened 2022-01-25
Add sidebar navigation skeleton

### LINKED ISSUE #896: Skeleton for SidebarNavigation
image

Right now we're using Spinner to indicate that the SidebarNavigation is loading. But we can make it looks better by using skeleton

ahmedriad1:
### Summary
resolves #896 

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |

### DIFF
--- a/src/components/QuranReader/SidebarNavigation/SidebarSelectionSkeleton.tsx
+++ b/src/components/QuranReader/SidebarNavigation/SidebarSelectionSkeleton.tsx
@@ -2,7 +2,7 @@ import styles from './SidebarSelectionSkeleton.module.scss';
 
 import Skeleton from 'src/components/dls/Skeleton/Skeleton';
 
-const ROWS_COUNT = 6;
+const ROWS_COUNT = 12;
 const rowsArr = Array(ROWS_COUNT).fill(null);
 
 const SidebarSelectionSkeleton = () => {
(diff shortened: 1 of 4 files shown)


src/components/QuranReader/SidebarNavigation/SidebarSelectionSkeleton.tsx
```diff
@@ -0,0 +1,19 @@
+import styles from './SidebarSelectionSkeleton.module.scss';
+
+import Skeleton from 'src/components/dls/Skeleton/Skeleton';
+
+const ROWS_COUNT = 6;
```
muhajirdev, line 5: Salam brother @ahmedriad1 , jazakAllah for helping.

I think we'll need more than 6 row? For example here's how the sidebar height on my laptop, I think 6 will not cover the sidebar from top to bottom.
image

What do you think?

ahmedriad1: Salam bro @muhajirdev

Yeah should probably be. I used the settings drawer skeleton as a reference, and here is how it looked on larger resolutions: 
image

let me know what do you think, thanks

muhajirdev: I see, we probably should update the settings skeleton as well in the future.

PR #1158, opened 2022-01-28
use DLS radio group in reciter settings

ahmedriad1:
### Summary
- use DLS radio group in reciter selection settings
- accept `renderItem` prop in `RadioGroup`
- add `Item` & `Indicator` components in `RadioGroup` to allow further customization

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |

### DIFF
--- a/src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
+++ b/src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
@@ -70,40 +70,32 @@ const SettingsReciter = () => {
           const filteredReciters = searchQuery
             ? filterReciters(data.reciters, searchQuery)
             : data.reciters;
-          return (
-            <RadioGroup
-              value={selectedReciter.id.toString()}
-              orientation={RadioGroupOrientation.Vertical}
-              label="reciter"
-              onChange={(newId) => onSelectedReciterChange(newId, data.reciters)}
-              items={filteredReciters
-                .sort((a, b) => (a.name > b.name ? 1 : -1))
-                .map((reciter) => ({
-                  value: reciter.id.toString(),
-                  id: reciter.id.toString(),
-                  label:
-                    reciter.style.name !== DEFAULT_RECITATION_STYLE
-                      ? `${reciter.translatedName.name}/${reciter.style.name}`
-                      : reciter.translatedName.name,
-                }))}
-              renderItem={(item) => {
-                const [reciterName, reciterStyle] = item.label.split('/');
 
-                return (
-                  <div className={styles.reciter} key={item.id}>
-                    <RadioGroup.Item value={item.value} id={item.id}>
-                      <RadioGroup.Indicator />
-                    </RadioGroup.Item>
-                    <label htmlFor={item.id} className={styles.reciterLabel}>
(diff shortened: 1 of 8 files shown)


src/components/dls/Forms/RadioGroup/RadioGroup.tsx
```diff
@@ -32,9 +32,13 @@ interface Props {
   required?: boolean;
   loop?: boolean;
   orientation?: RadioGroupOrientation;
+  renderItem?: (item: RadioItem) => React.ReactNode;
 }
 
-const RadioGroup: React.FC<Props> = ({
+const RadioGroup: React.FC<Props> & {
+  Item: (props: RadioGroupPrimitive.RadioGroupItemProps) => React.ReactElement;
+  Indicator: (props: RadioGroupPrimitive.RadioIndicatorProps) => React.ReactElement;
```
muhajirdev, line 40: Salam bro @ahmedriad1 , thanks making this PR

```
Item: (props: RadioGroupPrimitive.RadioGroupItemProps) => React.ReactElement;
Indicator: (props: RadioGroupPrimitive.RadioIndicatorProps) => React.ReactElement;
```

is this part required. In Modal.tsx, we don't have typing. Maybe we should follow the same approach? what do you think?

src/components/Navbar/SettingsDrawer/ReciterSelectionBody.tsx
```diff
@@ -70,34 +71,39 @@ const SettingsReciter = () => {
...
-                  </label>
-                ))}
-            </div>
+                .map((reciter) => ({
+                  value: reciter.id.toString(),
+                  id: reciter.id.toString(),
+                  label:
+                    reciter.style.name !== DEFAULT_RECITATION_STYLE
+                      ? `${reciter.translatedName.name}/${reciter.style.name}`
+                      : reciter.translatedName.name,
```
muhajirdev, line 87: hmm, i see. So what we're doing here is basically we put the 2 informations inside a string and we seperate it with a  `/` right.

I know the current dls doesn't support rendering the recitation style in a radio component. 

Would it be better to expose the RadioGroup root component as well? (now that we already expose `.Item` and `.Indicator`)

So we can do 

```
<RadioGroup.Root>
   {filteredReciters.map( reciter =>
     <>
        <RadioGroup.Item ... />
        <label>...</label>
     </>
    )
</RadioGroup.Root>
```

This way we don't need to 
1. Encode 2 informations inside a string inside `label`
2. introduce renderProps `renderItem`

it's much more composable.

muhajirdev: We can use `RadioGroup` for simpler use case.

and use the `Root` `Item` and `Indicator` for a complex use case that needs a lower level API

ahmedriad1: Salam bro @muhajirdev, yeah makes sense sorry I must've wrote it before sleep 😄

muhajirdev: no worries, may Allah reward you for the contributions 😁

PR #1178, opened 2022-02-01
fix: can't select multiple translations

muhajirdev:


src/components/Navbar/Drawer/index.tsx
```diff
@@ -108,13 +108,13 @@ const Drawer: React.FC<Props> = ({
...
 
   // Hide navbar after successful navigation
-  useEffect(() => {
-    router.events.on('routeChangeComplete', () => {
-      if (isOpen) {
-        closeDrawer('navigation');
-      }
-    });
-  }, [closeDrawer, router.events, isOpen]);
+  // useEffect(() => {
```
osamasayed, line 111: This is an okay approach for the translations issue but it is not working as before for other drawers like navigation or search drawers when navigating away, the drawers stay opened which I think is not what we want. We can selectively choose to ignore this for the settings drawer only.

osamasayed (approved):

PR #1196, opened 2022-02-03
fix: firefox direction alignment issue

muhajirdev:
I am still not 100% sure why this thing happend and why removing this code fixed the issue. Need some input from @osamasayed 

## How to test
- open normal next.quran.com with firefox
- compare it with the preview in this PR

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -18,20 +18,6 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 
-  // .ltr,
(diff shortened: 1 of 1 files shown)


osamasayed:
It seems this issue was introduced by [this bit](https://github.com/quran/quran.com-frontend-next/pull/1131/files?diff=split&w=0#diff-2c0189636ff225dd87da97bcea9b585e70a25a5f601b816456f20bf146bec8ad) of Naveed's PR. I think it's safe to revert.

osamasayed (approved):

PR #1249, opened 2022-02-15
fix: debounce search query on main search bar

### LINKED ISSUE #1248: [bug]: main search queries not being debounced

### Describe the bug

When searching for a particular Juz, Surah etc from the main search bar it makes a network call on every character entered.

### Relevant log output

_No response_

### Steps to reproduce

1. Go to beta.quran.com
2. Start typing on the main search bar
3. Have network tab opened and see the queries for each search term character added

OS: any
Browser: any

### Environment

Production(beta.quran.com)

yj7o5:
### Summary
This PR addresses the bug #1248 where search queries aren't being de-bounced, resulting in N number of search queries. 

### Test Plan
Visit beta.quran.com to search for any juz or surah from the main search bar while having the network tab opened. 

### Screenshots

| Before | After |
| ------ | ------ |
| before | after |

closes #1248

### DIFF
--- a/src/components/CommandBar/CommandBarBody/index.tsx
+++ b/src/components/CommandBar/CommandBarBody/index.tsx
@@ -7,6 +7,7 @@ import useTranslation from 'next-translate/useTranslation';
 import { shallowEqual, useSelector } from 'react-redux';
 
 import IconSearch from '../../../../public/icons/search.svg';
+import useDebounce from '../../../hooks/useDebounce';
 import CommandsList, { Command } from '../CommandsList';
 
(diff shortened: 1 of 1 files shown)


yj7o5:
@osamasayed whenever you get chance this PR as well :), JazakAllah.

osamasayed:
> @osamasayed whenever you get chance this PR as well :), JazakAllah.

JazakAllah Kahiran for your contribution. Your solution makes perfect sense from a technical perspective but we are intentionally not debouncing it because we want the results to return as fast as possible to the user. Will close this PR. Thank you again for your contribution!

PR #1268, opened 2022-02-18
highlight current verse juz and page in sidebar navigation

muhajirdev:
* Current `page`, `juz`, `verse` and `surah` is now highlighted 
* when opening the page. it should scroll to the current `page/juz/verse/surah`


src/hooks/useScrollTo.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useRef } from 'react';
+
+const DEFAULT_SCROLL_MARGIN = 300;
+
+const useScrollTo = (
```
muhajirdev, line 5: we're using `scrollTo`, because`.scrollIntoView` is not reliable enough for this use case.

src/hooks/useScrollTo.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Scroll to a specific element
+ * How to use: pass the returned `selectedItemRef` to the element you want to scroll to
+ *
+ * @param {Function} scrollAbleParentMapper - A function that returns the scrollable parent of the element.
+ * (element that has overflow: auto/scroll)
+ * @param {number} scrollMargin - The margin to be added to the scrollable parent.
+ * @returns {MutableRefObject} selectedItemRef - React ref
+ */
+const useScrollTo = (
+  scrollAbleParentMapper = (node) => node.parentNode,
```
muhajirdev, line 16: scrollable

src/components/QuranReader/SidebarNavigation/PageSelection.tsx
```diff
@@ -19,6 +22,7 @@ const PageSelection = () => {
       searchPlaceholder={t('sidebar.search-page')}
       renderItem={(page) => `${t('page')} ${page.label}`}
       isJuz={false}
+      selectedItem={selectedPage}
```
osamasayed, line 25: nit: no need for `selectedPage` var since it's only used once. We can set `selectedItem={Number(lastReadPage)}` directly

src/components/QuranReader/SidebarNavigation/SidebarNavigation.module.scss
```diff
@@ -44,23 +47,38 @@
 
 .contentContainer {
   margin-block-start: var(--spacing-medium);
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+  padding-block-end: 1rem;
```
osamasayed, line 53: this should be `--spacing-medium` which is `1rem`

src/components/QuranReader/SidebarNavigation/ScrollableSelection.tsx
```diff
@@ -21,20 +31,37 @@ const ScrollableSelection = ({ items, searchPlaceholder, renderItem, getHref, is
     }
   }, [searchQuery, filteredItems, isJuz]);
 
+  const [scroll, selectedItemRef] = useScrollToElement<HTMLDivElement>({
```
osamasayed, line 34: The scrolling options should be taken out of the entire component and stored as a const since its reference will change on every re-render also since the 3 components are using the same options, I suggest we export a common const that the 3 components will use. (`ComboboxItems` seems to also use the same exact config so the 4 components :) ). Maybe you can add it inside `useScrollToElement` itself something like:

```
export const SCROLL_TO_NEAREST_BLOCK_OPTIONS = {
  block: 'nearest', // 'block' relates to vertical alignment. see: https://stackoverflow.com/a/48635751/1931451 for nearest.
} as ScrollIntoViewOptions;
```

src/components/QuranReader/SidebarNavigation/ScrollableSelection.tsx
```diff
@@ -21,20 +31,37 @@ const ScrollableSelection = ({ items, searchPlaceholder, renderItem, getHref, is
...
+    block: 'nearest',
+  });
+  useEffect(() => {
+    scroll();
+  }, [selectedItemRef, selectedItem, scroll]);
```
osamasayed, line 39: is there a reason why we are adding `selectedItemRef` as a dependency?

src/components/QuranReader/SidebarNavigation/SurahList.tsx
```diff
@@ -50,6 +51,13 @@ const SurahList = () => {
   const filteredChapters = searchQuery
     ? filterSurah(chapterDataArray, searchQuery)
     : chapterDataArray;
+
+  const [scrollTo, selectedChapterRef] = useScrollToElement<HTMLDivElement>({ block: 'nearest' });
+
+  useEffect(() => {
+    scrollTo();
+  }, [selectedChapterRef, currentChapterId, scrollTo]);
```
osamasayed, line 59: Ditto: is there a reason why we are using `selectedChapterRef` as a dependency?

src/components/QuranReader/SidebarNavigation/VerseList.tsx
```diff
@@ -37,6 +42,12 @@ const VerseList = () => {
     }
   }, [searchQuery, filteredVerseKeys]);
 
+  const [scrollTo, selectedVerseRef] = useScrollToElement<HTMLDivElement>({ block: 'nearest' });
+
+  useEffect(() => {
+    scrollTo();
+  }, [scrollTo, selectedVerseRef, lastReadVerseKey.verseKey]);
```
osamasayed, line 49: Ditto: is there a reason why we are using `selectedVerseRef` as a dependency?

osamasayed (approved):

PR #1270, opened 2022-02-18
Sidebar navigation: UX when pressing `Enter`

muhajirdev:
In Sidebar Navigation try
* verse navigation
* surah navigation
* page navigation
* juz navigation

How to test
* Open surah sidebar navigation, write a value in the inputbox. For example write `Verse` `33` and then press enter. 
* it should navigate you to verse 33 in the current chapter

### DIFF
--- a/src/components/QuranReader/SidebarNavigation/VerseList.tsx
+++ b/src/components/QuranReader/SidebarNavigation/VerseList.tsx
@@ -1,6 +1,7 @@
 import { useState, useMemo, useEffect } from 'react';
 
 import useTranslation from 'next-translate/useTranslation';
+import { useRouter } from 'next/router';
 
 import styles from './SidebarNavigation.module.scss';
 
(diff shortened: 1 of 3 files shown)


src/components/QuranReader/SidebarNavigation/VerseList.tsx
```diff
@@ -37,14 +39,26 @@ const VerseList = () => {
     }
   }, [searchQuery, filteredVerseKeys]);
 
+  // Handle when user press `Enter` in input box
+  const handleVerseInputSubmit = (e) => {
+    e.preventDefault();
+    const selectedVerseKey = filteredVerseKeys.find((verseKey) => verseKey.endsWith(searchQuery));
```
osamasayed, line 45: a qq, why cannot we use `filteredVerseKeys[0]` here instead of having to search again? this will have 2 advantages:

1. less code duplication
2. consistency with the way we are using `SurahList`.
3. Since `filteredVerseKeys` handles localized numbers, searching for localized verse key then pressing enter will work.

osamasayed (approved):

PR #1280, opened 2022-02-19
fix: word actions color on sepia mode

muhajirdev:
Before
image 

After
image

Before
image

After
image


src/components/QuranReader/ReadingView/WordPopover/WordPopover.module.scss
```diff
@@ -9,20 +9,9 @@ $white: #fff;
   padding-block-end: var(--spacing-micro);
```
osamasayed, line 9: We should remove `@use "src/styles/theme";` since we are no longer using it.

osamasayed (approved):

PR #1292, opened 2022-02-21
Change Tafsir's font size

osamasayed:
### Summary
This PR adds the ability to change the font size of the Tafsir content.

### Screenshots

| Screenshots |
| ------ |
|Screen Shot 2022-02-21 at 15 06 18|

### DIFF
--- a/src/components/QuranReader/TafsirView/TafsirText/index.tsx
+++ b/src/components/QuranReader/TafsirView/TafsirText/index.tsx
@@ -15,7 +15,7 @@ import {
 } from 'src/redux/slices/QuranReader/styles';
 import { logValueChange } from 'src/utils/eventLogger';
 
-type Props = {
+type TafsirTextProps = {
   direction: string;
   languageCode: string;
   text: string;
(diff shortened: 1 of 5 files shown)


src/components/QuranReader/TafsirView/TafsirText/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+import {
+  MAXIMUM_FONT_STEP,
+  MINIMUM_FONT_STEP,
+  selectQuranReaderStyles,
+  increaseTafsirFontScale,
+  decreaseTafsirFontScale,
+} from 'src/redux/slices/QuranReader/styles';
+import { logValueChange } from 'src/utils/eventLogger';
+
+type Props = {
```
muhajirdev, line 18: nit: `TafsirTextProps`?
I actually prefer `Props` lol, but other components seems to have component name prefix

muhajirdev (approved):

src/components/QuranReader/TafsirView/TafsirText/index.tsx
osamasayed: Well the components I create always have `Props` 😆 but we should standardize this. Will change to `TafsirTextProps` for now

PR #1300, opened 2022-02-22
fix: janky width changes when user is scrolling (reading view)

muhajirdev:


src/styles/_utility.scss
```diff
@@ -353,6 +354,57 @@ $skeleton-height-map: (
...
+  qpc_uthmani_hafs: (
+    tablet: (
+      1: 55vh,
+      2: 59vh,
+      3: 63.8vh,
+      4: 65vh,
+      5: 66vh,
+    ),
+  ),
+  text_indopak: (
```
osamasayed, line 379: I think we should separate this into 15 and 16 because the scales will not be 100% identical

src/components/QuranReader/ReadingView/Line.tsx
```diff
@@ -23,9 +25,17 @@ export type LineProps = {
   quranReaderStyles: QuranReaderStyles;
   pageIndex: number;
   lineIndex: number;
+  isFixedWidth?: boolean;
```
osamasayed, line 28: `Line` is only used by the ReadingView I think we don't need this prop and always apply the styling `isFixedWidth`

src/components/QuranReader/ReadingView/Line.tsx
```diff
@@ -40,13 +50,20 @@ const Line = ({ lineKey, words, isBigTextLayout, pageIndex, lineIndex }: LinePro
   const firstWordData = getWordDataByLocation(words[0].location);
   const shouldShowChapterHeader = firstWordData[1] === '1' && firstWordData[2] === '1';
 
+  const { quranFont, quranTextFontScale } = useSelector(
+    selectQuranReaderStyles,
+    shallowEqual,
+  ) as QuranReaderStyles;
```
osamasayed, line 56: No need for `as QuranReaderStyles` since the type is auto-resolved.

src/components/QuranReader/ReadingView/Line.tsx
```diff
@@ -40,13 +50,20 @@ const Line = ({ lineKey, words, isBigTextLayout, pageIndex, lineIndex }: LinePro
...
+
   return (
     <div
       ref={selectedItemRef}
       id={lineKey}
       className={classNames(styles.container, {
         [styles.highlighted]: isHighlighted,
         [styles.mobileInline]: isBigTextLayout,
+        [styles[getLineWidthClassName(quranFont, quranTextFontScale)]]: isFixedWidth,
```
osamasayed, line 65: as mentioned above, this should always be applied so we can add it as a param inside classNames e.g.`classNames(styles.container, styles[getLineWidthClassName(quranFont, quranTextFontScale)]`

src/components/QuranReader/ReadingView/Line.tsx
```diff
@@ -40,13 +50,20 @@ const Line = ({ lineKey, words, isBigTextLayout, pageIndex, lineIndex }: LinePro
...
+        [getLineWidthClassName(quranFont, quranTextFontScale)]: true,
```
osamasayed, line 66: 2 things: 

1. any reason why we are having this without `styles`? and if we do, it's a duplicate of the previous line
2. if it's always `true`, we should make it as a param to `classNames(` instead

src/components/QuranReader/ReadingView/Page.tsx
```diff
@@ -48,6 +48,7 @@ const Page = ({ verses, pageNumber, quranReaderStyles, pageIndex }: PageProps) =
           key={key}
           isBigTextLayout={isBigTextLayout}
           quranReaderStyles={quranReaderStyles}
+          isFixedWidth
```
osamasayed, line 51: ditto, this should be removed

src/utils/fontFaceHelper.ts
```diff
@@ -80,3 +80,13 @@ export const getFontClassName = (
...
 };
+
+/**
+ * Dynamically generate the className of the combination between the font
+ * name + size that will match the output of
+ * generate-font-scales function inside {@see _utility.scss}.
+ */
+
+export const getLineWidthClassName = (quranFont, fontScale) => {
+  return `${quranFont}-line-width-${fontScale}`;
```
osamasayed, line 91: as mentioned above, we should separate 16 lines from 15 lines and add the condition here similar to `getFontClassName` function

src/styles/_utility.scss
```diff
@@ -105,16 +105,17 @@ $line-heights-map: (
...
     (
       tablet: (
         1: 3.2vh,
         2: 3.5vh,
-        3: 4vh,
-        4: 4vh,
-        5: 4.4vh,
+        3: 3.6vh,
+        4: 3.7vh,
+        5: 3.8vh,
```
osamasayed, line 118: These changes should reflect also in `$line-heights-map` and `$skeleton-height-map`:
1. `$line-heights-map` to make sure the entire viewport is filled for 1 page for the reading view.
2.   `$skeleton-height-map` to avoid having a mismatch between the page height when it's showing the skeleton and when it's actually showing the Quranic text.

src/styles/_utility.scss
```diff
@@ -178,7 +179,7 @@ $scales-map: (
       2: 3.8vh,
       3: 4vh,
       4: 4.3vh,
-      5: 4.5vh,
+      5: 4.4vh,
```
osamasayed, line 182: Ditto, this should be reflected in `$line-heights-map` and `$skeleton-height-map`.

src/components/QuranReader/ReadingView/ReadingViewSkeleton.module.scss
```diff
@@ -8,9 +8,8 @@
 }
 
 .skeleton {
-  width: 100%;
-  max-width: calc(15 * var(--font-size));
-  min-width: constants.$reading-view-page-min-width;
+  width: var(--line-width);
```
osamasayed, line 11: should it not be `var(--line-width, auto);`?

src/components/QuranReader/ReadingView/index.tsx
```diff
@@ -188,6 +197,9 @@ const ReadingView = ({
             }}
           />
         )}
+        <div className={styles.virtuosoScroller}>
```
osamasayed, line 200: I think this was forgotten here :)

muhajirdev:
I noticed a few things
- it seems that the fallback font is not loaded on Indopak (But I actually love this experience, less vertical layout shift, and less movement)
- there's still some width changes in code_v2, (skeleton -> fallback_font -> code_v2). At least on my screen

This is already huge improvements 🚀, I think we should ship it. 

One improvement we can do in the future is to have fixed height / line-height, across fonts and skeleton. This way we won't have vertical layout shift.

osamasayed (approved):

PR #1308, opened 2022-02-26
hide repetitive translation author information

muhajirdev:
Based on the feedback from our designer (yusuf). We want to hide repetitive information (translation author and book name)

Proposal: When there's only 1 translation selected, hide the author name & book name. 
Because for people who only use default translation (1 translation), he's probably on casual reading mode, (not deep reading mode).
So resource name, might be less relevant. 

## Before
image

## After
image

### DIFF
--- a/src/components/QuranReader/TranslationView/TranslationText/index.tsx
+++ b/src/components/QuranReader/TranslationView/TranslationText/index.tsx
@@ -19,7 +19,7 @@ import Footnote from 'types/Footnote';
 interface Props {
   translationFontScale: number;
   text: string;
-  resourceName: string;
+  resourceName?: string;
   languageId: number;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/QuranReader/TranslationView/TranslationText/index.tsx
```diff
@@ -19,7 +19,7 @@ import Footnote from 'types/Footnote';
 interface Props {
   translationFontScale: number;
   text: string;
-  resourceName: string;
+  resourceName?: string;
```
osamasayed, line 22: IMO, this should be left as it was and we introduce a new `boolean` called `showResourceName` or something similar.

muhajirdev: I have a few reason to keep the API this way. In my opinion we should keep the API Surface Area minimal and a component should only receive the data it needs.

an implementation in a `Card` component that only need to show `title`, looks like this

```
<Card title="card title" />
```

vs

```
<Card title="card title" description={null} coverImage={null} shouldShowDescription={false} shouldShowCoverImage={false} />
```

the first example is better IMO

osamasayed: Yes I agree. Thanks Muhajir!

src/components/chapters/ChapterHeader/index.tsx
```diff
@@ -42,6 +46,11 @@ const ChapterHeader: React.FC<Props> = ({ chapterId, pageNumber, hizbNumber }) =
   const { translatedName } = chapterData;
   const { transliteratedName } = chapterData;
 
+  const onChangeTranslationClicked = () => {
+    dispatch(setIsSettingsDrawerOpen(true));
+    dispatch(setSettingsView(SettingsView.Translation));
```
osamasayed, line 51: Can we add logging here?

osamasayed (approved):

PR #1328, opened 2022-03-01
Make page number relative to vh

osamasayed:
### Summary
This PR makes the font size of the page number in the Reading View relative to the viewport's height to match that of the Quranic text. 

### Screenshots (large screens)

| Before | After |
| ------ | ------ |
|Screen Shot 2022-03-01 at 19 29 35|Screen Shot 2022-03-01 at 19 29 53|

### DIFF
--- a/src/components/QuranReader/ReadingView/PageFooter.module.scss
+++ b/src/components/QuranReader/ReadingView/PageFooter.module.scss
@@ -1,5 +1,3 @@
-@use "src/styles/breakpoints";
-
 .pageText {
   text-align: center;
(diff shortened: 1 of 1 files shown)


src/components/QuranReader/ReadingView/PageFooter.module.scss
```diff
@@ -1,17 +1,14 @@
+@use "src/styles/breakpoints";
```
muhajirdev, line 1: I think you forget to remove this? :D

muhajirdev (approved):

PR #1331, opened 2022-03-02
fix: remove startVerse query param when switching mode

muhajirdev:
to solve the problem
- when user have startingVerse=68 param, 
- he scroll up
- and then switch from translation -> reading mode. It scroll to verse 68 on reading mode
- removing this param, fixed the behavior -> doesn't scroll anymore

### DIFF
--- a/src/components/QuranReader/ReadingPreferenceSwitcher.tsx
+++ b/src/components/QuranReader/ReadingPreferenceSwitcher.tsx
@@ -1,6 +1,5 @@
 import { useMemo } from 'react';
 
-import omit from 'lodash/omit';
 import useTranslation from 'next-translate/useTranslation';
 import { useRouter } from 'next/router';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


src/components/QuranReader/ReadingPreferenceSwitcher.tsx
```diff
@@ -30,15 +33,28 @@ const ReadingPreferenceSwitcher = () => {
     ],
     [t],
   );
+
+  const onViewSwitched = (view: ReadingPreference) => {
+    logValueChange('reading_preference', readingPreference, view);
+
+    // drop `startingVerse` from query params
+    const newQueryParams = omit(router.query, ['startingVerse']);
```
osamasayed, line 41: I think we don't need to use `omit` at all. You can check a reference for what you are trying to do inside `src/components/Navbar/SettingsDrawer/ResetButton.tsx`.`onResetSettingsClicked`

muhajirdev: got it 👍

PR #1340, opened 2022-03-03
fix: settings drawer hidden when audio auto scroll

muhajirdev:
How to test
* go to next.quran.com
* go to al baqarah
* play audio
* open settings drawer
* wait until the page auto scroll
* the "close" icon will be hidden.

Now try on this PR preview

### DIFF
--- a/src/components/Navbar/Drawer/index.tsx
+++ b/src/components/Navbar/Drawer/index.tsx
@@ -83,7 +83,7 @@ const Drawer: React.FC<Props> = ({
   hideCloseButton = false,
   closeOnNavigation = true,
 }) => {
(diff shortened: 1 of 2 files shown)


src/components/Navbar/Drawer/index.tsx
```diff
@@ -83,6 +83,7 @@ const Drawer: React.FC<Props> = ({
   hideCloseButton = false,
   closeOnNavigation = true,
 }) => {
+  const { isVisible } = useSelector(selectNavbar, shallowEqual);
```
tifa2UP, line 86: `const isNavbarVisible = useSelector(selectNavbar, shallowEqual).isVisible;`

If you use isVisible variable it might be confused for the drawer being visible. Feel free to merge after :)

PR #1355, opened 2022-03-04
Fix text size on devices with screen larger than phones

RagibHasin:
### Summary

Arabic text size in screen sizes over and including tablet is very small even at size 5. My investigation shows that it is caused by writing `vh` instead of `vw` erroneously. This pull request attempts to fix that, bi-iznillah.

### Test Plan

Don't know if applicable.

### Screenshots

The screenshots are showing the state of the reading view on a 1080p screen on the latest Firefox on Windows 11.

| Before | After |
| ------ | ------ |
| before | after |


RagibHasin:
Assalamu 'Alaykum.

It seems that #1358 by @osamasayed supersedes this PR. So, I think this PR should be closed.

PR #1375, opened 2022-03-08
Improve icons DX

ahmedriad1:
### Summary

Currently importing or using icons is a pain:
- Importing from the public folder results in an ugly import, ie: `../../../../../../public/icons/search.svg`
- No TS support
- Inconsistent naming all over app

Solution:
I created an icons folder, created a base TS interface, and made sure names are consistent all over the app. Also, I fixed all icon imports in the app

### Screenshots


ahmedriad1:
I don't think we need this anymore. Closing this PR in favor of https://github.com/quran/quran.com-frontend-next/pull/1781

PR #1396, opened 2022-03-11
fix: translation param not working

muhajirdev:
on Incognito mode

* try opening this page `https://quran.com/surah-al-kahf?translations=206%2C20`, 
* ideally it should show saheeh international & Ruwwad Center. 
* But in this case it shows "Dr. Mustafa Khattab, The Clear Quran"

This PR fixed the issue

### DIFF
--- a/src/components/QuranReader/TranslationView/TranslationPage/index.tsx
+++ b/src/components/QuranReader/TranslationView/TranslationPage/index.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect } from 'react';
 
 import useTranslation from 'next-translate/useTranslation';
+import { useRouter } from 'next/router';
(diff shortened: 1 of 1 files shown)


src/components/QuranReader/TranslationView/TranslationPage/index.tsx
```diff
@@ -45,6 +48,12 @@ const TranslationPage: React.FC<Props> = ({
   setApiPageToVersesMap,
 }) => {
   const { lang, t } = useTranslation('common');
+  const router = useRouter();
+  const defaultTranslations = getTranslationsInitialState(lang).selectedTranslations;
+  const translationParams = (router.query.translations as string)
```
osamasayed, line 53: We should memoize this

PR #1415, opened 2022-03-14
Remove url handler from ContentModal

muhajirdev:
When opening contentModal, the `fakeNavigate` function is called multiple times. Because the function is tied to react lifecycle via useEffect. This adds unnecessary complexity. 

`fakeNavigate` function should be triggered inside `onClick`, not on `render`.

Removing this handler from contentModal to avoid other complexity in the future

---

Here's a few problem that's already found 

It's called multiple times (more than what we need) when opening tafsir,
image

When it's called often enough, it breaks safari.
image

### DIFF
--- a/src/components/QuranReader/TranslationView/QuranReflectButton.tsx
+++ b/src/components/QuranReader/TranslationView/QuranReflectButton.tsx
@@ -3,6 +3,7 @@ import { useRef, useState } from 'react';
 import classNames from 'classnames';
 import useTranslation from 'next-translate/useTranslation';
 import dynamic from 'next/dynamic';
+import { useRouter } from 'next/router';
 
 import ChatIcon from '../../../../public/icons/chat.svg';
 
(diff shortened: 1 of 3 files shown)


src/components/QuranReader/TranslationView/QuranReflectButton.tsx
```diff
@@ -12,8 +13,9 @@ import Button, { ButtonShape, ButtonSize, ButtonVariant } from 'src/components/d
 import ContentModal from 'src/components/dls/ContentModal/ContentModal';
 import { logButtonClick } from 'src/utils/eventLogger';
 import {
+  fakeNavigate,
   getQuranReflectVerseUrl,
-  getVerseSelectedReflectionNavigationUrl,
+  // getVerseSelectedReflectionNavigationUrl,
```
osamasayed, line 18: nit: remove un-used code instead of commenting

osamasayed (approved):

PR #1432, opened 2022-03-15
Add Tarteel deep linking

tifa2UP:
TSIA

### DIFF
--- a/public/.well-known/apple-app-site-association
+++ b/public/.well-known/apple-app-site-association
@@ -1,0 +1,11 @@
+{
+    "applinks": {
+        "apps": [],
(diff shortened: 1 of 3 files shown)


public/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,11 @@
+{
+    "applinks": {
+        "apps": [],
+        "details": [
+            {
+                "appID": "GLXDQKH95X.com.iqraapp.Iqra",
+                "paths": ["/*"]
```
mmmoussa, line 7: Change this to `"*"`

PR #1470, opened 2022-03-21
Quran Reflect share functionality

muhajirdev:
image


muhajirdev:
There's a bug in this PR, when we click button "share" and then click the "three dots". User will be stuck.

I tried to update the radix and it solve the problem. But I will put in another PR

osamasayed (approved):

PR #1498, opened 2022-03-24
[PWA] Add apple-touch-icon

osamasayed:
| Screenshots |
| ------ |
|Screen Shot 2022-03-24 at 12 56 40|


tifa2UP (approved):

src/pages/_app.tsx
```diff
@@ -50,6 +50,7 @@ function MyApp({ Component, pageProps }): JSX.Element {
   return (
     <>
       <Head>
+        <link rel="apple-touch-icon" sizes="192x192" href="/images/logo/Logo@192x192.png" />
```
tifa2UP, line 53: this probably shouldn't be here, we have somewhere else where we specify all the meta tags and links

PR #1501, opened 2022-03-24
QR: use html

muhajirdev:


package.json
```diff
@@ -50,6 +50,7 @@
     "clipboard-copy": "^4.0.1",
     "firebase": "^9.6.9",
     "fuse.js": "^6.5.3",
+    "html-truncate": "^1.2.2",
```
muhajirdev, line 53: This package doesn't have external dependency (https://github.com/huang47/nodejs-html-truncate/blob/master/package.json)

https://bundlephobia.com/package/html-truncate
image

osamasayed (approved):

PR #1565, opened 2022-04-06
Update footer language selector position

osamasayed:
### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2022-04-06 at 7 36 50 PM|Screen Shot 2022-04-06 at 7 36 17 PM|


src/components/dls/PopoverMenu/PopoverMenu.tsx
```diff
@@ -23,6 +31,7 @@ const PopoverMenu = ({
   isPortalled = true,
   isModal = true,
   onOpenChange,
+  side = PopoverMenuSide.BOTTOM,
```
tifa2UP, line 34: should we call this `expandDirection` instead? side isn't very clean

tifa2UP (approved):

PR #1621, opened 2022-04-30
fix #1618

### LINKED ISSUE #1618: [bug]: Getting play undefined instead on Popup in production site

### Describe the bug

When some surah is playing and then I play next then get the popup of 

**Click on "Start Over" if you would like to play undefined instead**

The issue is getting play undefined,  expected output was to show the surah name

image

### Relevant log output

```shell
no logs
```

### Steps to reproduce

1. Play any surah
2. select on any other
3. will get the pupup

Mac OS
Chrome

### Environment

Production(quran.com)

mo-othman:
### Summary
This is a fix for https://github.com/quran/quran.com-frontend-next/issues/1618
The audio player was missing the chapter data provider 
I have placed it in the app file since the audio player is a component not a page 
let me know if it should go to a diff place 

### Screenshots

| Before | After |
| ------ | ------ |
| Screen Shot 2022-04-30 at 13 21 28 |  Screen Shot 2022-04-30 at 13 19 50 |


osamasayed:
JazakAllah Khairan @mo-othman, this issue will be fixed in #1741

PR #1632, opened 2022-05-07
Fix #1527

### LINKED ISSUE #1527: [bug]: when sharing a translated verse on facebook, the arabic text is used

### Describe the bug

when sharing a link like https://quran.com/49/11?translations=22

it is expected that the translation is given priority over the arabic. instead it is the arabic that is seen in the preview of the link

### Relevant log output

_No response_

### Steps to reproduce

share a translation link

facebook.com

### Environment

Staging(next.quran.com)

craxrev:
### Summary
This is a partial fix for the open graph description tag not taking in consideration the translation parameter.

The reason I said this is a partial fix, because I'm actually using the verse response to get the translation: `versesResponse.verses[0].translations?.[0].text`, and this response is not reflecting changes made in the UI or even if we change the translation param in the url: `?translation=x`

Thus the better solution is to reflect the translation changes in the UI or URL to the api call, which I did not figure out yet

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |

### DIFF
--- a/src/pages/[chapterId]/[verseId]/index.tsx
+++ b/src/pages/[chapterId]/[verseId]/index.tsx
@@ -47,7 +47,7 @@ const Verse: NextPage<VerseProps> = ({
 }) => {
   const { t, lang } = useTranslation('common');
   const {
-    query: { verseId },
+    query: { verseId, translations },
   } = useRouter();
   if (hasError || !versesResponse.verses.length) {
     return <Error statusCode={500} />;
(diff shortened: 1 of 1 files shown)


osamasayed:
Salam @craxrev we currently do [ISR](https://vercel.com/docs/concepts/next.js/incremental-static-regeneration) for each path e.g. `/49/11` and when doing ISR, the query params cannot be used until the client side code loads so setting the `description` tag in this case will only reflect the data that was server side generated which will have different the default translation content rather than the on in the url. JazakAllah Khairan for your effor.

craxrev:
Salam, that's fine, I'm glad I put effort into that, now I'll try to understand how ISR is working there 👍

PR #1635, opened 2022-05-08
open donation popup every 10 sessions

muhajirdev:


src/components/OneTimePopup/OneTimePopup.tsx
```diff
@@ -10,14 +10,18 @@ import Modal from '../dls/Modal/Modal';
...
 
+const OPEN_POPUP_EVERY_N_SESSIONS = 10;
 const OneTimePopup = () => {
   const { t } = useTranslation('common');
-  const isPopupVisible = useSelector(selectIsPopupVisible);
+  const sessionCount = useSelector(selectSessionCount);
   const [isDonateButtonLoading, setIsDonateButtonLoading] = useState(false);
-  const dispatch = useDispatch();
+
+  const initialIsPopupVisible =
```
osamasayed, line 22: nit: no need to have a var called `initialIsPopupVisible` we can simply calculate the value directly:

```
useState(()=> {
return sessionCount % OPEN_POPUP_EVERY_N_SESSIONS === 0 && sessionCount > 0
}}
```

src/redux/slices/popup.ts
```diff
@@ -1,28 +1,28 @@
...
+    incrementSessionCount: (state: PopupState) => ({
       ...state,
-      isPopupVisible: action.payload,
+      sessionCount: state.sessionCount + 1,
     }),
   },
 });
 
-export const { setIsPopupVisible } = popupSlice.actions;
+export const { incrementSessionCount } = popupSlice.actions;
```
osamasayed, line 24: We should add migration and set `sessionCount` to 0.

osamasayed (approved):

src/components/OneTimePopup/OneTimePopup.tsx
```diff
@@ -10,14 +10,18 @@ import Modal from '../dls/Modal/Modal';
 
 import styles from './OneTimePopup.module.scss';
 
-import { selectIsPopupVisible, setIsPopupVisible } from 'src/redux/slices/popup';
+import { selectSessionCount } from 'src/redux/slices/popup';
 import openGivingLoopPopup from 'src/utils/givingloop';
 
+const OPEN_POPUP_EVERY_N_SESSIONS = 10;
 const OneTimePopup = () => {
```
tifa2UP, line 17: DonatePopup

tifa2UP (changes requested):

src/components/SessionIncrementor.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import { useDispatch } from 'react-redux';
+
+import { incrementSessionCount } from 'src/redux/slices/session';
+
+const SessionIncrementor = () => {
+  const dispatch = useDispatch();
+  useEffect(() => {
+    dispatch(incrementSessionCount());
+  }, [dispatch]);
```
muhajirdev, line 11: 1. This part unfortunately, depends heavily on hooks `useDispatch` and `useEffect`. That's basically all it does.

So there's almost 0 logic we can extract into a util function

2. We cannot put this in _app.tsx. Since this hook need to be under `ReduxProvider`. (Cannot be called outside `ReduxProvider`). So it need to be inside another component.

- Similar pattern is also implemented in `GlobalScrollListener` (which actually can be just a hook inside `GlobalListener` ), and `AudioRepeatManager`. Which literally renders nothing. 
- I am not a big fan of this pattern, but I am following this for consistency

src/redux/slices/session.ts
```diff
@@ -0,0 +1,28 @@
...
+      count: state.count + 1,
+    }),
+  },
+});
+
+export const { incrementSessionCount } = popupSlice.actions;
+
+export const selectSessionCount = (state: RootState) => state.session.count;
+
+export default popupSlice.reducer;
```
tifa2UP, line 28: sessionSlice

tifa2UP (approved):

PR #1658, opened 2022-05-27
feat: make visible sidebar navigation by default

muhajirdev:


src/components/QuranReader/ContextMenu.tsx
```diff
@@ -78,13 +79,20 @@ const ContextMenu = () => {
                   `sidebar_navigation_${isSidebarNavigationVisible ? 'close' : 'open'}_trigger`,
                 );
                 e.stopPropagation();
-                dispatch(setIsVisible(!isSidebarNavigationVisible));
+                if (isSidebarNavigationVisible === 'auto') {
+                  // eslint-disable-next-line no-unneeded-ternary
+                  const shouldBeVisible = isMobile() ? true : false;
```
osamasayed, line 84: nit: `shouldBeVisible = isMobile() || false;` and no need for `// eslint-disable-next-line no-unneeded-ternary`

src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
```diff
@@ -104,7 +105,7 @@ const SidebarNavigation = () => {
 };
 
 const TABLET_WIDTH = 768;
-const isMobile = () => {
+export const isMobile = () => {
```
osamasayed, line 108: I think we should take this out to a util since the logic is not related to `SidebarNavigation`.

osamasayed (approved):

PR #1662, opened 2022-06-06
implement reading session integration with BE

muhajirdev:


src/components/QuranReader/index.tsx
```diff
@@ -30,6 +33,8 @@ type QuranReaderProps = {
   quranReaderDataType?: QuranReaderDataType;
 };
 
+const DEBOUNCE_WAIT_TIME = 2000; // 2 seconds
```
osamasayed, line 36: nit: this should be renamed to `WAITING_TIME_DEBOUNCE_WAIT_TIME` since it's in the QuranReader

src/utils/auth/api.ts
```diff
@@ -56,3 +61,11 @@ export const getIsResourceBookmarked = async (
...
 ): Promise<boolean> =>
   privateFetcher(makeIsResourceBookmarkedUrl(mushafId, key, type, verseNumber));
+
+export const addReadingSession = async (chapterNumber: number, verseNumber: number) =>
+  postRequest(makeReadingSessionsUrl(), {
+    chapterNumber,
+    verseNumber,
+  });
+
+export const isLoggedIn = () => !!Cookies.get(USER_ID);
```
osamasayed, line 71: I don't think this belongs here

PR #1667, opened 2022-06-09
React hook form validation

muhajirdev:
Component is loaded seperately (lazy loaded)
image

Form with error validation message
image

Form not submitted yet
image


types/FormField.ts
```diff
@@ -0,0 +1,30 @@
...
+  type: RuleType;
+  value: RuleValue;
+  errorId?: ErrorMessageId;
+};
+
+type FormField = {
+  field: string;
+  rules?: FieldRule[];
+  type: FormFieldType;
+};
```
osamasayed, line 28: nit: ditto, we should split this so it would easier to maintain

src/utils/validation.ts
```diff
@@ -0,0 +1,60 @@
...
+        return false;
+      },
+    };
+  }
+
+  return rules;
+};
+
+const buildTranslatedErrorMessage = (rule: FieldRule, t: Translate, fieldName: string) => {
+  return t(`validation.${rule.errorId}`, { field: capitalize(fieldName) });
```
osamasayed, line 59: What happens if the error returned from BE is new and doesn't exist on our side? I think we should return a general error

muhajirdev: I agree

osamasayed (approved):

PR #1668, opened 2022-06-10
prevent execessive data fetching & add typing

muhajirdev:


types/ReadingSession.ts
```diff
@@ -0,0 +1,9 @@
+interface ReadingSession {
+  chapterNumber: number;
+  id: number;
+  updatedAt: string;
+  userId: string;
+  verseNumber: 1;
```
osamasayed, line 6: `verseNumber: number`. Also I believe `userId` should not be sent from BE.

osamasayed (approved):

PR #1673, opened 2022-06-12
Persist audio settings

osamasayed:


muhajirdev:
tested, all values seems to be persisted to DB

muhajirdev (approved):

PR #1681, opened 2022-06-15
Email verification code

muhajirdev:


src/utils/auth/api.ts
```diff
@@ -29,7 +29,7 @@ export const privateFetcher = async <T>(input: RequestInfo, init?: RequestInit):
  * @param {RequestData} requestData
  * @returns {Promise<T>}
  */
-const postRequest = <T>(url: string, requestData: RequestData): Promise<T> =>
+export const postRequest = <T>(url: string, requestData: RequestData): Promise<T> =>
```
osamasayed, line 32: This shouldn't be exported and instead we can export the function that does the actual calling.

src/components/Login/EmailVerificationForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+type EmailVerificationFormProps = {
+  emailFormField: FormField;
+};
+const EmailVerificationForm = ({ emailFormField }: EmailVerificationFormProps) => {
+  const [isSubmitted, setIsSubmitted] = useState(false);
+  const { mutate } = useSWRConfig();
+  const [email, setEmail] = useState();
+  const { t } = useTranslation('common');
+  const requestVerificationCode = async (emailToVerify) => {
+    return postRequest(makeVerificationCodeUrl(), { email: emailToVerify });
```
osamasayed, line 27: We need to have a function that does the actual API call instead of using `postRequest` directly

src/components/Login/EmailVerificationForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+              components={{
+                strong: <strong className={styles.strong} />,
+              }}
+            />
+          </p>
+          <FormBuilder
+            formFields={[
+              {
+                field: 'code',
+                type: FormFieldType.Text,
```
osamasayed, line 73: Shouldn't this be a number?

src/components/Login/EmailVerificationForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [isSubmitted, setIsSubmitted] = useState(false);
+  const { mutate } = useSWRConfig();
+  const [email, setEmail] = useState();
+  const { t } = useTranslation('common');
+  const requestVerificationCode = async (emailToVerify) => {
+    return postRequest(makeVerificationCodeUrl(), { email: emailToVerify });
+  };
+  const [forceRerenderKey, setForceRerenderKey] = useState(0);
+
+  const onEmailSubmitted = (data) => {
```
osamasayed, line 31: `data` should have a type.

src/components/Login/EmailVerificationForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+              {
+                field: 'code',
+                type: FormFieldType.Text,
+                rules: [
+                  { type: RuleType.Required, value: true, errorId: ErrorMessageId.RequiredField },
+                ],
+              },
+            ]}
+            onSubmit={onVerificationCodeSubmitted}
+            actionText="Verify Code"
```
osamasayed, line 80: This text should be localized

src/components/Login/EmailVerificationForm.tsx
osamasayed, line 73: Not a strong opinion, but this can be extracted to a const outside of the component since it's not dynamic

muhajirdev: I completely agree

osamasayed (approved):

PR #1686, opened 2022-06-18
Image radius fix

tumon-art:
**A small change to remove image radius while the container itself has border-radius.**
My fix:
fixed2

Real look: 
prob1


osamasayed (approved):
Jazak Allah Khairn for your contribution @tumon-art, I have removed the un-used css line instead of commenting it out for better code organization.

PR #1698, opened 2022-06-21
added a custom scrollbar for RecentReadingSessions

tumon-art:
**🚨 Currently, we don't have any scrollbar to see overflowed contents when using the desktop.**

**_This scrollbar will only appear for desktops ✔️
Suitable with all themes ✔️_**

### Screenshots

| Before | After |
| ------ | ------ |
| old3|new3|


devmahdi404:
Nice job.

I think there should be custom scrollbars in some other places too.
The default scrollbar doesn't go well with the UI.
Most of the modern browsers support custom scrollbar.

PR #1705, opened 2022-06-24
fix: bring back recent reading sessions for logged out user

muhajirdev:


src/components/Verses/RecentReadingSessions.tsx
```diff
@@ -31,17 +34,27 @@ const RecentReadingSessions = () => {
     privateFetcher,
   );
 
-  if (!data) return null;
+  const recentReadingSessionsVerseKey = useMemo(() => {
```
osamasayed, line 37: nit: should be `recentReadVerseKeys`

osamasayed (approved):

PR #1706, opened 2022-06-24
fix: fix lang switcher for logged out user

muhajirdev:


src/components/Navbar/LanguageSelector.tsx
```diff
@@ -59,14 +63,38 @@ const LanguageSelector = ({
...
+        PreferenceGroup.LANGUAGE,
+      ).catch(() => {
+        toast(t('error.pref-persist-fail'), {
+          status: ToastStatus.Warning,
+          actions: [
+            {
+              text: t('undo'),
+              primary: true,
+              onClick: () => {
+                setLanguage(newLocale);
```
osamasayed, line 83: we should add `await`

osamasayed (approved):

PR #1725, opened 2022-07-24
Feature toggle notion

Ashraf-Ali-aa:
### Summary
With this change, we check if notion env is set before initialising the client


Ashraf-Ali-aa:
Assalamu Alaikum @osamasayed any chance of a code review

osamasayed:
> Assalamu Alaikum @osamasayed any chance of a code review

Wa'likum Alsalam, Jazak Allah Kahiran for your contribution. Will take a look shortly

src/lib/notion.ts
```diff
@@ -1,11 +1,23 @@
...
-const notion = new Client({
-  auth: process.env.NOTION_TOKEN,
-});
+const isNotionEnabled = () => {
+  return !!process.env.NOTION_TOKEN;
+};
+
+let notion: Client;
+
+if (isNotionEnabled) {
```
osamasayed, line 10: Since `isNotionEnabled` is a function, shouldn't this be `if (isNotionEnabled()) {` instead of `if (isNotionEnabled) {`? What is even better is to keep it as it is and change `isNotionEnabled` from being a function to just a fixed const e.g. `const isNotionEnabled = !!process.env.NOTION_TOKEN`.

PR #1733, opened 2022-08-03
[Auth] stop calling logout on delete

osamasayed:


muhajirdev (approved):

src/components/Profile/DeleteAccountButton.tsx
```diff
@@ -33,8 +31,6 @@ const DeleteAccountButton = ({ isDisabled }: DeleteAccountButtonProps) => {
   const onDeleteConfirmed = () => {
     closeModal();
     deleteAccount()
-      .then(() => logoutUser())
-      .then(() => mutate(makeUserProfileUrl()))
```
muhajirdev, line 37: I think we still need to mutate the profile @osamasayed ?

to clear out the previous data?

Issue #1760, opened 2022-08-26
Add route for Hizb


osamasayed:
Currently Quran.com supports routes for:

- a Surah e.g. https://quran.com/al-kahf
- A verse e.g. https://quran.com/al-kahf/1
- A juz e.g. https://quran.com/juz/1
- A page e.g. https://quran.com/page/1

We should also support for reading a specific hizb as well. The route should be like `/hizb/{hizbNumber}`. Quran.com already has an [API](https://quran.api-docs.io/v4/verses/by-hizb-number) that gets the verses of a specific hizb.

ahmedriad1:
Working on this now

PR #1771, opened 2022-08-28
Add hizb route

ahmedriad1:
### Summary
This is an implementation for #1760 

### Screenshots


ahmedriad1:
Still need to update the search API to add hizb navigation (https://staging.quran.com/api/qdc/search?query=hizb & https://api.qurancdn.com/api/qdc/search?query=hizb)

osamasayed:
Jazak Allah Khairan @ahmedriad1 taking a look

locales/ar/common.json
```diff
@@ -247,6 +247,7 @@
     "shared": "تم نسخ الرابط إلى الحافظة!",
     "sidebar": {
         "search-juz": "ابحث عن جزء",
+        "search-hizb": "ابحث عن حزب",
```
osamasayed, line 250: @ahmedriad1 usually we add a new string in the English file only and use [Lokalise](https://lokalise.com/) to generate the same string in all the other languages we support. I will do that in a follow up. We should undo `/ar/quran-reader.json` and `/ar/common.json`.

osamasayed:
@ahmedriad1 code looks great mashAllah. I have 2 comments:

1. I think that adding the Hizbs list to the sidebar navigation made it look a bit tight and given that a Rub' will also be added, it will be even tighter particularly on mobiles so the team needs to think about how the UI should look like first before adding this section. So for now, you can comment out the `if` statement that renders the `HizbSelection` component as we will still need it after we have a plan about the UI inshAllah.

Screen Shot 2022-08-29 at 9 15 22 AM

2. I noticed an issue when you open a Hizb's route e.g. `/hizb/1` the Surahs selection list doesn't show the current Surah being read. Page and Juz show correctly but the current Surah is not being highlighted in the Surahs list.
Screen Shot 2022-08-29 at 9 09 12 AM

locales/ar/common.json
ahmedriad1: I see. I'll remove the arabic translation now

ahmedriad1:
JazakAllah khairan for the review Osama. I'll comment the `HizbSelection` and fix the sidebar navigator issue.

ahmedriad1:
I see that `/juz/[juzId]` also had a problem in sidebar navigation (the chapter and verse list were not being updated when reading another surah):
screen



The last commit solves this problem, fixing the issue in `/hizb/[hizbId]` out of the box. This is its new behavior :
screen2

src/hooks/useChapterId.ts
```diff
@@ -53,9 +54,11 @@ const useChapterIdsByUrlPath = (lang: string): string[] => {
         setChapterIds(chapterIdsForPage);
       } else if (juzId) {
         setChapterIds(await getChapterIdsForJuz(formatStringNumber(juzId as string)));
-      }
+      } // else if (hizbId) {
```
osamasayed, line 57: Since we are not using this anymore, I suggest reverting this page hook back to how it was.

src/utils/hizb.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param {string} hizbId
+ * @returns {string[]} chapterIds
+ */
+// export const getChapterIdsForHizb = async (hizbId: string): Promise<string[]> => {
+//   return new Promise((res) => {
+//     import(`../../data/hizb-to-chapter-mappings.json`).then((data) => {
+//       res(data.default[hizbId]);
+//     });
+//   });
+// };
```
osamasayed, line 42: This bit should be removed from the file.

src/hooks/useChapterId.ts
ahmedriad1: Sure, but I thought we'd need it in the future because we're using this hook in `components/AudioPlayer/Buttons/PlayPauseButton.tsx`. I haven't tested the audio player, yet.

osamasayed: Oh sorry let me think of it. Will get back to you inshAllah

osamasayed (approved):
Salam @ahmedriad1 PR looks great! Feel free to merge. I update it tiny bit to be used with audio.

ahmedriad1:
> Salam @ahmedriad1 PR looks great! Feel free to merge. I update it tiny bit to be used with audio.

Salam, @osamasayed. Great but I don't think I have push access to the repo 😅

osamasayed:
> > Salam @ahmedriad1 PR looks great! Feel free to merge. I update it tiny bit to be used with audio.
> 
> Salam, @osamasayed. Great but I don't think I have push access to the repo 😅

Oh okay let me merge it :)

Issue #1773, opened 2022-08-29
[bug]: Selected search page doesn't change after changing search query


osamasayed:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

When changing the search query that has fewer than 10 results while on page 2, it:

1. Doesn't show the new search results until you click on page 1 again.
2. The pagination footer showing the wrong number of results e.g. "11-8 OF 8 SEARCH RESULTS".

<img width="847" alt="Screen Shot 2022-08-29 at 7 34 37 PM" src="https://user-images.githubusercontent.com/15169499/187202075-31a35b20-790c-4b56-a153-8b531745aafe.png">

### Steps to reproduce

1. Go to https://quran.com/search
2. Type search query 'earth' and click search
3. Scroll down the results and go to page 2 (or any page that is not page 1)
4. Now search for 'ten' in the search bar
5. Search results will be blank until you have to click page '1' to see the results

### Environment (please complete the following information) and Add any other context about this bug

OS: Mac Monterey 12.5.1
Browser: Chrome
Version: 104.0.5112.101

### Environment

Production(quran.com)

PR #1774, opened 2022-08-29
Fix search pagination

ahmedriad1:
### Summary
This is a fix for #1773

### Screenshots

| Before | After |
| ------ | ------ |
| Before |  |

### DIFF
--- a/src/pages/search.tsx
+++ b/src/pages/search.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-lines */
-import React, { useState, useEffect, useMemo, useCallback } from 'react';
+import React, { useState, useEffect, useMemo, useCallback, useRef } from 'react';
 
 import { GetStaticProps, NextPage } from 'next';
 import useTranslation from 'next-translate/useTranslation';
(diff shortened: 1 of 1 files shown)


AtakanColak:
Let's say user searches 'earth' and goes to page 35. Then searches 'heaven', 

My PR will redirect the user to page 25, the maximum possible page for that word.

In this PR, user will be redirected back to page 1. 

I'm going to suggest that when a new search is made, we should always force the user to page 1 and avoid adding any other code either with this PR or mine. I don't see any point in the user wanting to stay in the same page number for a new search result. Please let me know if I'm overlooking anything.

ahmedriad1:
> Let's say user searches 'earth' and goes to page 35. Then searches 'heaven',
> 
> My PR will redirect the user to page 25, the maximum possible page for that word.
> 
> In this PR, user will be redirected back to page 1.
> 
> I'm going to suggest that when a new search is made, we should always force the user to page 1 and avoid adding any other code either with this PR or mine. I don't see any point in the user wanting to stay in the same page number for a new search result. Please let me know if I'm overlooking anything.

Yeah that's exactly what I did

> Edit: Hmm. Pardon me, I think that is literally what you have done in this PR.

@AtakanColak No worries.

src/pages/search.tsx
```diff
@@ -158,17 +158,36 @@ const Search: NextPage<SearchProps> = ({ translations, chaptersData }): JSX.Elem
     [],
   );
 
+  // a ref to know weather this is the initial search request made when the user loads the page or not
```
osamasayed, line 161: nit-picking: weather -> `whether`.

src/pages/search.tsx
```diff
@@ -158,17 +158,36 @@ const Search: NextPage<SearchProps> = ({ translations, chaptersData }): JSX.Elem
...
+    // we don't want to run this effect when currentPage is changed
+    // because we are already handeling this in onPageChange
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [debouncedSearchQuery, getResults, selectedLanguages, selectedTranslations]);
 
-  const onPageChange = (page: number) => {
+  const onPageChange = async (page: number) => {
     logEvent('search_page_number_change');
     setCurrentPage(page);
+    await getResults(debouncedSearchQuery, page, selectedTranslations, selectedLanguages);
```
osamasayed, line 190: It shows this message since `getResults` doesn't return a `Promise`.

Screen Shot 2022-09-01 at 4 20 54 PM

src/pages/search.tsx
```diff
@@ -158,17 +158,36 @@ const Search: NextPage<SearchProps> = ({ translations, chaptersData }): JSX.Elem
...
+  // a ref to know weather this is the initial search request made when the user loads the page or not
+  const isInitialSearch = useRef(true);
+
   // listen to any changes in the API params and call BE on change.
   useEffect(() => {
     // only when the search query has a value we call the API.
     if (debouncedSearchQuery) {
-      getResults(debouncedSearchQuery, currentPage, selectedTranslations, selectedLanguages);
+      // we don't want to reset pagination when the user reloads the page with a ?page={number} in the url query
+      if (!isInitialSearch.current) setCurrentPage(1);
```
osamasayed, line 169: not-picking, we have been using curly brackets even for a single-line condition.
```
if (!isInitialSearch.current) {
  setCurrentPage(1);
}
```

src/pages/search.tsx
```diff
@@ -158,17 +158,36 @@ const Search: NextPage<SearchProps> = ({ translations, chaptersData }): JSX.Elem
...
+      getResults(
+        debouncedSearchQuery,
+        // if it is the initial search request, use the page number in the url, otherwise, reset it
+        isInitialSearch.current ? currentPage : 1,
+        selectedTranslations,
+        selectedLanguages,
+      );
+
+      // if it was the initial request, update the ref
+      if (isInitialSearch.current) isInitialSearch.current = false;
```
osamasayed, line 180: ditto. We should use `{}`.

src/pages/search.tsx
ahmedriad1: Done

src/pages/search.tsx
ahmedriad1: Done

src/pages/search.tsx
ahmedriad1: Maybe we should add an eslint rule ?

src/pages/search.tsx
ahmedriad1: Maybe we should add an eslint rule ?

osamasayed: Yes would be great to do so in another PR! Thanks a lot

osamasayed (approved):

PR #1779, opened 2022-08-31
Migrate from Jest to Vitest

ahmedriad1:
### Summary
Currently, we use [jest](https://jestjs.io/) for unit tests. However, jest is too slow and requires a lot of dependencies to work properly. So, I propose migrating to [vitest](https://vitest.dev/) - a fast ⚡, light-weight, and better jest-compatible testing framework. Also, the stats after migrating are insane, tests went from `52s` to `4s`.

### Stats

| Before | After |
| ------ | ------ |
| jest | vitest |


shakib2018:
very very good decision

muhajirdev:
52s to 4s is very significant. 
If there's no gotchas. Looks good to me

muhajirdev:
I found someone is writing this https://dev.to/jjhiggz/comment/1o4a9 . 
```
Cons

it seems to crash a bit more often
````

I don't know how valid the statement is

ahmedriad1:
@muhajirdev 
> I found someone is writing this https://dev.to/jjhiggz/comment/1o4a9 .
> 
> ```
> Cons
> 
> it seems to crash a bit more often
> ```
> 
> I don't know how valid the statement is

The comment was posted on May 2nd. I guess it crashed more often back then. But I think it's production ready now since Vue removed the warning from their [docs](https://vuejs.org/guide/scaling-up/testing.html#unit-testing), and Remix is also using it for their [stacks](https://remix.run/stacks). But, let me know what you think 🤔

osamasayed:
I think it looks promising!! Let's 🛳 it!

osamasayed (approved):

PR #1803, opened 2022-09-18
feat: add delete button to bookmark list

muhajirdev:


src/components/Verses/BookmarkedVersesList.tsx
```diff
@@ -56,6 +68,30 @@ const BookmarkedVersesList: React.FC = () => {
     return null;
   }
 
+  const onBookmarkDeleted = (verseKey) => () => {
+    if (isLoggedIn()) {
```
osamasayed, line 72: Add logging

src/components/Verses/BookmarkedVersesList.tsx
```diff
@@ -69,18 +105,25 @@ const BookmarkedVersesList: React.FC = () => {
...
-                  }}
-                >
-                  {bookmarkText}
-                </Button>
+                <div key={verseKey} className={styles.bookmarkItem}>
+                  <Link href={getVerseNavigationUrlByVerseKey(verseKey)}>
+                    <a
+                      className={styles.linkButtonContainer}
+                      onClick={() => {
+                        logButtonClick('homepage_bookmark');
```
osamasayed, line 113: nit: this component is shared between the home page and profile page so we need to rename the log key. Also we should extract the function outside

osamasayed (approved):

PR #1804, opened 2022-09-19
feat: click play ayah should always start from the verse

### LINKED ISSUE #1788: [feature]: Two Play buttons do the same thing

### Summary

Hello,
As we know, there are two play buttons on each particular Surah, and in the previous version of Quran.com the play button on each Ayah started it from the beginning, and this feature made it helpful while you try to memorize it. But in this version, both play buttons play the same role by playing the Ayah from the place where you stopped.
Untitled design

### Your purposed solution for this feature

In my opinion, it is better if the play button for each Ayah starts it from beginning and the play button on the bottom plays the Ayah from the place where you stopped.

muhajirdev:
Fixes #1788

### DIFF
--- a/src/xstate/actors/audioPlayer/audioPlayerMachine.ts
+++ b/src/xstate/actors/audioPlayer/audioPlayerMachine.ts
@@ -829,10 +829,7 @@ export const audioPlayerMachine =
         }),
         setSurahAndAyahNumbers: assign({
           surah: (context, event) => event.surah,
-          ayahNumber: (context, event) => {
-            console.log(event);
-            return event.ayahNumber;
-          },
+          ayahNumber: (context, event) => event.ayahNumber,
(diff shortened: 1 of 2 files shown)


src/xstate/actors/audioPlayer/audioPlayerMachine.ts
```diff
@@ -824,7 +829,10 @@ export const audioPlayerMachine =
         }),
         setSurahAndAyahNumbers: assign({
           surah: (context, event) => event.surah,
-          ayahNumber: (context, event) => event.ayahNumber,
+          ayahNumber: (context, event) => {
+            console.log(event);
```
osamasayed, line 833: `console.log(event);` should be removed

osamasayed (approved):

PR #1809, opened 2022-09-24
Remove audio mismatch modal

muhajirdev:


src/xstate/actors/audioPlayer/audioPlayerMachine.ts
```diff
@@ -419,10 +409,10 @@ export const audioPlayerMachine =
                     },
                     PLAY_AYAH: [
                       {
-                        actions: ['setNewSurahAndAyahNumbers'],
+                        actions: ['setSurahAndAyahNumbers'],
```
osamasayed, line 412: everywhere we added `actions: ['setSurahAndAyahNumbers'],`, we should also add `exitRadio` and `stopRepeatActor` actions which we were triggering when `CONFIRM_PLAY_MISMATCHED_SURAH ` is fired. This is necessary to exit repeat mode/radio mode when the user plays another Surah.
Screen Shot 2022-09-24 at 7 35 42 PM

osamasayed (approved):

Issue #1770, opened 2022-08-28
[bug]: Loading skeleton shows text


osamasayed:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

When I open the settings drawer and while the translations list is loading, I see the current translation Id being loaded showing in the skeleton:

![RPReplay_Final1661669923](https://user-images.githubusercontent.com/15169499/187062886-81929033-26d4-4724-bbea-1622132b2cac.gif)

### Steps to reproduce

1. Go to Surah https://quran.com
2. Open the settings drawer
3. Scroll down to the translations or reciters section
4. While the translations/reciters list is loading, you will see text on top of the skeleton.

### Environment (please complete the following information) and Add any other context about this bug

OS: iPhone X 15.6
Browser: Chrome
Version: 104.0.5112.99

### Environment

Production(quran.com)

Mohammed-Yasin-Mulla:
#1772

njmulsqb:
Assign this issue to me!

abdulhamidOumer:
@osamasayed I created a PR for this issue. Please let me know if this is an ideal solution. This also resolves the same issue in the tafsir section (#895 )

@njmulsqb Afwen, I just saw that you wanted to work on this issue after I created a PR.

njmulsqb:
No problem dude. Glad you took this, hope the maintainers approve your PR.

PR #1819, opened 2022-10-02
Fix: Loading skeleton showing text on setting drawer

abdulhamidOumer:
### Summary
This PR resolves issue #1770. Skeleton of translation and tafsir sections in the setting drawer were rendering IDs inside the skeleton. I am guessing it was implemented like that previously in order to have the id texts as a placeholder. 
This PR resolves the issue by removing ID renders and adding a style to the skeletons like it's done in other places like [LanguageAndTafsirSelection](https://github.com/quran/quran.com-frontend-next/blob/master/src/components/QuranReader/TafsirView/LanguageAndTafsirSelection.tsx#L32)

### Test Plan
I was not able to reproduce the issue on Chrome 106.0 Mac OS, but I was able to replicate it in Safari 16.0. In order to test this issue, you can replace `DataFetcher` block (line 139) with `translationLoading()` on [TranslationSection](https://github.com/quran/quran.com-frontend-next/blob/master/src/components/Navbar/SettingsDrawer/TranslationSection.tsx)

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2022-10-02 at 5 30 15 PM|Screen Shot 2022-10-02 at 5 31 34 PM|

### DIFF
--- a/src/components/Navbar/SettingsDrawer/TafsirSection.tsx
+++ b/src/components/Navbar/SettingsDrawer/TafsirSection.tsx
@@ -36,9 +36,7 @@ const TafsirSection = () => {
     () => (
       <div>
         {selectedTafsirs.map((id) => (
-          <Skeleton key={id}>
-            <div>{id}</div>
-          </Skeleton>
+          <Skeleton className={styles.skeleton} key={id} />
         ))}
       </div>
     ),
--- a/src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
+++ b/src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
@@ -8,6 +8,6 @@
 }
 
 .skeleton {
-  height: calc(1.5 * var(--spacing-mega));
+  height: calc(2.2 * var(--spacing-mega));
   width: 100%;
 }
--- a/src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
+++ b/src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
@@ -6,3 +6,8 @@
   flex-direction: column;
   margin-block-start: var(--spacing-xsmall);
 }
+
+.skeleton {
+  height: calc(1.5 * var(--spacing-mega));
+  width: 100%;
+}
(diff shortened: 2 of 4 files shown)


src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
```diff
@@ -6,3 +6,8 @@
   flex-direction: column;
   margin-block-start: var(--spacing-xsmall);
 }
+
+.skeleton {
+  height: calc(1.5 * var(--spacing-mega));
```
muhajirdev, line 11: Salam brother @abdulhamidOumer . JazakAllah for the contribution. 

I just did a quick test. On my safari macbook, it seems that we need a bigger height?

For a reference, this 1.5 * spacing mega skeleton side by side with the loaded component
Screen Shot 2022-10-04 at 13 37 12

and this is 2.2
image

other than that, looks good Masha Allah

src/components/Navbar/SettingsDrawer/TafsirSection.tsx
```diff
@@ -36,9 +36,7 @@ const TafsirSection = () => {
     () => (
       <div>
         {selectedTafsirs.map((id) => (
-          <Skeleton key={id}>
```
muhajirdev, line 39: and this tafsir component doesn't seems to be used in the settings drawer anymore. but it's okay :), lets get the changes in

src/components/Navbar/SettingsDrawer/TranslationSection.module.scss
abdulhamidOumer: Thank you for the review @muhajirdev 👍 . You're right when it's multiplied by `1.5` the skeleton height is smaller than what will be rendered. As suggested multiplying it with `2.2` matches the height somehow. 

I have pushed a fix for that.

src/components/Navbar/SettingsDrawer/TafsirSection.tsx
abdulhamidOumer: Yes, I had to add it to settings in order to test it as well 🙂 .  I added it just in case it's an upcoming feature.

muhajirdev:
Thank you @abdulhamidOumer

PR #1838, opened 2022-11-11
styles: add isSidebarNavigationVisible check & update the footer styles

### LINKED ISSUE #1837: [bug]: SidebarNavigation overlap the Footer in the Quran reader page

### Describe the bug

Salam Alaikum, 

Just this is a small UI issue.

On the Quran reader page, when opening the side navbar we see that the Footer is hidden under the side navbar while the reset of the page adjust to the available width.

2022-11-11 22 54 12 quran com d70f3efc3d9b

### Relevant log output

_No response_

### Steps to reproduce

1. Select any Surah
2. Scroll to the bottom of the page until you see the footer
3. Open the side navbar

OS: Linux mint
Browser: Brave v1.25.72

### Environment

Production(quran.com)

AM-77:
### Summary
Fixes #1837

On the Quran reader page, when opening the side navbar we see that the Footer is hidden under the side navbar while the reset of the page adjust to the available width.

### Test Plan

1. Select any Surah
2. Scroll to the bottom of the page until you see the footer
3. Open the side navbar

### Screenshots

| Before | After |
| ------ | ------ |
| 2022-11-11 22 54 12 quran com d70f3efc3d9b | 2022-11-11 23 20 42 localhost bfdb7a858ef2 |


tifa2UP (changes requested):
Thank you for the change @AM-77. We should modify the footer when the SidebarNavigation is visible AND the user is in the quran reader. Otherwise other pages will have an incorrect UI like this:

image

AM-77:
Nice catch @tifa2UP ,  it should be fixed now.

PR #1841, opened 2022-11-13
fix: add chaptersData to fix the audio player running on some pages.

### LINKED ISSUE #1840: [bug]: The app crash when playing audio, and navigating to /about-us, /apps, /developers ... pages

### Describe the bug

Salam Alaikoum,

When the Audio Player is open, and navigating to one of this pages, the app crashes.

- https://quran.com/about-us
- https://quran.com/apps
- https://quran.com/developers
- https://quran.com/privacy
- https://quran.com/product-updates
- https://quran.com/support
- https://quran.com/login

#### Relevant log output:

 [quran.com-1668354047077.log](https://github.com/quran/quran.com-frontend-next/files/9997631/quran.com-1668354047077.log)

### Relevant log output

_No response_

### Steps to reproduce

1. navigate to any sourah and play the audio.
2. keep the audio playing and navigate to one of this pages
- https://quran.com/about-us
- https://quran.com/apps
- https://quran.com/developers
- https://quran.com/privacy
- https://quran.com/product-updates
- https://quran.com/support
- https://quran.com/login

you can find all this links in the footer, in the **Navigate** section. 

OS: Linux mint
Browser: Brave v1.25.72

### Environment

Production(quran.com)

AM-77:
### Summary
Salam Alaikoum,

This PR is an attempt to fix #1840

### Test Plan

1. navigate to any sourah and play the audio.
2. keep the audio playing and navigate to one of this pages

https://quran.com/about-us
https://quran.com/apps
https://quran.com/developers
https://quran.com/privacy
https://quran.com/product-updates
https://quran.com/support
https://quran.com/login

you can find all this links in the footer, in the Navigate section.

### Screenshots

| Before | After |
| ------ | ------ |
| 2022-11-13 16 35 21 quran com 91fd2d2d34e9 | 2022-11-13 16 37 59 localhost 589ec2debfa1 |

### DIFF
--- a/src/pages/about-us.tsx
+++ b/src/pages/about-us.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable jsx-a11y/anchor-has-content */
 /* eslint-disable jsx-a11y/control-has-associated-label */
+import { GetStaticProps, NextPage } from 'next';
 import Trans from 'next-translate/Trans';
 import useTranslation from 'next-translate/useTranslation';
 import Image from 'next/image';
@@ -9,13 +10,20 @@ import styles from './contentPage.module.scss';
 import NextSeoWrapper from '@/components/NextSeoWrapper';
 import PageContainer from '@/components/PageContainer';
 import Link from '@/dls/Link/Link';
+import ChaptersData from '@/types/ChaptersData';
+import { getAllChaptersData } from '@/utils/chapter';
 import { logTarteelLinkClick } from '@/utils/eventLogger';
 import { getBlurDataUrl } from '@/utils/image';
 import { getLanguageAlternates } from '@/utils/locale';
 import { getCanonicalUrl } from '@/utils/navigation';
 
 const path = '/about-us';
-const AboutUsPage = () => {
+
+type AboutUsPageProps = {
+  chaptersData: ChaptersData;
+};
+
+const AboutUsPage: NextPage<AboutUsPageProps> = (): JSX.Element => {
   const { t, lang } = useTranslation('about');
 
   const onTarteelLinkClicked = () => {
@@ -148,6 +156,16 @@ const AboutUsPage = () => {
       </PageContainer>
     </>
   );
+};
+
+export const getStaticProps: GetStaticProps = async ({ locale }) => {
+  const allChaptersData = await getAllChaptersData(locale);
+
+  return {
+    props: {
+      chaptersData: allChaptersData,
+    },
+  };
 };
 
 export default AboutUsPage;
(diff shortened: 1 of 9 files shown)


tifa2UP (changes requested):
Thank you for making the changes @AM-77. While this fixes the audio player issue, pages such as about and apps don't need chapter data to be in their scope. Can we instead load chapter data into the audio player directly whenever it's active?

src/pages/about-us.tsx
```diff
@@ -9,13 +10,61 @@ import styles from './contentPage.module.scss';
...
+    alt: 'Lokalise',
+  },
+  {
+    key: 6,
+    i18nKey: 'about:credits.vercel',
+    href: 'https://vercel.com/?utm_source=quran-pro&utm_campaign=oss',
+    image: '/images/powered-by-vercel.svg',
+    alt: 'Vercel',
+  },
+];
```
hazem3500, line 61: This change is out of the scope of the PR, right? 🤔 

I would leave refactoring work in a separate issue and just narrow down the changes to what's required in the PR 😄

hazem3500:
@tifa2UP re-evaluating this and thinking a bit from first principles, why don't we store the `chaptersData` on the client side to begin with 😅
Last I checked, the Quran isn't getting any updates anytime soon 😆👀 so it's basically calling the backend for static data that would never change 🤷 (unless we want to add a new property to the `chaptersData` object? 👀)

I've looked into the code, and it felt a bit off to me that we needed the whole `DataContext`, `useContext`, `DataContext.Provider` dance just to get the `chaptersData`it feels a bit overkill for me 😅

I've checked the size of the `chaptersData` JSON, and it should be ~15KB, so it shouldn't be much of a problem 🤷 + if we really want to squeeze the initial bundle size, we can defer loading the `chaptersData` JSON, and just lazy-load it after initial render 🤷 

Though since `DataContext` is referenced 135 times in the code, it's probably won't be a quick fix 😅
Another easy solution would be to cache the `chaptersData` API response indefinitely, and that way, each time we call that endpoint, it gets the JSON straight from the cache 👍 

Also, side note, your original suggestion was moving fetching `chaptersData` to the `AudioPlayer` instead of the page. Wouldn't that move the fetching from the server side to the client side? 👀 would that be something that we'd want? 🤔 I guess it's a small request anyways, so shouldn't be much of a problem 😅

hazem3500:
Okay, relooking at this, it looks like I've had the oversight of not seeing that there is a JSON for each locale 😅👀🙈

So storing client-side might not be the best idea? 🤔 
Though, caching indefinitely still sounds like a good idea to me 🤔 (or at least keep it indefinite until we bust it if needed 🤷 )

PR #1857, opened 2022-12-22
Fix persist gate double mounting

tifa2UP:
Previously we had redux persist gate logic that would cause all the DOM elements to be unmounted on initial page load during the redux-persist hydration, and mounted again right after. 

This was caused by the following piece of code: 

image

I've modified the code to use a single dom tree for both the hydrated and non-hydrated states. Redux uses the non persist initial state, and then dispatches a `REHYDRATE` event to populate the state. With this approach, a bunch of animations would happen on initial load (because the state is modified). I added a global utility that prevents all animations until the hydration is complete.

### DIFF
--- a/src/components/DisableAnimationsUntilHydration.tsx
+++ b/src/components/DisableAnimationsUntilHydration.tsx
@@ -2,13 +2,13 @@ import React from 'react';
 
 import { useSelector } from 'react-redux';
 
-import { selectisPersistGateHydrationComplete } from '../redux/slices/persistGateHydration';
-import isClient from '../utils/isClient';
+import { selectIsPersistGateHydrationComplete } from '@/redux/slices/persistGateHydration';
+import isClient from '@/utils/isClient';
 
 // This utility disables all animations until the redux-persist store hydration is complete.
-// This is done to prevent a bunch of animations from happening on iniitial page load.
+// This is done to prevent a bunch of animations from happening on initial page load.
 const DisableAnimationsUntilHydration = () => {
-  const isPersistGateHydrationComplete = useSelector(selectisPersistGateHydrationComplete);
+  const isPersistGateHydrationComplete = useSelector(selectIsPersistGateHydrationComplete);
 
   if (isClient && !isPersistGateHydrationComplete) {
     return (
--- a/src/components/DisableAnimationsUntilHydration.tsx
+++ b/src/components/DisableAnimationsUntilHydration.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+
+import { useSelector } from 'react-redux';
+
+import { selectisPersistGateHydrationComplete } from '../redux/slices/persistGateHydration';
+import isClient from '../utils/isClient';
+
+// This utility disables all animations until the redux-persist store hydration is complete.
+// This is done to prevent a bunch of animations from happening on iniitial page load.
+const DisableAnimationsUntilHydration = () => {
+  const isPersistGateHydrationComplete = useSelector(selectisPersistGateHydrationComplete);
+
+  if (isClient && !isPersistGateHydrationComplete) {
+    return (
+      <style jsx global>
+        {`
+          * {
+            transition: all 0s !important;
+          }
+        `}
+      </style>
+    );
+  }
+
+  return <></>;
+};
+
+export default DisableAnimationsUntilHydration;
--- a/src/components/DisableAnimationsUntilHydration.tsx
+++ b/src/components/DisableAnimationsUntilHydration.tsx
@@ -12,7 +12,7 @@ const DisableAnimationsUntilHydration = () => {
 
   if (isClient && !isPersistGateHydrationComplete) {
     return (
-      <style jsx global>
+      <style>
         {`
           * {
             transition: all 0s !important;
--- a/src/components/GlobalPersistGateHydrationListener.tsx
+++ b/src/components/GlobalPersistGateHydrationListener.tsx
@@ -1,0 +1,33 @@
+import React, { useEffect } from 'react';
+
+import { useDispatch, useSelector } from 'react-redux';
+
+import {
+  selectisPersistGateHydrationInProgress,
+  setIsPersistGateHydrationComplete,
+  setIsPersistGateHydrationInProgress,
+} from '../redux/slices/persistGateHydration';
+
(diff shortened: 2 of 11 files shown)


src/redux/slices/persistGateHydration.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+  },
+});
+
+export const { setIsPersistGateHydrationInProgress, setIsPersistGateHydrationComplete } =
+  persistGateHydrationSlice.actions;
+
+export const selectisPersistGateHydrationInProgress = (state: RootState) =>
+  state.persistGateHydration.isPersistGateHydrationInProgress;
+export const selectisPersistGateHydrationComplete = (state: RootState) =>
```
osamasayed, line 53: nit: `selectIsPersistGateHydrationComplete` and `selectIsPersistGateHydrationInProgress`.

src/utils/isClient.ts
```diff
@@ -0,0 +1,8 @@
+// A utility to check if the code is running on the client (vs. nextjs server)
+const isClient = !!(
```
osamasayed, line 2: Thank you for adding this, it was bothering me having to copy and paste `isClient` multiple times haha

src/components/DisableAnimationsUntilHydration.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { useSelector } from 'react-redux';
+
+import { selectisPersistGateHydrationComplete } from '../redux/slices/persistGateHydration';
+import isClient from '../utils/isClient';
+
+// This utility disables all animations until the redux-persist store hydration is complete.
+// This is done to prevent a bunch of animations from happening on iniitial page load.
```
osamasayed, line 9: nit: `iniitial` -> `initial`.

src/components/DisableAnimationsUntilHydration.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { useSelector } from 'react-redux';
+
+import { selectisPersistGateHydrationComplete } from '../redux/slices/persistGateHydration';
+import isClient from '../utils/isClient';
```
osamasayed, line 6: nit: we should use `@` when importing instead of `..`.

src/redux/slices/persistGateHydration.ts
```diff
@@ -0,0 +1,56 @@
...
+      state: PersistGateHydrationState,
+      action: PayloadAction<boolean>,
+    ) => ({
+      ...state,
+      isPersistGateHydrationComplete: action.payload,
+    }),
+  },
+  extraReducers: (builder) => {
+    builder.addCase(REHYDRATE, (state: PersistGateHydrationState) => {
+      // eslint-disable-next-line no-param-reassign
```
osamasayed, line 42: I think the reducer should not directly manipulate `state` but rather return a new object e.g.:

```
{
      ...state,
      isPersistGateHydrationInProgress: action.payload,
    }
```

You can refer to other instances of `builder.addCase` for example:

Screen Shot 2022-12-27 at 11 45 08 AM

src/redux/slices/persistGateHydration.ts
```diff
@@ -0,0 +1,56 @@
...
+import SliceName from '@/redux/types/SliceName';
+
+export type PersistGateHydrationState = {
+  isPersistGateHydrationInProgress: boolean;
+  isPersistGateHydrationComplete: boolean;
+};
+
+const initialState: PersistGateHydrationState = {
+  isPersistGateHydrationInProgress: false,
+  isPersistGateHydrationComplete: false,
```
osamasayed, line 15: Quick question, do we need 2 states? cannot we just set only 1 state since from what I am getting, we only use either one of them and not both at the same time.

src/components/GlobalPersistGateHydrationListener.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const dispatch = useDispatch();
+  const isPersistGateHydrationInProgress = useSelector(selectisPersistGateHydrationInProgress);
+
+  useEffect(() => {
+    if (isPersistGateHydrationInProgress) {
+      setTimeout(() => {
+        dispatch({ type: setIsPersistGateHydrationComplete.type, payload: true });
+        dispatch({ type: setIsPersistGateHydrationInProgress.type, payload: false });
+      }, PERSIST_GATE_HYDRATION_DURATION_MS);
+    }
```
osamasayed, line 28: I believe we should add `useEffect`'s cleanup function to clear the internal.

```
return () => {
        clearTimeout(handler);
      };
```

You can refer to `src/hooks/useDebounce.ts` for an example.

src/components/GlobalPersistGateHydrationListener.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect } from 'react';
+
+import { useDispatch, useSelector } from 'react-redux';
+
+import {
+  selectisPersistGateHydrationInProgress,
+  setIsPersistGateHydrationComplete,
+  setIsPersistGateHydrationInProgress,
+} from '../redux/slices/persistGateHydration';
```
osamasayed, line 9: Another nit: `@` instead of `..`

src/components/GlobalPersistGateHydrationListener.tsx
tifa2UP: We don't want to clear the timeout in this case. We always want the dispatch action to go through.

src/redux/slices/persistGateHydration.ts
tifa2UP: Good question, my initial implementation involved a single state. 

tl;dr because the redux store is synchronous, you can't tell it do dispatch an action *AFTER* one occurs. You'll need to dispatch it separately after. 

The `isPersistGateHydrationInProgress` flags that the Hydration is in progress/done, which triggers a useEffect that sets `isPersistGateHydrationComplete` to true. 

The application logic will be using the `isPersistGateHydrationComplete`. Let me know if it makes sense.

src/utils/isClient.ts
tifa2UP: yeah, same haha.

osamasayed (approved):

PR #1864, opened 2022-12-29
Make the navigation sidebar closed by default

tifa2UP:
TSIA


sharabash:
I thought we wanted it open by default so people know it's there. After closing it, then we can remember their preference through session state.

osamasayed:
> I thought we wanted it open by default so people know it's there. After closing it, then we can remember their preference through session state.

You are right. This is an outdated issue. Closing it. Jazak Allahu Khairan Nour.

PR #1872, opened 2023-01-11
Migrate to next-i18next

ahmedriad1:
### Summary
We currently use `next-translate` . However, there are some [compatibility issues with next 13](https://github.com/vercel/next.js/issues/43097). This PR migrates to a similar package [`next-i18next`](https://github.com/i18next/next-i18next). 

These are the key differences in terms of usage:
- `const { lang } = useTranslation();` will be `const { locale } = useRouter();`
- The package does not provide a `setLanguage` function since we can use next's native method. So I made a new hook in `@/hooks/useSetLanguage`:
```ts
import { useRouter } from 'next/router';

const useSetLanguage = () => {
  const router = useRouter();

  const setLanguage = (language: string) => {
    const { pathname, asPath, query } = router;
    router.push({ pathname, query }, asPath, { locale: language });
  };

  return setLanguage;
};

export default useSetLanguage;
```
- There is no way of defining a map of pages to namespaces like we did in `i18n.json`. However, we can specify namespaces in each page's `getServerSideProps` or `getStaticProps` by using a custom function I built like this:
```ts
export const getStaticProps: GetStaticProps = async ({ locale }) => {
  // ...

  return {
    props: {
      ...(await loadPageNamespaces(locale, ['home', 'collection'])),
    },
  };
};
```
and I'm still working on migrating all pages to do so


ahmedriad1:
Fixed the hydration error caused by `next-translate` 🎉

Issue #1882, opened 2023-01-15
Storybook link not working 


adelpro:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

First thanks for your effort with this application, and especially for making it free and open source.

The link to Storybook in  the GitHub readme page are not working.

### Steps to reproduce

1. In the reame file of the GitHub repository
2. Click on storybook link ( there is two links in this page), 
3. a 404 page is shown.

### Environment (please complete the following information) and Add any other context about this bug

Os: android
Browser: chrome

### Environment

Staging(next.quran.com)

PR #1893, opened 2023-01-21
docs: fix storybook file issue (#1882)

mrhrifat:
### Summary
There was a `404 Not Found` issue with `storybook.js / .storybook`. This problem was described in (#1882). I don't find any file name with `storybook.js`. 
But ‌‌‌`.storybook` is simillar with it. So I think this is the right file. If **yes**, it will be grace of almighty Allah (SWT).

### Screenshots

| Before | After |
| ------ | ------ |
| image| image |


hazem3500:
@mrhrifat I don't believe the storybook link was intended to point to a specific file 😅 

It was rather intended to point to a hosted [storybook](https://storybook.js.org/)

You can see it locally by running `yarn storybook` 😄 
For some reason it's not hosted properly 🤷

PR #1905, opened 2023-01-26
Use dynamic min-height for reading view container

ahmedriad1:
### Summary

We currently specify a `min-height` for the reading view container to avoid CLS. However `100vh` doesn't seem to fit shorter chapters. This PR takes a dynamic approach by taking the minimum of either `the number of lines * 20vh` or `100vh`.

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |

### DIFF
--- a/src/components/QuranReader/index.tsx
+++ b/src/components/QuranReader/index.tsx
@@ -13,6 +13,7 @@ import Notes from './Notes/Notes';
 import { getObservedVersePayload, getOptions, QURAN_READER_OBSERVER_ID } from './observer';
 import styles from './QuranReader.module.scss';
 import QuranReaderView from './QuranReaderView';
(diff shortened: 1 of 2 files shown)


src/components/QuranReader/index.tsx
```diff
@@ -91,6 +92,16 @@ const QuranReader = ({
     QURAN_READER_OBSERVER_ID,
   );
 
+  const lines = useMemo(
```
osamasayed, line 95: There is an issue with this is that:

1. For `ReadingView`, the resource-intensive function of `groupLinesByVerses` will be executed at least twice. One time here and another inside `src/components/QuranReader/ReadingView/Page.tsx`.
2. For `TranslationView`, this adds extra calculations and won't be used since `QuranReader` component is shared between both views.

Issue #1910, opened 2023-01-29
[bug]: Arabic text not rendered correctly in Tafsir View due to "meQuran" font style used


moedaaboul:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

1- Go to https://quran.com/2:2/tafsirs/en-tafisr-ibn-kathir
2- Select Tafsir ibn Kathir in English
3- Scroll down to extracts from the Quran or Hadith in Arabic
4- You'll notice that there are letters that are not being displayed due to the font "meQuran" being used.

Note: This error is recurring throughout all pages. 

<img width="826" alt="bug -uthmani font meQuran (image 1)" src="https://user-images.githubusercontent.com/77852733/215360123-17aa814b-1290-4b8d-b60d-e1de696101fb.PNG">
<img width="965" alt="bug -uthmani font meQuran (image 2)" src="https://user-images.githubusercontent.com/77852733/215360126-a43f434f-dced-4e96-93eb-7fe88a340d1b.PNG">

### Steps to reproduce

1- Go to https://quran.com/2:2/tafsirs/en-tafisr-ibn-kathir
2- Select Tafsir ibn Kathir in English
3- Scroll down to extracts from the Quran or Hadith in Arabic
4- You'll notice that there are letters that are not being displayed due to the font "meQuran" being used.

Note: This error is recurring throughout all pages. 

### Environment (please complete the following information) and Add any other context about this bug

I've only checked this using Chrome

### Environment

Production(quran.com)

PR #1911, opened 2023-01-29
fixes bug related to Arabic text in Tafsir View

moedaaboul:
### Summary
I have raised an issue ticket that describes the error behaviour in https://github.com/quran/quran.com-frontend-next/issues/1910

### Test Plan
This has minimal effect and corrects an error.

### Screenshots

| Before | After |
| ------ | ------ |
| 
before

 | 
after
 |

### DIFF
--- a/src/components/QuranReader/TafsirView/TafsirView.module.scss
+++ b/src/components/QuranReader/TafsirView/TafsirView.module.scss
@@ -69,7 +69,8 @@ $container-max-width: calc(23 * var(--spacing-mega));
   }
   // target any item which has a class that contains uthmani
   *[class~="uthmani"] {
-    font-family: meQuran;
+    font-family: UthmanicHafs;
+    font-size: larger;
   }
 }
 
--- a/src/components/dls/QuranWord/TextWord.module.scss
+++ b/src/components/dls/QuranWord/TextWord.module.scss
@@ -6,10 +6,6 @@
   font-family: UthmanicHafs !important; // important is needed to over-ride the language specific font-family
 }
 
-.meQuran {
-  font-family: meQuran !important; // important is needed to over-ride the language specific font-family
-}
-
 .IndoPak {
   font-family: IndoPak !important; // important is needed to over-ride the language specific font-family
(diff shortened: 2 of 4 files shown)


hazem3500:
So the `meQuran` font should be deprecated? 🤔

moedaaboul:
Yes, it's not used anywhere in the app, and it is causing all the Quranic text to be corrupted throughout the Ibn Kathir Tafsir translation.  It looks like it was intended to be deleted but was missed out, which can be inferred from this comment: 

```
// TODO: we're not rendering text_uthmani anymore, do we need this font face?
@font-face {
  font-family: "meQuran";
  src: local("me_quran2"),
    url("/fonts/quran/hafs/me_quran/me_quran-2.woff2") format("woff2"),
    url("/fonts/quran/hafs/me_quran/me_quran-2.woff2") format("truetype");
  font-display: swap;
}
```

moedaaboul:
It looks like the font styling is being applied randomly across the pages. You can see from the below that this is the intended styling and now looks fine on the following page:

preview-after

However, the styling is not applied on all the pages consistently as you can see from the below the text is left aligned but is actually meant to be right aligned as it completely misses out both the `arabic` and `uthmani` css rules: 
preview-after-error

I am not sure why there is an inconsistency in the CSS applied to the text, but I suppose it is something to do with how Sass is applied in production though I don't have a clue at this point. **I found this behaviour also on dev whilst debugging locally prior to making any changes, so don't suppose it has to do anything with these changes. Does anyone know whether this will be taken care of once pushed to production?**

hazem3500:
I'm guessing your last point is a separate (yet related) issue? 🤔

Maybe we should write a new issue and tackle it in another PR? 👀

moedaaboul:
Yes, agreed. Let's push these changes then as it looks like Sass rules are applied correctly in production.

osamasayed (approved):

osamasayed:
@moedaaboul Jazak Allah Kahiran for the fix and your contribution!

PR #1948, opened 2023-04-15
Fix tafsir language option update issue

ninearif:
### Summary
I've found that there's an issue that occurred when the user switched the language setting in tafsir view.

If I came from **Arabic > Ibn Kathir** tafsir option selected, then I switch the language to **English** the result will display with 4 tafsirs options, and you can see there're 2 issue: 
1. Tafsir selection has 4 options, 2 of them are duplicated "Ibnu Kathir"
2. The tafsir content still displays in arabic.

I've changed the logic here when the user switches the tafsir language setting
1. Tafsir selection list should only display the tafsirs that matched with newly selected Language.
2. Update the selected tafsir to the first tafsir available in a newly selected language, also reload the content of tafsir.

### Test Plan

1. Go to tafsir view
2. Select tafsir language to **Arabic > Ibnu Kathir** 
3. Change tafsir language to **English**

### Screenshots
BEFORE
image

AFTER
image


src/components/QuranReader/TafsirView/TafsirBody.tsx
```diff
@@ -129,6 +129,40 @@ const TafsirBody = ({
...
+          setSelectedTafsirs({
+            tafsirs: [slug],
+            locale: lang,
+          }),
+          setSelectedTafsirs({
+            tafsirs: tafsirsState.selectedTafsirs,
+            locale: lang,
+          }),
+          PreferenceGroup.TAFSIRS,
+        );
```
ahmedriad1, line 154: Salam alykum Arif,

JAK for your contribution, much appreciated!

Just a comment here. When you change the language, you set the tafsir to the first one in the new language. However, the URL doesn't change to reflect the new tafsir.

You can refactor this function to directly use `onTafsirSelected` and pass the `id` and `slug` of the new tafsir.

PR #1955, opened 2023-04-24
improve performance on areArraysEqual function

MedoAngel:
### Summary
- remove the need to use lodash's isEqual function
- use map instead of sorting and comparing the two arrays which enhances performance from O(5(n+m)) to O(n+m)
- handles some edge cases and exit's early which also increases performance

### Test Plan
this is a fiddle with the test:

https://jsfiddle.net/elussich/4tv9cwjd/

### Screenshots
this is a screenshot showing the performance improvement , almost to 5x

### DIFF
--- a/src/utils/array.ts
+++ b/src/utils/array.ts
@@ -1,5 +1,3 @@
-import isEqual from 'lodash/isEqual';
-
 /**
  * Convert an array of numbers to an array of strings.
  *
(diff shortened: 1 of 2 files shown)


osamasayed:
Alsalamu Alikum, @MedoAngel I think the link to the fiddle is in-correct. It shows this 
Screen Shot 2023-04-24 at 10 36 49 PM

MedoAngel:
> Alsalamu Alikum, @MedoAngel I think the link to the fiddle is in-correct. It shows this Screen Shot 2023-04-24 at 10 36 49 PM

I updated it sorry for the confusion

osamasayed (approved):

PR #1956, opened 2023-04-25
Improve performance for mergeTwoArraysUniquely

MedoAngel:
### Summary
- replace array includes with map look-up, which improves time complexity from O(n*m) to O(n+m)

### Test Plan
this is a fiddle with the test

https://jsfiddle.net/mahmoud_alragabi/v3wL0ob4/

### Screenshots

and here is a screenshot of the performance improvement almost to 3x

image

### DIFF
--- a/src/utils/array.ts
+++ b/src/utils/array.ts
@@ -67,5 +67,10 @@ export const removeItemFromArray = <T>(itemToRemove: T, array: T[]): T[] =>
  * @param {T[]} array2
  * @returns {T[]}
  */
-export const mergeTwoArraysUniquely = <T>(array1: T[], array2: T[]): T[] =>
-  array1.filter((t) => !array2.includes(t)).concat(array2);
+export const mergeTwoArraysUniquely = <T>(array1: T[], array2: T[]): T[] => {
+  const array2Map = new Map();
+
+  array2.map((item) => array2Map.set(item, true));
+
+  return array1.filter((t) => array2Map.get(t) == null).concat(array2);
+};
--- a/src/utils/array.ts
+++ b/src/utils/array.ts
@@ -68,12 +68,5 @@ export const removeItemFromArray = <T>(itemToRemove: T, array: T[]): T[] =>
  * @returns {T[]}
  */
 export const mergeTwoArraysUniquely = <T>(array1: T[], array2: T[]): T[] => {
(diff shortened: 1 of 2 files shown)


ahmedriad1:
Salam alykum @MedoAngel ,

JAK for your contributions!

Just wanted to ask. Is there any reason why you're using a `Map` instead of a `Set`?

The function with a Set would like like this:

```ts
const mergeTwoArraysUniquelySet = (array1, array2) => {
  return Array.from(new Set(array1.concat(array2)));
};
```

ahmedriad1:
cc @osamasayed it looks like we're using `mergeTwoArraysUniquely` only here https://github.com/quran/quran.com-frontend-next/blob/master/src/utils/wordByWord.ts#L36. Should be fine if we returned this order right? `[a,b,c,d,e,f]`

osamasayed:
> cc @osamasayed it looks like we're using `mergeTwoArraysUniquely` only here https://github.com/quran/quran.com-frontend-next/blob/master/src/utils/wordByWord.ts#L36. Should be fine if we returned this order right? `[a,b,c,d,e,f]`

`[a,b,c,d,e,f]` or `[d,e,f,a,b,c]` both are fine as `mergeTwoArraysUniquely`'s usage does not care about the order. Having said this, the implementation of `Set` is simpler than `Map` so I am leaning towards `Set` @MedoAngel

MedoAngel:
@osamasayed I'm leaning towards it as well , should i then update my PR and update the failing tests in wordByWord ? Because they're the only reason that i implemented it this way

osamasayed:
> @osamasayed I'm leaning towards it as well , should i then update my PR and update the failing tests in wordByWord ? Because they're the only reason that i implemented it this way

Yes feel free to update the PR. Jazak Allahu Khairan

MedoAngel:
@osamasayed can you please run the tests using  the set method, when I fix one test others fail for the same reason, the order of the output, I just want to make sure that order doesn't matter before I correct all tests, so please double check

osamasayed:
> @osamasayed can you please run the tests using  the set method, when I fix one test others fail for the same reason, the order of the output, I just want to make sure that order doesn't matter before I correct all tests, so please double check

Yes the order doesn't matter so feel free to update the tests so long as the end result is unique values in an array

osamasayed (approved):

PR #1961, opened 2023-04-29
Add ability to change wbw font size

osamasayed:
### Summary
- We received some [feedback](https://feedback.quran.com/feature-requests/p/font-size-for-word-by-word-translation) regarding having the ability to change the font size of the word-by-word translation/transliteration.
- We will have 6 different font sizes.

### Screenshots

| Before | After |
| ------ | ------ |
|Screen Shot 2023-04-30 at 1 05 59 AM|Screen Shot 2023-04-30 at 1 07 00 AM|
|Screen Shot 2023-04-30 at 1 10 12 AM|Screen Shot 2023-04-30 at 1 10 29 AM|
|Screen Shot 2023-04-30 at 1 12 10 AM|Screen Shot 2023-04-30 at 1 11 53 AM|

### DIFF
--- a/src/components/Navbar/SettingsDrawer/WordByWordSection.tsx
+++ b/src/components/Navbar/SettingsDrawer/WordByWordSection.tsx
@@ -11,6 +11,7 @@ import { shallowEqual, useSelector } from 'react-redux';
 import Section from './Section';
 import styles from './WordByWordSection.module.scss';
 
+import Counter from '@/dls/Counter/Counter';
 import Checkbox from '@/dls/Forms/Checkbox/Checkbox';
 import Select, { SelectSize } from '@/dls/Forms/Select';
 import Link, { LinkVariant } from '@/dls/Link/Link';
@@ -23,6 +24,13 @@ import {
   setWordByWordContentType,
   setWordClickFunctionality,
 } from '@/redux/slices/QuranReader/readingPreferences';
+import {
+  MAXIMUM_WORD_BY_WORD_FONT_STEP,
+  MINIMUM_FONT_STEP,
+  decreaseWordByWordFontScale,
+  increaseWordByWordFontScale,
+  selectWordByWordFontScale,
+} from '@/redux/slices/QuranReader/styles';
 import QueryParam from '@/types/QueryParam';
 import { WordByWordDisplay, WordByWordType, WordClickFunctionality } from '@/types/QuranReader';
 import { removeItemFromArray } from '@/utils/array';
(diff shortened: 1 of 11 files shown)


src/components/Navbar/SettingsDrawer/WordByWordSection.tsx
```diff
@@ -209,6 +251,20 @@ const WordByWordSection = () => {
...
+        <Counter
+          count={wordByWordFontScale}
+          onIncrement={
+            MAXIMUM_WORD_BY_WORD_FONT_STEP === wordByWordFontScale
+              ? null
+              : onFontScaleIncreaseClicked
+          }
+          onDecrement={
+            MINIMUM_FONT_STEP === wordByWordFontScale ? null : onFontScaleDecreaseClicked
+          }
```
ahmedriad1, line 265: Do you think we should put these checks in the callback function instead of doing it here?

src/components/Verse/VerseText.module.scss
```diff
@@ -5,7 +5,8 @@
 
 .verseText {
   display: flex;
-  align-items: center;
+  // flex-start for when the inline wbw translation/transliteration is enabled and its content is wrapped across multiple lines.
```
ahmedriad1, line 8: nit: let's either rename "wbw" to "word by word", or change the "word by word" comment in `InlineWordByWord.module.scss` to "wbw". Let's also put a link for the case you're talking about. Maybe https://quran.com/2:74, right?

src/components/dls/InlineWordByWord/index.tsx
ahmedriad1, line 1: Since this is a DLS component, do you think we should add a story for it?

src/components/dls/InlineWordByWord/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  1: styles.xs,
+  2: styles.sm,
+  3: styles.md,
+  4: styles.lg,
+  5: styles.xl,
+  6: styles.xxl,
+};
+
+type Props = {
+  word?: string;
```
ahmedriad1, line 20: nit: I think we should rename this prop to `text` since it's not a word and can be multiple words.

osamasayed: Agree will update.

src/components/dls/InlineWordByWord/index.tsx
osamasayed: Let's do that later inshAllah. We have quite a few of DLS components that don't have stories.

PR #1980, opened 2023-05-31
Render Quran Reader footer using virtuoso

ahmedriad1:
### Summary

This PR prevents an extra HTTP request that gets triggered when reading to refresh the streak data. Instead of using Virtuoso's Footer, we add `1` to the list count and we render the footer as the last element.

### DIFF
--- a/src/components/QuranReader/ReadingView/index.tsx
+++ b/src/components/QuranReader/ReadingView/index.tsx
@@ -154,19 +154,35 @@ const ReadingView = ({
   useHotkeys('Up', onUpClicked, { enabled: allowKeyboardNavigation }, [scrollToPreviousPage]);
   useHotkeys('Down', onDownClicked, { enabled: allowKeyboardNavigation }, [scrollToNextPage]);
 
-  const itemContentRenderer = (pageIndex: number) => (
-    <PageContainer
-      isUsingDefaultFont={isUsingDefaultFont}
-      pagesVersesRange={pagesVersesRange}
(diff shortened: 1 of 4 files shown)


src/components/QuranReader/ReadingView/index.tsx
```diff
@@ -154,19 +154,35 @@ const ReadingView = ({
...
-      wordByWordLocale={wordByWordLocale}
-      pageIndex={pageIndex}
-      setMushafPageToVersesMap={setMushafPageToVersesMap}
-      initialData={initialData}
-    />
-  );
+  const itemContentRenderer = (pageIndex: number) => {
+    if (pageIndex === pagesCount) {
+      const pageVerses = mushafPageToVersesMap[lastReadPageNumber];
+      const lastVerse = pageVerses?.[pageVerses.length - 1];
```
osamasayed, line 160: nit: `isLastVerse`

src/components/QuranReader/TranslationView/index.tsx
```diff
@@ -105,19 +115,10 @@ const TranslationView = ({
         <Virtuoso
           ref={virtuosoRef}
           useWindowScroll
-          totalCount={initialData.metaData.numberOfVerses}
+          totalCount={initialData.metaData.numberOfVerses + 1}
```
osamasayed, line 118: Why are we adding an extra 1 here?

osamasayed (approved):

Issue #1991, opened 2023-06-23
[bug]: صفة العدد لا تتبع العدد في خانة عدد الآيات


m-adel-ibrahim:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

الأعداد من 3 إلى 10 تقترن باللفظ "آيات"، أما الأعداد أكبر من 10 فتقترن باللفظ "آية".

![image](https://github.com/quran/quran.com-frontend-next/assets/38296884/75c0f681-3216-416e-9255-89226b71c515)

### Steps to reproduce

1. Go to https://quran.com/ar
2. scroll down to see the chapter (Surah) names in Arabic and the number of verses right next to it. 

### Environment (please complete the following information) and Add any other context about this bug

OS: Windows 10
Browser: Google Chrome, version 114.0.5735.134 (Official Build) (64-bit)

### Environment

Production(quran.com)

draqist:
This has been fixed? coz It's displaying correctly on my end

F-Yousri:
<img width="1382" alt="Screenshot 2023-06-27 at 1 52 18 PM" src="https://github.com/quran/quran.com-frontend-next/assets/33264607/54cc292d-4022-4301-baf1-6a476f6bf9ff">
@draqist I don't believe it is

Adaminite:
Forgive me InshAllah, but I don't seem to understand the issue.

Adaminite:
Walaikum assalam. I'm not working on the issue.

ahmedelesawy511:
I've created a pull request with the fix

PR #1992, opened 2023-06-30
Correct the Ayah's format for ayah counts greater than 10 and less than 10

Khalifa1997:
### Summary

This PR deals with https://github.com/quran/quran.com-frontend-next/issues/1991 Issue, I used just the conditional ternary operator here due to the fact that, this issue only exists in two components on the website, I could easily create a function for this if necessary please let me know, -I'd also really appreciate where I would add this function too I initially thought `verse.ts` could be the right place for it so please let me know-

### Test Plan

1. Go to [localhost](http://localhost:3000/ar) scroll down to "سورة" view
2. You will find Surahs with verse count more than 10 to have the value "آية x" while those with verse counts less than 10 to have the value "آيات x"
3. You will find the same behaviour if you choose the "جزء" view

### Screenshots

| Before | After |
| ------ | ------ |
| a| b|
| c |d|


Khalifa1997:
@osamasayed can you please review my PR?

PR #2003, opened 2023-07-22
Log reflection views

osamasayed:
### Summary
This PR adds the following:

1. An intersection observer to detect when a reflection has been totally visible for 3 seconds.
2. When the above condition is met, an API call to QuranReflect is sent to increment reflection views count.
3. On dev environment, a debugger element was added to be able to debug when a reflection is detected as visible.

### DIFF
--- a/src/components/QuranReader/ReflectionView/ReflectionBodyContainer/index.tsx
+++ b/src/components/QuranReader/ReflectionView/ReflectionBodyContainer/index.tsx
@@ -5,9 +5,11 @@ import dynamic from 'next/dynamic';
 import { shallowEqual, useSelector } from 'react-redux';
 
 import DataFetcher from '@/components/DataFetcher';
+import { REFLECTIONS_OBSERVER_ID } from '@/components/QuranReader/observer';
 import TafsirSkeleton from '@/components/QuranReader/TafsirView/TafsirSkeleton';
+import useIntersectionObserver from '@/hooks/useGlobalIntersectionObserverWithDelay';
 import { selectQuranReaderStyles } from '@/redux/slices/QuranReader/styles';
-import { makeAyahReflectionsUrl } from '@/utils/quranReflect/apiPaths';
+import { makeAyahReflectionsUrl, postReflectionViews } from '@/utils/quranReflect/apiPaths';
 import AyahReflectionsResponse from 'types/QuranReflect/AyahReflectionsResponse';
 
 const ReflectionSurahAndAyahSelection = dynamic(() => import('./ReflectionSurahAndAyahSelection'), {
@@ -25,6 +27,8 @@ type ReflectionBodyProps = {
   render: (renderProps: { surahAndAyahSelection: JSX.Element; body: JSX.Element }) => JSX.Element;
 };
 
+const OBSERVER_DELAY_FOR_MS = 3000; // 3 seconds
+
 const ReflectionBodyContainer = ({
   render,
   initialChapterId,
@@ -35,6 +39,19 @@ const ReflectionBodyContainer = ({
   const [selectedVerseNumber, setSelectedVerseNumber] = useState(initialVerseNumber);
   const { translationFontScale } = useSelector(selectQuranReaderStyles, shallowEqual);
   const { lang } = useTranslation();
+
+  const onReflectionViewed = useCallback((reflectionContainer: Element) => {
+    const postId = reflectionContainer.getAttribute('data-post-id');
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    postReflectionViews(postId).catch(() => {});
+  }, []);
+  useIntersectionObserver(
+    { threshold: 1 },
+    onReflectionViewed,
+    REFLECTIONS_OBSERVER_ID,
+    'postId',
+    OBSERVER_DELAY_FOR_MS,
+  );
 
   const renderBody = useCallback(
     (data: AyahReflectionsResponse) => (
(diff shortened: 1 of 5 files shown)


src/components/QuranReader/ReflectionView/ReflectionBodyContainer/index.tsx
```diff
@@ -36,6 +40,19 @@ const ReflectionBodyContainer = ({
   const { translationFontScale } = useSelector(selectQuranReaderStyles, shallowEqual);
   const { lang } = useTranslation();
 
+  const onReflectionViewed = useCallback((reflectionContainer: Element) => {
+    const postId = reflectionContainer.getAttribute('data-post-id');
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    postReflectionViews(postId).catch(() => {});
+  }, []);
```
ahmedriad1, line 47: I think we need to have some sort of caching & batching here because:
1. If you scroll down and go up again, you'll send 2 requests (or more depending on how much you scroll) for the same reflection. 
2. As you scroll, you probably don't see reflections 1 by 1, so we can group reflections viewed at a time and send only 1 request.

src/hooks/useGlobalIntersectionObserverWithDelay.ts
```diff
@@ -0,0 +1,125 @@
...
+  observerId: string,
+  elementIdKey: string,
+  delayForMS: number,
+) => {
+  /**
+   * An object that will hold a reference of each unique identifier and
+   * the ID of the timeout function that will trigger when the element
+   * is visible for delayForMS number of seconds.
+   */
+  const timeouts = useRef<Record<any, number>>({});
```
ahmedriad1, line 48: Let's use a `Map` instead, it's recommended to use maps over objects when you frequently add/delete elements ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps)).

src/hooks/useGlobalIntersectionObserverWithDelay.ts
```diff
@@ -0,0 +1,125 @@
...
+        const targetNode = visibleEntry.target as HTMLElement;
+        const elementId = targetNode.dataset[elementIdKey];
+        // if the element content container is interacting
+        if (visibleEntry.isIntersecting) {
+          setIntersectionDebuggerNodeText(
+            `Element is fully visible, triggering callback in ${delayForMS / 1000} seconds...`,
+            elementIntersectionDebuggerNode,
+          );
+          // Call the API after delayForMS
+          timeouts.current = {
```
ahmedriad1, line 69: we can just do `timeouts.set(key, ...)` after refactoring it to a Map

src/hooks/useGlobalIntersectionObserverWithDelay.ts
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+          // if it's no longer intersecting, clear the timeout
+          clearTimeout(timeouts.current[elementId]);
+          /**
+           * Delete the timeout function's ID from the global array since it's no longer
+           * needed and next time if the same element intersects, a new ID will be generated
+           * for that element.
+           */
+          const tempTimeouts = { ...timeouts.current };
+          delete tempTimeouts[elementId];
```
ahmedriad1, line 88: ditto: `delete` keyword has bad performance. change to `timeouts.delete()` after map refactor.

src/hooks/useGlobalIntersectionObserverWithDelay.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    // no need to create a new observer if it already exists.
+    if (!window[observerId]) {
+      window[observerId] = new IntersectionObserver(updateEntry, {
+        threshold,
+        root,
+        rootMargin,
+      });
+    }
+    return () => {
```
ahmedriad1, line 116: We don't clear timeouts here, causing requests to be sent even if you closed the reflection modal. Loop over all `timeouts`, clear them, and empty the timeouts `Map` after.

ahmedriad1 (changes requested):

src/components/QuranReader/ReflectionView/ReflectionBodyContainer/index.tsx
osamasayed: 1. This is intentional. If you scroll past a reflection and stay long enough for the timeout to trigger then scrolled to it again it and stayed long enough for the timeout to trigger again, we should send two requests.
2. We don't fire the API call instantly but rather wait for a few seconds. If the user scrolls a bit making the post container not 100% visible we will clear the timeout so it will stop. I believe this makes it un-necessary to have queuing in place given.

src/hooks/useGlobalIntersectionObserverWithDelay.ts
osamasayed: Sure will us Map instead.

src/hooks/useGlobalIntersectionObserverWithDelay.ts
osamasayed: Yes you are right my bad let me do that.

osamasayed: I just double-checked. Closing the reflections modal will not trigger the clean-up function since the component `ReflectionBodyContainer` will still be rendered even after you close the modal and closing it will cause all the timeout functions to be cleared anyways since none of the elements be observed will be intersecting. I will add a manual clear timeout thought just to be extra sure.

PR #2012, opened 2023-08-12
Add reading preference switch to navbar

### LINKED ISSUE #1924: [feature]: Make floating buttons for Translation mode and Reading mode

### Summary

As for now, Mode can be switched from start page of Surah/Chapter. In case of user want to see translation of translation of verse, then he has go scroll to start and click on" translation" button
which is so bad experience

### Your purposed solution for this feature

I would Like to see floating buttons to switch between modes

talal-najam:
### Summary
Attempt to fix #1924 . Add reading preference switch to header.

- Show preference toggle in header if the default one disappears
- Remember the verse when switching
- Adjusts sizing for desktop / mobile devices
- Hides Juz / Hizb info on the header on mobile once user has scrolled past the beginning of the page (noticed otherwise the header occupies more than 1/3 of the screen

### Screenshots

| Before | After |
| ------ | ------ |
| Before - desktop | After - desktop |
| Before - mobile | After - mobile |

### DIFF
--- a/src/components/GlobalScrollListener.tsx
+++ b/src/components/GlobalScrollListener.tsx
@@ -4,8 +4,8 @@ import { useDispatch } from 'react-redux';
 
 import useScrollDirection, { ScrollDirection } from '@/hooks/useScrollDirection';
 import { setIsVisible } from '@/redux/slices/navbar';
-import { 
-  setIsExpanded, 
+import {
+  setIsExpanded,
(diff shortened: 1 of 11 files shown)


talal-najam:
I understand it's mentioned this was planned. So if someone is working / has already implemented this, feel free to close.

talal-najam:
Fixed type and linting errors. Could we try to redeploy preview please?

PR #2019, opened 2023-08-20
Feature/2014 Random surah ayah picker

### LINKED ISSUE #2014: [feature]: Randomly choose a surah or ayah for revision

### Summary

Bismillah Al-Rahman Al-Rahim,
As salaam alaykum all.

### Background
For context, I use this website a lot for memorizing and revising the Qur'an. However, I noticed that there is no option to randomly choose a surah or ayah that you've previously read. Maybe I'm looking in the wrong places, but I've tried searching for it in the home page, sidebar, footer, command bar, etc, but there does not seem to be an option for this to my knowledge.

### How will this improve the current experience?
Many muslims use this app for memorization as well, I think it would be a good decision to allow users to solidify their memorization of the Qur'an by having the website randomly choose a surah for them to recite, or an ayah for them to continue from. From personal experience, I've found this way of revision to be quite helpful, and I think others will feel the same.

Furthermore, there may be those who choose to pick a random surah to recite after every salah, or to choose a random ayah to learn the tafsir of. This feature could be beneficial to many people, both for people trying to memorize, and others with different intentions. I think giving them all this feature would be a good idea.

---

I plan on taking a look at the codebase and trying to implement this feature soon insyaAllah. I will attach a PR here once ready. May Allah reward you all greatly for your contributions.

### Where should the user access this feature?

I have a few ideas in mind:

1. Adding a group to the command bar above the existing `navigating-to` group, with each item being one of the following:
 a. `Any surah`
 b. `Any ayah`
 c. `Previously read surah`
 d. `Previously read ayah`
 This would look something like this:
 

2. Adding a button next to the quick links that shows the previous options when clicking on a dropdown. Clicking on the larger button would bring the user to the `/random` page where they can specify which surahs to exclude/include. This button would look like the following:
 

3. A dedicated page (`/random`) to specify which surahs to exclude/include, would mostly be composed of a list and checkboxes next to each item. Each surah would correspond to one item, and the user can also specify until which verse to include with a simple number input. Additionally, a weighting could be given based on a simple rating (memorized, some memorization needed, needs work) to prioritize certain surahs over others. The settings set in this page would be remembered for future visits, with a fallback to the surahs the user has previously read.

### How would this be implemented?

Firstly, implementing the `any random surah/ayah picker` wouldn't be too complicated. My plan for implementing this would be similar to the following:

1. Pick a random surah (number) from the 114 surahs
2. If searching for ayah, look up the number of verses of the chosen surah from `./data/chapters/[locale].json`
3. Navigate to `/[surah]?startingVerse=[ayah]`

As for implementing `any previously read surah/ayah`, I'm thinking of extending the `readingTracker` slice to include previously read surahs as well, since currently it stores both the surahs and the verse. The new version would keep track of which surahs have been opened as a simple list of objects, with each object having properties for the key, last read verse, and last time it was opened.

```json
{
  "key": "mulk",
  "lastRead" : 5,
  "timestamp": 1692160170841
}
```

This way, the random picker would just need to pick a random index from this list, and then pick a random number from 1 to `lastRead`.

The `/random` page would first check if it has previously been set before, if not, it would default back to this list and set the UI to match it. After the user has set their preferences and started the randomization process, the website should remember the options and create a new key under the `readingTracker` called `randomList` which stores this data for future visits.

### Future work

Some additional work could be done in the future to improve this feature:

1. After picking a random list, the user should be able to keep clicking random in the Qur'an reader to continue choosing from that list
2. The user could directly specify which surahs he has trouble with based on a rating, and the website would prioritise these choices more

MFauzanAP:
### Summary

Bismillah Al-Rahman Al-Rahim,
As salaam alaykum all.

I apologise in advance for any inconveniences or if this is not the ideal way of creating PRs, I don't contribute to open source projects very often.

With that being said, This PR aims to add a new feature where the site pick any surah or ayah for the user, including picking from a custom list.

I tried my best to maintain the current code style and design theme, add comments, and reuse existing components whenever I could. However, feel free to let me know if there could be some improvements in this department.

Closes #2014

#### Features
- New random picker UI, located in:
  - Command Bar
    - clicking on an option will also add it to the search history
    - the modal was also made slightly bigger for better visibility of the options
  - Quick Links
  - Search Drawer
- New `/random` page
  - This is used to specify which surahs to include in the randomizer
  - The user can select/deselect surahs to include as well as which ayat to go up to
  - It defaults to the latest surah logs if the user hasn't specified a custom selection
  - The user can search for a surah to include/exclude
  - A page was used instead of a simple modal for better UX and better compatibility for mobile/tablet devices
- New `random.ts` utility file
  - This file allows you to either
    - Pick a random surah id from all surahs
    - Pick a random ayah id from a given surah
    - Pick a random surah and ayah from all surahs
    - The previous options but for a given set of surahs and ayat
  - The file has also been tested with 100% coverage

#### Refactors
- Add `surahLogs` and `customSelection` keys to the `readingTracker` redux slice
  - `surahLogs` is similar to the currently-used `recentReadingSessions` with the difference being it stores the furthest ayat a user has reached instead of where they last left off
  - `customSelection` is used to specify which surahs/ayats the user wants to include in the randomizer if it does not exist, it falls back to `surah logs`
  - `recentReadingSessions` was kept intact to maintain stability and backwards-compatibility

#### New Translation Keys

- `common.json`
  - `random`
    - `pick-random`
    - `any-surah`
    - `any-ayah`
    - `selected-surah`
    - `edit`
- `quick-links.json`
  - `pick-random`
- `random.json`
  - `all`
  - `desc`
  - `load-previous`
  - `reset-changes`
  - `save-randomize`
  - `search`
  - `selected`
  - `title`

### Test Plan

Below are the scenarios I tested for:
- New user with no surah logs or custom selection
- Existing user with some surah logs
- Existing user with both surah logs and custom selection

The new `random.ts` utility file has also been tested with 100% coverage using `random.test.ts`.

### Future Work

As discussed in the original issue, this feature can be improved in the following ways:
- Better UX in `/random`, in particular quick-selecting multiple surahs at once
- Different selections/lists for different purposes for e.g revising a specific juz, memorizing recent surahs, etc.
- After picking a random list, the user should be able to keep clicking random in the Qur'an reader to continue choosing from that list
- The user could directly specify which surahs he has trouble with based on a rating, and the website would prioritise these choices more

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |
| image | image |
| image | image |

image
image


MFauzanAP:
As salaam alaykum.

Would it be worthwhile to add a "Next Random" button in the `ChapterControls` component? This result in better UX as it allows the user to quickly choose a random surah without opening the command bar or search drawer.

image

However, it would mean refactoring some of the current state to keep track of which choice was last picked by the user. Please let me know your thoughts and/or if you have a better solution.

PR #2025, opened 2023-09-13
Fix Quran reader when switching languages

### LINKED ISSUE #1960: [bug]: Arabic not showing when language is changed

### Describe the bug

When switching between languages, the Arabic text disappears. Issue originally reported [here](https://feedback.quran.com/bugs/p/arabic-not-showing-when-change-language).

### Relevant log output

_No response_

### Steps to reproduce

1. Open [Ayatul Kursi](https://quran.com/ayatul-kursi) from Incognito
2. Switch to Bengali
3. Switch back to English
4. Switch to Bengali again
5. You will notice the Arabic text no longer appears.

Apr-28-2023 07-34-27

OS: MacOs Monetery 12.6.4
Browser: Chrome
Version: 112.0.5615.137

### Environment

Production(quran.com)

ahmedriad1:
### Summary

This PR fixes #1960 along with other reported issues on [feedback.quran.com](https://feedback.quran.com/bugs/p/the-quranic-text-appears-as-gibberish).

This issue is caused due to how we detect if the user is using the default settings for their language in order to use the first 9 verses passed by the server (`initialData`). Currently, `isUsingDefaultFont` for an example is a piece of state and not a computed property. This resulted in a wrong value when the user is defaulted to a language (i.e. english), and then changes to another language with different settings (i.e. bengali). 

### Solution

This PR adds a new hook (`useIsUsingDefaultSettings`) that manually checks is default font settings match the current settings.

### DIFF
--- a/src/hooks/useIsUsingDefaultSettings.tsx
+++ b/src/hooks/useIsUsingDefaultSettings.tsx
@@ -15,7 +15,7 @@ import { areArraysEqual } from '@/utils/array';
 import { selectIsUsingDefaultReciter } from '@/xstate/actors/audioPlayer/selectors';
 import { AudioPlayerMachineContext } from '@/xstate/AudioPlayerMachineContext';
 
-const useIsUsingDefaultSettings = ({ translations }: { translations?: number[] }) => {
+const useIsUsingDefaultSettings = ({ translations }: { translations?: number[] } = {}) => {
   const { lang } = useTranslation();
   const audioService = useContext(AudioPlayerMachineContext);
   const isUsingDefaultReciter = useXstateSelector(audioService, (state) =>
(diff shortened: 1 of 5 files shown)


src/hooks/useIsUsingDefaultSettings.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const mushafLines = useSelector(selectQuranMushafLines);
+
+  const defaultState = useMemo(() => {
+    return {
+      quranReaderStyles: getQuranReaderStylesInitialState(lang),
+      translations: getTranslationsInitialState(lang),
+    };
+  }, [lang]);
+
+  const isUsingDefaultFont =
```
osamasayed, line 37: I see that we are no longer using `selectIsUsingDefaultFont`. Is this intentional?

src/hooks/useIsUsingDefaultSettings.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    isUsingDefaultFont &&
+    isUsingDefaultReciter &&
+    isUsingDefaultWordByWordLocale &&
+    isUsingDefaultTranslations &&
+    areTranslationsEqual
+  );
+};
+
+export default useIsUsingDefaultSettings;
```
osamasayed, line 54: I see this is only being used in translation view, what about reading view?

osamasayed (approved):

PR #2033, opened 2023-09-28
Integrate novu

osamasayed:


package.json
```diff
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3005",
```
osamasayed, line 7: We should revert this back.

src/components/Navbar/InAppNotifications/NotificationBell/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  unseenNotificationsCount: number;
+};
+
+const NotificationBell: React.FC<Props> = ({ onBellClicked, unseenNotificationsCount }) => {
+  const { t, lang } = useTranslation('common');
+  let count = null;
+  if (unseenNotificationsCount > 0) {
+    count =
+      unseenNotificationsCount > 9
+        ? `${toLocalizedNumber(9, lang)}+`
```
osamasayed, line 22: `9` should be moved to a const for easier refactor.

src/components/Navbar/InAppNotifications/NotificationBell/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+import useTranslation from 'next-translate/useTranslation';
+
+import styles from './NotificationBell.module.scss';
+
+import Button, { ButtonShape, ButtonVariant } from '@/dls/Button/Button';
+import GlobeIcon from '@/icons/notification-bell.svg';
```
osamasayed, line 8: nit: rename to `NotificationBellIcon`.

src/components/Navbar/InAppNotifications/NotificationsList/NotificationsList.module.scss
```diff
@@ -0,0 +1,37 @@
+.bellContainer {
+  // height: calc(15 * var(--spacing-mega));
```
osamasayed, line 2: nit: this line should be removed

src/components/Navbar/InAppNotifications/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { selectUnseenCount } from '@/redux/slices/notifications';
+
+const InAppNotifications = () => {
+  const [showModal, setShowModal] = useState(false);
+
+  const { fetchNotifications } = useNotifications();
+  const unseenCount = useSelector(selectUnseenCount);
+
+  const onBellClicked = async () => {
+    setShowModal((prevShowModal) => !prevShowModal);
```
osamasayed, line 20: Add logging

src/notifications/queryKeys.ts
```diff
@@ -0,0 +1,6 @@
+enum NotificationQueryKey {
```
osamasayed, line 1: This does not seem to be used.

src/notifications/useUnseenCount.ts
```diff
@@ -0,0 +1,39 @@
...
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (status === HeadlessServiceStatus.READY) {
+      headlessService.fetchUnseenCount({
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        listener: () => {},
+        onSuccess: ({ count }) => {
+          dispatch({ type: setUnseenCount.type, payload: count });
+          setError(null);
```
osamasayed, line 21: Do we need to explicitly set the error to `null` since it's the default status?

src/notifications/useUnseenCount.ts
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
```
osamasayed, line 1: nit: rename file name to match the name of the hook `useInitializeUnseenCount`.

src/notifications/useUnseenCount.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { setUnseenCount } from '@/redux/slices/notifications';
+
+const useInitializeUnseenCount = () => {
+  const { headlessService, status } = useHeadlessService();
+  const [error, setError] = useState<unknown>(null);
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (status === HeadlessServiceStatus.READY) {
```
osamasayed, line 15: shouldn't we use `isReady` flag instead?

src/components/Navbar/InAppNotifications/NotificationsList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const isLoading = useSelector(selectNotificationsIsLoading);
+  const isFetching = useSelector(selectNotificationsIsFetching);
+
+  const onMarkAllAsReadClicked = () => {
+    logButtonClick('notification_mark_all_as_read');
+    markAllAsRead.mutate();
+  };
+
+  const loadMoreNotifications = () => {
+    console.log({ hasMoreNotifications, lastPage });
```
osamasayed, line 40: We should remove this.

src/components/Navbar/InAppNotifications/NotificationsList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            {t('notification.mark-all-as-read')}
+          </Button>
+        </div>
+      </div>
+      <Separator />
+      <div>
+        {notifications.length > 0 ? (
+          <Virtuoso
+            data={notifications}
+            overscan={100}
```
osamasayed, line 69: I believe this value is too big we should reduce it to 10 maybe

src/components/Navbar/InAppNotifications/NotificationsList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            className={styles.notificationsList}
+            endReached={loadMoreNotifications}
+            itemContent={(index, notification) => (
```
osamasayed, line 72: nit: move this to a const instead of inline function

src/notifications/useHeadlessService.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    headlessService.initializeSession({
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      listener: () => {},
+      onSuccess: () => {
+        setStatus(HeadlessServiceStatus.READY);
+      },
+      onError: (error) => {
+        setStatus(HeadlessServiceStatus.ERROR);
+        console.log('headless error:', error);
```
osamasayed, line 41: remove this

src/components/Navbar/InAppNotifications/NotificationsList/NotificationItem/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const { markNotificationAsRead, deleteNotification } = useNotifications();
+
+  const onMarkNotificationAsReadClicked = (notificationId: string) => {
+    logButtonClick('notification_mark_as_read', { content: notification.content, notificationId });
+    markNotificationAsRead.mutate(notificationId);
+    setIsPopoverOpen(false);
+  };
+
+  const onDeletedNotificationClicked = (notificationId: string) => {
+    logButtonClick('notification_delete', { content: notification.content, notificationId });
```
osamasayed, line 35: ditto: I don't think we should be passing `content` in the event.

src/components/Navbar/InAppNotifications/NotificationsList/NotificationItem/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    deleteNotification.mutate(notificationId);
+  };
+
+  const onOpenChange = (open: boolean) => {
+    logButtonClick('notification_more', {
+      content: notification.content,
```
osamasayed, line 41: ditto: I don't think we should be passing content in the event.

src/notifications/useFetchNotifications.ts
```diff
@@ -0,0 +1,83 @@
...
+  const [error, setError] = useState<unknown>(null);
+  const dispatch = useDispatch();
+
+  const { mutate: markNotificationAsSeen } = useMarkNotificationAsSeen();
+
+  const fetchNotifications = useCallback(
+    // eslint-disable-next-line react-func/max-lines-per-function
+    ({
+      shouldMarkAsSeenOnSuccess = true,
+      page: forcedPage,
```
osamasayed, line 26: nit: `page` instead of `forcedPage`.

src/notifications/useFetchNotifications.ts
```diff
@@ -0,0 +1,83 @@
...
+    }: {
+      shouldMarkAsSeenOnSuccess?: boolean;
+      shouldResetOldData?: boolean;
+      page: number;
+    }) => {
+      if (isReady) {
+        headlessService.fetchNotifications({
+          page: forcedPage, // page number to be fetched
+          query: {
+            limit: 10,
```
osamasayed, line 37: nit: let's move `10` to a const

src/redux/slices/notifications.ts
```diff
@@ -0,0 +1,205 @@
...
+      }>,
+    ) => ({
+      ...state,
+      isLoadingNotifications: action.payload.isLoading,
+      isFetchingNotifications: action.payload.isFetching,
+      ...(action.payload.shouldResetOldData
+        ? { notifications: [], paginatedNotifications: {} }
+        : {}),
+    }),
+    setNotificationsPageAndFinishLoading: (
```
osamasayed, line 29: We should add a comment inside regarding what we are trying to achieve

src/redux/slices/notifications.ts
```diff
@@ -0,0 +1,205 @@
...
+        notifications: newNotifications,
+        isLoadingNotifications: false,
+        isFetchingNotifications: false,
+      };
+    },
+    setUnseenCount: (state: NotificationsState, action: PayloadAction<number>) => ({
+      ...state,
+      unseenCount: action.payload,
+    }),
+    setAllAsSeen: (state: NotificationsState) => {
```
osamasayed, line 58: I see that we are not using `setAllAsSeen` as of now. We should remove it.

src/redux/slices/notifications.ts
```diff
@@ -0,0 +1,205 @@
...
+      };
+    },
+    setDeleteNotification: (
+      state: NotificationsState,
+      action: PayloadAction<{ messageId: string }>,
+    ) => {
+      const newPagedNotifications = {};
+      const newNotifications = [];
+      Object.keys(state.paginatedNotifications).forEach((pageKey) => {
+        const old = {
```
osamasayed, line 153: I noticed the bits 
```
const old = {
          ...state.paginatedNotifications[pageKey],
          data: [...state.paginatedNotifications[pageKey].data],
        };
```
 and 

```
newPagedNotifications[pageKey] = {
          ...old,
          data: newData,
        };
        newNotifications.push(...newData);
```
Are repeated multiple times but the middle part changes, is there a better way to keep the code DRY?

src/notifications/useDeleteNotification.ts
```diff
@@ -0,0 +1,47 @@
...
+
+            if (isError) {
+              setError(err);
+            }
+          },
+          onSuccess: () => {
+            setError(null);
+            dispatch({ type: setDeleteNotification.type, payload: { messageId } });
+          },
+          onError: () => {
```
osamasayed, line 35: Report error to Sentry

src/notifications/useFetchNotifications.ts
```diff
@@ -0,0 +1,83 @@
...
+                payload: {
+                  isLoading: loading,
+                  isFetching: fetching,
+                  shouldResetOldData,
+                },
+              });
+            }
+
+            if (isError) {
+              setError(err);
```
osamasayed, line 61: Report error to Sentry either from here or by adding `onError` handler which is what i prefer

src/notifications/useHeadlessService.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, []);
+
+  useEffect(() => {
+    headlessService.initializeSession({
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      listener: () => {},
+      onSuccess: () => {
+        setStatus(HeadlessServiceStatus.READY);
+      },
+      onError: (error) => {
```
osamasayed, line 39: Report error to Sentry

src/notifications/useMarkAllAsRead.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+      },
+      onSuccess: () => {
+        setError(null);
+
+        // manually set all notifications to read instead of calling the API again
+        dispatch({ type: setAllAsRead.type });
+      },
+      onError: () => {
+        toast(t('error.general'), { status: ToastStatus.Error });
```
osamasayed, line 36: Report error to Sentry

src/notifications/useMarkNotificationAsRead.ts
```diff
@@ -0,0 +1,49 @@
...
+              setError(err);
+            }
+          },
+          onSuccess: () => {
+            setError(null);
+
+            // manually update the read state of the notification instead of calling the API again
+            dispatch({ type: setNotificationAsRead.type, payload: { messageId } });
+          },
+          onError: () => {
```
osamasayed, line 37: Report error to Sentry

src/notifications/useMarkNotificationAsSeen.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const markNotificationsAsSeen = useCallback(
+    (messageIds: IMessageId) => {
+      if (isReady) {
+        headlessService.markNotificationsAsSeen({
+          messageId: messageIds,
+          listener: ({ isLoading, isError, error: err }) => {
+            setIsMutating(isLoading);
+
+            if (isError) {
```
osamasayed, line 20: Report error to Sentry either from here or from inside `onError` which is what I prefer

src/notifications/useUnseenCount.ts
```diff
@@ -0,0 +1,39 @@
...
+      headlessService.fetchUnseenCount({
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        listener: () => {},
+        onSuccess: ({ count }) => {
+          dispatch({ type: setUnseenCount.type, payload: count });
+          setError(null);
+        },
+        onError: (err: unknown) => {
```
osamasayed, line 23: Report error to Sentry

src/notifications/useUnseenCount.ts
ahmedriad1: No, but if there was an error from a previous request, we need to reset it it.

src/notifications/useHeadlessService.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  useEffect(() => {
+    headlessService.initializeSession({
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      listener: () => {},
+      onSuccess: () => {
+        setStatus(HeadlessServiceStatus.READY);
+      },
+      onError: (err) => {
+        toast(t('error.general'), { status: ToastStatus.Error });
```
osamasayed, line 46: We should not show a toast since it's not user initiated action.

src/notifications/useFetchNotifications.ts
```diff
@@ -0,0 +1,101 @@
...
+  setNotificationsPageAndFinishLoading,
+} from '@/redux/slices/notifications';
+import { areArraysEqual } from '@/utils/array';
+
+const NOTIFICATIONS_PAGE_SIZE = 10;
+
+const useFetchNotifications = () => {
+  const { headlessService, isReady } = useHeadlessService();
+
+  // const [pageNumber, setPageNumber] = useState(0);
```
osamasayed, line 21: nit: remove this

src/notifications/useFetchNotifications.ts
```diff
@@ -0,0 +1,101 @@
...
+  selectLoadedNotificationsPages,
+  selectNotificationsIsFetching,
+  setNotificationsLoading,
+  setNotificationsPageAndFinishLoading,
+} from '@/redux/slices/notifications';
+import { areArraysEqual } from '@/utils/array';
+
+const NOTIFICATIONS_PAGE_SIZE = 10;
+
+const useFetchNotifications = () => {
```
osamasayed, line 18: As discussed, something causes an infinite loop of calling `fetch`

ahmedriad1: Turns out it was `useHeadlessService` that was infinitely initializing. It was fixed.

sentry.client.config.js
```diff
@@ -5,17 +5,13 @@
...
-  // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 1.0,
-  // ...
-  // Note: if you want to override the automatic release value, do not set a
-  // `release` value here - use the environment variable `SENTRY_RELEASE`, so
-  // that it will also get attached to your source maps
+  enabled: SENTRY_ENABLED,
+  dsn: SENTRY_ENABLED ? SENTRY_DSN : null,
+  debug: isDev,
+  defaultIntegrations: false,
```
osamasayed, line 15: We no longer need to set `tracesSampleRate`?

src/hooks/auth/useGetUserFeatureFlags.ts
```diff
@@ -0,0 +1,20 @@
...
+    isLoggedIn() ? makeUserFeatureFlagsUrl() : null,
+    getUserFeatureFlags,
+  );
+
+  return {
+    data,
+    isLoading: isValidating && !data,
+    error,
+  };
+};
```
ahmedriad1, line 18: I think we can refactor this hook to make it easier to use. For an example, when we use, it check for the feature flag everytime manually. What about accepting the flag in the hook, just return `isEnabled` (will be `false` if not logged in) and `isLoading`, and handle the error as well since it's a passive request so I think the handling will always be the same (whether to just ignore it or maybe show a toast).

So the final result would look like:
```ts
const useGetUserFeatureFlags = (flag: string) => {
  const { data, isValidating, error } = useSWR(
    isLoggedIn() ? makeUserFeatureFlagsUrl() : null,
    getUserFeatureFlags,
  );


  const isLoading = isValidating && !data;
  return {
    isEnabled: isLoggedIn() && !isLoading && !error && data?.[flag],
    isLoading,
  };
};
```

We also shouldn't worry about using the hook many times in different places since SWR will handle deduping them, though I think we might need `useSWRImmutable`.

src/components/Navbar/NavbarBody/index.tsx
```diff
@@ -54,6 +56,9 @@ const NavbarBody: React.FC = () => {
     dispatch({ type: setIsSettingsDrawerOpen.type, payload: true });
   };
 
+  const inAppNotificationsAllowed =
+    isLoggedIn() && !isLoading && !error && featureFlags?.inAppNotifications;
+
```
ahmedriad1, line 61: Refer the to comment on `src/hooks/auth/useGetUserFeatureFlags.ts`

ahmedriad1 (changes requested):

src/hooks/auth/useGetUserFeatureFlags.ts
osamasayed: Yes I like the refactor suggestion it's much more reusabled. Just bare in mind that this is temporary until we deploy Unleash :) I will refactor the code. Regarding using `useSWRImmutable`, I think `useSWR` is better incase we have online users and we happen to change their feature flags and we want to reflect the changes as soon as possible (disaster plan for when Novu deployment goes wrong :) )

PR #2015, opened 2023-08-17
 Fixed typo Al-Noor to An-Noor (Alif lam syamsiyah)

muhamiyan:
### Screenshots

| Before | After |
| ------ | ------ |
| image| image |

PR #2035, opened 2023-10-01
Changed Al-noor to An-noor in the footer.

farhanahmad15:
### Summary
In the description of the footer, it should be An-noor instead of Al-noor


muhamiyan:
Thanks @farhanahmad15, but this has been fixed in #2015. @osamasayed please review it.

PR #2037, opened 2023-10-23
Add user consent flow

osamasayed:


src/components/Auth/Consents/CommunicationConsent/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  isLoading: boolean;
+  onCompleted: (consentType: ConsentType, consented: boolean) => void;
+};
+
+const CommunicationConsent: React.FC<Props> = ({ onCompleted, consentType, isLoading }) => {
+  const onButtonClicked = (consented: boolean) => {
+    logButtonClick(`${consentType}_${consentType}`);
+    onCompleted(consentType, consented);
+  };
+
```
ahmedriad1, line 23: I see we're trying to make this dynamic and work for all consent types (which is amazing), however some stuff still are hard-coded. 

1. We need to rename the component to something like `ConsentModal`
2. We need to dynamically get the `i18nKey` using the `consentType` instead of hard-coding `communication`

src/utils/auth/api.ts
```diff
@@ -134,6 +136,13 @@ export const completeAnnouncement = async (data: CompleteAnnouncementRequest): P
   return postRequest(makeCompleteAnnouncementUrl(), data);
 };
 
+export const userConsent = async (data: {
```
ahmedriad1, line 139: nit: rename to `updateUserConsent`

ahmedriad1 (approved):

PR #2047, opened 2023-11-26
Made Al-Nur spelling consistent (Al-Noor => An-Nur)

DriftingWhiteMist:
### Summary
In the paragraph imaged below An-Nur was incorrectly spelled as Al-Noor. There were multiplie instances of this in different languages, they have also been replaced.

### Screenshots

| Before |
Screenshot 2023-11-26 at 4 11 03 PM
| After |
Screenshot 2023-11-26 at 4 19 52 PM

### DIFF
--- a/locales/en/common.json
+++ b/locales/en/common.json
@@ -172,7 +172,7 @@
     "text_uthmani": "Uthmani"
   },
   "footer": {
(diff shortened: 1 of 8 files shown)


DriftingWhiteMist:
I did not see that this was a duplicate, the pull request is now closed.

PR #2059, opened 2023-12-17
Add courses

osamasayed:


src/utils/mockoon/api.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable max-lines */
```
osamasayed, line 1: TODO: remove this

src/utils/mockoon/apiPath.ts
```diff
@@ -0,0 +1,15 @@
+import stringify from '../qs-stringify';
```
osamasayed, line 1: TODO: remove this

src/utils/navigation.ts
```diff
@@ -237,6 +237,23 @@ export const getSurahInfoNavigationUrl = (chapterIdOrSlug: string): string =>
...
+export const getCourseNavigationUrl = (courseSlug: string): string => `/courses/${courseSlug}`;
+
+/**
+ * Get href link to the lesson page
+ *
+ * @param {string} courseSlug
+ * @returns {string} lessonPageUrl
+ */
+export const getLessonNavigationUrl = (courseSlug: string, lessonSlug: string): string =>
+  `/courses/${courseSlug}/lessons/${lessonSlug}`;
```
ahmedriad1, line 255: nit: just a personal preference, but I prefer to remove `/lessons` so that the url is: `/courses/${courseSlug}/${lessonSlug}`

src/hooks/useMutateWithoutRevalidation.ts
ahmedriad1, line 1: Is this hook worth it? I see it's just an abstraction over `{ revalidate: false }` which is not a big deal

src/components/dls/TabsNew/Tabs.tsx
ahmedriad1, line 1: We don't use this component anymore. I think we should remove it and refactor the `Tabs` component to use Radix in a separate PR

src/components/dls/MultiStepProgressBar/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+              key={id}
+              className={classNames(styles.circle, {
+                [styles.active]: index === activeIndex,
+              })}
+              role="button"
+              tabIndex={0}
+              onClick={onClicked}
+              onKeyDown={onClicked}
+            >
+              {isCompleted ? '✓' : index + 1}
```
ahmedriad1, line 56: instead of hiding the lesson number and showing a checkmark, I think we can highlight the circle in the active color and leave the number. This is better for UX so that the user doesn't have to count the circles manually

src/components/HomePage/PlayRadioButton.tsx
```diff
@@ -55,6 +60,16 @@ const PlayRadioButton = () => {
...
+      <Button
+        href={getCoursesNavigationUrl()}
+        variant={ButtonVariant.Outlined}
+        className={styles.ctaButton}
+        size={ButtonSize.Small}
+        type={ButtonType.Secondary}
+        onClick={onCoursesClicked}
+      >
+        {t('common:courses')}
+      </Button>
```
ahmedriad1, line 72: I don't think this is the best place for the courses button. I still don't know what would the best place be. But what I have in mind is to add it in this section:

CleanShot 2024-01-16 at 19 41 52@2x

But maybe with a different color

src/components/Course/LessonView/CourseMaterial/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            onClick={onClickHandler}
+            onKeyDown={onClickHandler}
+            key={id}
+            className={classNames(styles.lessonContainer, {
+              [styles.currentLesson]: id === currentLessonId,
+            })}
+          >
+            <div>
+              <span className={styles.day}>{`${
+                isCompleted ? '✓' : toLocalizedNumber(dayNumber, lang)
```
ahmedriad1, line 66: ditto: use an SVG

src/components/Course/CourseDetails/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        </div>
+        <StatusHeader course={course} />
+      </div>
+
+      <div className={styles.imgContainer}>
+        <Image alt={title} src={image} layout="fill" />
+      </div>
+
+      <Switch selected={selectedTab} items={tabs} onSelect={onTabChange} />
+      {selectedTab === Tab.MAIN ? <MainDetails course={course} /> : <Syllabus course={course} />}
```
ahmedriad1, line 81: A cleaner way to do this is to add it to the tabs array (or might be better to change it into an object in this case), and dynamically get the active tab.

src/components/Course/CourseDetails/Tabs/Syllabus/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+                href={
+                  isUserLoggedIn
+                    ? getLessonNavigationUrl(courseSlug, slug)
+                    : getLoginNavigationUrl()
+                }
+                variant={LinkVariant.Highlight}
+              >
+                {title}
+              </Link>
+              {isCompleted ? '✓' : ''}
```
ahmedriad1, line 58: ditto: use an SVG

ahmedriad1 (changes requested):

next.config.js
```diff
@@ -20,7 +20,15 @@ const config = {
...
-    domains: ['cdn.qurancdn.com', 'static.qurancdn.com', 'vercel.com', 'now.sh', 'quran.com'],
+    domains: [
+      'placehold.co',
+      'cdn.qurancdn.com',
+      'static.qurancdn.com',
+      'vercel.com',
+      'now.sh',
+      'quran.com',
+      'images.quran.com',
+    ],
```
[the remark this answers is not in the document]
osamasayed: We are hosting them on R2 under the domain, `images.quran.com` so I added it here but forgot to remove `placehold.co`, will do.

src/components/HomePage/PlayRadioButton.tsx
osamasayed: This was agreed on with Osama until we have a UX resource and we would re-design our nav bar to include it.

src/components/Course/CoursesList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+              href={getCourseNavigationUrl(course.slug)}
+            >
+              <Card
+                shouldShowFullTitle
+                imgSrc={course.image}
+                key={course.id}
+                title={`${course.title} ${course.isCompleted ? '✓' : ''}`}
+                imgAlt={course.title}
+                size={CardSize.Medium}
+                actionIcon={<SearchIcon />}
```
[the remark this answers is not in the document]
osamasayed: There is a nice hover effect that happens when we include an icon and would be removed if we remove the icon.
Jan-17-2024 10-22-53 am

src/utils/navigation.ts
osamasayed: I don't have a very strong opinion regarding this and each option has its pros and cons but in general I am leaning towards having `lessons` in the URL:

**Route 1: /courses/{courseSlug}/lessons/{lessonSlug}**

**Pros:**

* **Clearer hierarchy:** Explicitly specifies that the resource is a lesson within a course, aiding crawlers and users in understanding the URL structure.
* **Potential for semantic value:** If the course slug directly relates to the topic of the lesson, it can boost keyword targeting.
* **More flexible for future expansions:** Allows you to easily add additional subsections within lessons without changing the main URL structure.

**Cons:**

* **Longer URL:** May appear overwhelming to some users.

**Route 2: /courses/{courseSlug}/{lessonSlug}**

**Pros:**

* **Shorter and more concise:** Easier to remember and share.
* **Potentially cleaner aesthetic:** Looks less cluttered and can appear more user-friendly.

**Cons:**

* **Less explicit hierarchy:** The relationship between the course and lesson might be less apparent to crawlers, requiring additional internal linking or schema markup.
* **Less flexibility for future expansions:** Introducing sub-resources within lessons might require a URL revamp later.

src/hooks/useMutateWithoutRevalidation.ts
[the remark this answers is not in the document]
osamasayed: Yes I agree it does not add too much value except saving from retyping `{ revalidate: false }` multiple times, so you suggest we should remove it or keep it?

src/components/DataFetcher/index.tsx
```diff
@@ -58,6 +60,14 @@ const DataFetcher: React.FC<Props> = ({
    * or if we had an error when calling the API.
    */
   if (error) {
+    // if there is a custom error renderer, use it.
+    if (errorRenderer) {
+      const errorComponent = errorRenderer(error);
+      // if the custom error renderer returns false, it means that it doesn't want to render anything special.
+      if (errorComponent !== false) {
+        return errorComponent;
+      }
```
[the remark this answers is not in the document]
osamasayed: The use case I have is I want to have a custom `errorRenderer` that checks for a specific case only and return a custom error, otherwise, just render the default `return <Error onRetryClicked={onRetryClicked} error={error} />;` bit. This is how it looks like:

```
const renderError = (error: any) => {
    if (error?.message === ApiErrorMessage.CourseNotEnrolled) {
      return (
        <div className={styles.container}>
          <PageContainer>
            <Trans
              i18nKey="learn:not-enrolled"
              components={{
                link: (
                  <Link
                    onClick={onUnEnrolledNavigationLinkClicked}
                    key={0}
                    href={getCourseNavigationUrl(slug as string)}
                    variant={LinkVariant.Blend}
                  />
                ),
              }}
            />
          </PageContainer>
        </div>
      );
    }
    return false;
  };
  ```

File: src/pages/courses/[slug]/lessons/[lessonSlugOrId]/index.tsx [hunk 12]
```diff
@@ -0,0 +1,109 @@
...
+    if (lesson) {
+      return (
+        <>
+          <NextSeoWrapper
+            title={lesson.title}
+            url={getCanonicalUrl(
+              lang,
+              getLessonNavigationUrl(slug as string, lessonSlugOrId as string),
+            )}
+          />
```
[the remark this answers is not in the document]
Comment (osamasayed) [reply src/pages/courses/[slug]/lessons/[lessonSlugOrId]/index.tsx:84]: I agree, will add a followup task for SEO

PR #2075, opened 2024-01-25
User Onboarding

ahmedriad1:
### Summary

This PR adds the user onboarding feature!

### Screenshots

State: First visit

---

State: After dismiss

---

State: Onboarding active


src/components/dls/Button/Button.tsx
```diff
@@ -29,6 +29,7 @@ export enum ButtonType {
   Success = 'success',
   Error = 'error',
   Warning = 'warning',
+  Inverse = 'inverse',
```
osamasayed, line 32: Can we add an example to Storybook?

src/components/Onboarding/OnboardingChecklist/OnboardingChecklist.module.scss
```diff
@@ -0,0 +1,90 @@
+@use "src/styles/breakpoints";
+
+.checklistPosition {
+  position: fixed;
+  bottom: 20px;
+  inset-inline-end: 20px;
```
osamasayed, line 6: we should use existing `spacing-{...}` css variables.

src/components/Onboarding/OnboardingChecklist/OnboardingChecklist.module.scss
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.checklist {
+  background: var(--color-background-alternative-faded);
+  border: 1px solid var(--color-borders-hairline);
+  box-shadow: var(--shadow-small);
+  border-radius: var(--border-radius-rounded);
+  padding: var(--spacing-large) var(--spacing-large);
+  max-width: 100%;
+  width: 380px;
```
osamasayed, line 18: we should use existing `spacing-{...}` css variables.

src/components/Onboarding/OnboardingStep/OnboardingStep.module.scss
```diff
@@ -0,0 +1,31 @@
+.tooltipContainer {
+  border-radius: var(--border-radius-default);
+  padding: var(--spacing-small);
+  color: var(--color-text-inverse);
+  background-color: var(--color-background-inverse);
+  max-width: 100%;
+  width: 340px;
```
osamasayed, line 7: we should use existing `spacing-{...}` css variables.

src/components/Onboarding/OnboardingStep/OnboardingStep.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  z-index: 1100;
```
osamasayed, line 8: we should use existing `var(--z-index-{...}` css variables.

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+    spotlightPadding: 10,
+  };
+
+  const getDetails = (key: string) => ({
+    title: t(`onboarding:steps.${key}.title`),
+    description: t(`onboarding:steps.${key}.description`),
+  });
+
+  return {
+    [OnboardingGroup.HOMEPAGE]: [
```
osamasayed, line 47: not a very strong opinion but I think we should break this object into either multiple ones OR multiple files for easier readability and future refactoring.

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+  },
+  {
+    group: OnboardingGroup.READING_EXPERIENCE,
+    title: t('onboarding:checklist.3'),
+    href: getSurahNavigationUrl(1),
+  },
+  {
+    group: OnboardingGroup.PERSONALIZED_FEATURES,
+    title: t('onboarding:checklist.4'),
+    href: '/login',
```
osamasayed, line 259: nit: we have `getLoginNavigationUrl()`

adaheeye:
The skip button should probably be in white text since the background is black.

image

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+        ...getDetails('translations'),
+        step: {
+          ...commonStepOptions,
+          target: '#translation-section',
+          spotlightClicks: true,
+          placement: 'left-start',
+        },
+      },
+    ],
+    [OnboardingGroup.READING_EXPERIENCE]: [
```
osamasayed, line 170: There is an issue when you turn this step on while you are on the reading view:

Jan-26-2024 12-49-21 am

src/components/Onboarding/steps.ts
osamasayed, line 259: This should be conditionally done. It still redirects me to login page even though I am logged in.

src/components/Navbar/SettingsDrawer/Section/Row.tsx
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 
 import styles from './Row.module.scss';
 
-const Row = ({ children }) => <div className={styles.row}>{children}</div>;
+interface RowProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+const Row = (props: RowProps) => <div className={styles.row} {...props} />;
```
osamasayed, line 7: I prefer to explicitly show which props we are passing exactly since it's very vague looking at the component and we have to go check the usage to know what was passed etc.

src/components/Navbar/SettingsDrawer/Section/index.tsx
```diff
@@ -6,8 +6,10 @@ import Title from './Title';
 
 import Separator, { SeparatorWeight } from '@/dls/Separator/Separator';
 
-const Section = ({ children }) => (
-  <div className={styles.section}>
+interface SectionProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+const Section = ({ children, ...props }: SectionProps) => (
```
osamasayed, line 11: Ditto: I prefer to explicitly show which props we are passing exactly since it's very vague looking at the component and we have to go check the usage to know what was passed etc.

src/components/Navbar/SettingsDrawer/SettingsBody.tsx
```diff
@@ -1,19 +1,30 @@
...
-);
+import { useOnboarding } from '@/components/Onboarding/OnboardingProvider';
+
+const SettingsBody = () => {
+  const { isActive, nextStep } = useOnboarding();
+
+  useEffect(() => {
+    if (isActive) setTimeout(() => nextStep(), 400); // wait for transition to finish (.4s)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
osamasayed, line 17: We need to clear the timeout on un-mount. For reference: check `src/components/Verse/OverflowVerseActionsMenuBody.tsx`

osamasayed: Also, where is the 0.4s coming from? CSS?

src/redux/slices/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+    // checklist state
+    dismissChecklist: (
+      state: OnboardingState,
+      action: PayloadAction<{ countDismiss: boolean }>,
+    ) => ({
+      ...state,
+      isChecklistVisible: false,
+      ...(action.payload.countDismiss && {
+        checklistDismissals: state.checklistDismissals + 1,
+        lastChecklistDismissal: new Date(),
```
osamasayed, line 48: Clicking on next/prev buttons show this error:

Screenshot 2024-01-26 at 1 03 27 am

osamasayed:
There is an issue on mobile during the settings flow that prevents scrolling and exiting https://github.com/quran/quran.com-frontend-next/assets/15169499/cd0878db-aee2-4497-ae4f-14953b5c1413

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+  const commonStepOptions: Partial<Step> = {
+    content: '',
+    placement: 'auto',
+    disableBeacon: true,
+    hideBackButton: true,
+    disableOverlay: false,
+    disableOverlayClose: true,
+    tooltipComponent: component as any,
+    floaterProps: {
+      hideArrow: true,
```
osamasayed, line 36: I think we should set this to `hideArrow: false` as it's better UX

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+        step: {
+          ...commonStepOptions,
+          target: '#theme-section',
+          spotlightClicks: true,
+          placement: 'left-start',
+          disableScrolling: true,
+        },
+      },
+      {
+        ...getDetails('font-style'),
```
osamasayed, line 102: There is an issue on mobile during the settings flow that prevents scrolling and exiting


https://github.com/quran/quran.com-frontend-next/assets/15169499/427e97e2-32a7-49a6-88c4-44dab4215a2c

osamasayed (changes requested):

src/components/Onboarding/steps.ts
```diff
@@ -0,0 +1,261 @@
...
+      },
+      {
+        ...getDetails('3-dot-menu'),
+        step: {
+          ...commonStepOptions,
+          target: '.overflow-verse-actions-menu-trigger',
+        },
+      },
+      {
+        ...getDetails('reading-view'),
```
osamasayed, line 220: There is an issue if the user scrolls while flow is one:

src/components/Onboarding/OnboardingStep/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <p className={styles.description}>{stepData.description}</p>
+
+      {(showSkipButton || showNextButton) && (
+        <div className={styles.actionContainer}>
+          {isFirstStep && showSkipButton && (
+            <Button
+              {...skipProps}
+              onClick={handleSkipClick}
+              variant={ButtonVariant.Ghost}
+              type={ButtonType.Inverse}
```
osamasayed, line 97: This style is causing an issue with UI where the color of the text is not readable
Screenshot 2024-01-26 at 1 29 55 am

src/components/Onboarding/OnboardingProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+            index: 0,
+          }),
+        );
+      }, 0);
+    },
+    [dispatch],
+  );
+
+  const stopTour = useCallback(() => {
+    setTimeout(() => {
```
osamasayed, line 53: Why do we need to use `setTimeout`?

src/components/Onboarding/OnboardingChecklist/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <button
+                onClick={() => handleChecklistItemClick(item.group)}
+                type="button"
+                className={classNames(checked && styles.completed)}
+              >
+                <div className={classNames(styles.checkCircle, checked && styles.filled)}>
+                  {checked ? <CheckIcon /> : null}
+                </div>
+
+                <p>{item.title}</p>
```
osamasayed, line 148: We should add a styling here to make sure the text sticks to the left/right depending on RTL or LTR

Screenshot 2024-01-26 at 10 01 50 am

src/components/Onboarding/hooks/useStartOnboarding.ts
```diff
@@ -0,0 +1,13 @@
+import { useOnboarding } from '../OnboardingProvider';
+
+import useBrowserLayoutEffect from '@/hooks/useBrowserLayoutEffect';
+
+const useStartOnboarding = () => {
```
osamasayed, line 5: it seems this hook is not used, we should remove it

src/components/Navbar/Drawer/index.tsx
```diff
@@ -94,13 +97,17 @@ const Drawer: React.FC<Props> = ({
 
   const closeDrawer = useCallback(
     (actionSource = 'click') => {
+      if ((actionSource === 'esc_key' || actionSource === 'outside_click') && !closeOnBlur) {
```
osamasayed, line 100: 1. why do we need to check for those 2 action sources only? there is a third one `navigation`.

2. We should add an ActionSource enum and use it here instead of strings

src/components/Navbar/Drawer/index.tsx
```diff
@@ -41,6 +42,7 @@ interface Props {
   hideCloseButton?: boolean;
   children: ReactNode;
   closeOnNavigation?: boolean;
+  closeOnBlur?: boolean;
```
osamasayed, line 45: nit: we need a better naming for `closeOnBlur` as it's not clear what it stands for. Maybe `shouldCloseDrawer`?

src/redux/slices/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+  initialState,
+  reducers: {
+    // checklist state
+    dismissChecklist: (
+      state: OnboardingState,
+      action: PayloadAction<{ countDismiss: boolean }>,
+    ) => ({
+      ...state,
+      isChecklistVisible: false,
+      ...(action.payload.countDismiss && {
```
osamasayed, line 46: I think we should remove `countDismiss` since it's always `true` when `dismissChecklist` is called.

src/redux/slices/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+    >
+  >;
+};
+
+const initialState: OnboardingState = {
+  isChecklistVisible: true,
+  checklistDismissals: 0,
+  lastChecklistDismissal: null,
+  activeStep: {
```
osamasayed, line 28: nit: `activeStep` should be renamed to `activeGroup` or `activeGroupAndStep` since it's not only holding the current step but also the current group.

src/redux/slices/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+    group: OnboardingGroup.HOMEPAGE,
+    index: 0,
+  },
+  groups: {},
```
osamasayed, line 32: nit: should not we call this `completedGroups` since we only store info related to completed groups?

src/components/Onboarding/steps.ts
osamasayed: There is another issue, the first step does not work except when you click somewhere and scroll first:

Jan-26-2024 3-29-49 pm

src/components/AudioPlayer/AudioPlayerBody/index.tsx
```diff
@@ -9,11 +9,24 @@ import RadioPlaybackControl from '../RadioPlaybackControl';
...
+import OnboardingGroup from '@/types/OnboardingGroup';
 import { AudioPlayerMachineContext } from 'src/xstate/AudioPlayerMachineContext';
 
 const AudioPlayerBody = () => {
   const audioService = useContext(AudioPlayerMachineContext);
   const isRadioMode = useSelector(audioService, (state) => !!state.context.radioActor);
+  const { isActive, activeStepGroup, activeStepIndex, nextStep } = useOnboarding();
+
+  useEffect(() => {
+    if (
```
osamasayed, line 22: add a comment

src/components/Onboarding/OnboardingChecklist/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  };
+
+  const openChecklist = () => {
+    logButtonClick('onboarding_checklist_open');
+    dispatch(setIsChecklistVisible(true));
+  };
+
+  useEffect(() => {
+    // If the user didn’t finish onboarding then we should show the onboarding again after 1 month from the last time. Repeat this behaviuor for 3 months (once per month) and finally, the system should not alert the user anymore for this onboarding.
+    const didNotFinishOnboarding = !!checklistItems.find(({ group }) => {
```
osamasayed, line 51: nit: `isOnboardingFinished`

src/components/Onboarding/OnboardingChecklist/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const oneMonth = 1000 * 60 * 60 * 24 * 30;
+    const lastDismissal = new Date(lastChecklistDismissal);
+
+    if (now - lastDismissal.getTime() > oneMonth) {
+      dispatch(setIsChecklistVisible(true));
+      logEvent('onboarding_checklist_auto_open', {
+        totalDismissals: checklistDismissals,
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
osamasayed, line 75: why are we skipping all dependencies?

src/components/Onboarding/OnboardingChecklist/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+      checklistDismissals >= 3 ||
+      isChecklistVisible ||
+      !lastChecklistDismissal ||
+      !didNotFinishOnboarding
+    ) {
+      return;
+    }
+
+    const now = Date.now();
+    const oneMonth = 1000 * 60 * 60 * 24 * 30;
```
osamasayed, line 66: nit: move `oneMonth` into a const outside the `useEffect` and rename it to `showCheckListAfter` or something similar in-case `oneMonth` changes.

src/components/Onboarding/OnboardingChecklist/OnboardingChecklist.module.scss
osamasayed: `bottom` should be using `inset`

src/components/dls/Button/Button.tsx
ahmedriad1: We currently don't add an example for each variant, instead we add it as a prop and allow you to play with different values.

src/redux/slices/onboarding.ts
ahmedriad1: `activeStep` is the combination of `group` + `index`. A step is unique throughout the onboarding experience. Imo the current name is fine

PR #2091, opened 2024-02-24
Upgrading node engine to 20.x

omarzawahry:
Upgrading the node engine from `18.x` to `20.x`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "engines": {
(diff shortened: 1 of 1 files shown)


ahmedriad1:
Salam alykum Omar, thanks a lot for your PR, JAK! 

We already had a PR to upgrade to node v20 #2053; however, we decided to close it as it had a lot of compatibility issues with the codebase. We instead decided to just upgrade to v18 #2054.

PR #2005, opened 2023-07-24
Feature/1970 Volume control

sksiyer:
### Summary
Issue: 1970 https://github.com/quran/quran.com-frontend-next/issues/1970
Added a volume control to the existing audio controls
The button should
1) Keep in theme with the users theme preference
2) Persist in local storage so the next time they load the app their volume is saved
3) The volume icon should change from loud > quiet > mute
4) Now there is an extra control, the play button should remain in center
5) The volume icons should be from same source (Material io)
6) Use existing state machine to manage context

### Test Plan
Test: https://youtu.be/TnYPMSHa0u0

1) Test volume is controlled with slider
2) Test preference is stored in local storage after refresh
3) Test control is kept in user theme

### Screenshots

| Before | After |
| ------ | ------ |
|Screenshot 2023-07-24 at 20 01 38|Screenshot 2023-07-24 at 20 02 01 |

PR #2112, opened 2024-03-14
Add volume change functionality to audio player

mauwaz:
### Summary
User can now change volume of audio player without having to change system volume.
Volume will persist on reload.

- Volume slider with auto-close after user pick a value (_current delay = 1500ms_)
- Volume icon will update according to volume
  - For volume 0, mute icon will be displayed
  - For volume 1-49, low volume icon will be displayed
  - For volume 50-100, high volume icon will be displayed

### Test Plan
- Play audio of any Surah.
- Click on volume icon.
- Slide left or right to decrease or increase volume

### Known Issue:
Volume change not working on iOS mobile device due to [Apple's policy](https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html#:~:text=Volume%20Control%20in%20JavaScript).

### Screenshots

| Before | After |
| ------ | ------ |
| image | image |


osamasayed:
Alsalmau Alikum @mauwaz Jazak Allahu Khairan for the PR! mashAllah very neat UI and functionality. I am afraid there is already an opened PR that tries to add the same functionality
- #2005 

The above PR is still pending some improvements for it to be mergeable. Was wondering if you might be interested in working on the comments in the other PR better instead of creating a new one? Also, I found a couple of issues related to your PR:

1. Volume controller is not working on Chrome on iOS device.
2. Volume level does not persist when you refresh the page

mauwaz:
> Was wondering if you might be interested in working on the comments in the other PR better instead of creating a new one?

Thank you for heads up. While I appreciate the suggestion, I believe it would be more efficient to proceed with the new PR, as the existing one requires significant refactoring and would involve duplication of effort. 🤔

mauwaz:
#### Pt 1:
As per [Apple's documentation](https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html#:~:text=Volume%20Control%20in%20JavaScript) under section "_Volume Control in Javascript_":

> On iOS devices, the audio level is always under the user’s physical control. The volume property is not settable in JavaScript. Reading the volume property always returns 1.

#### Pt 2: 
Resolved

osamasayed:
> > Was wondering if you might be interested in working on the comments in the other PR better instead of creating a new one?
> 
> 
> 
> Thank you for heads up. While I appreciate the suggestion, I believe it would be more efficient to proceed with the new PR, as the existing one requires significant refactoring and would involve duplication of effort. 🤔 

I see. I just feel bad about the effort that has been duplicated already between this PR and the other one. My bad there is no clear features backlog for contributors to know which features are being worked on by whom. InshAllah we will improve this flow very soon. Anyways let's roll with this PR InshaAllah. Will be testing it a bit more and let you know If I have any comments.

src/redux/Provider.tsx
```diff
@@ -49,7 +49,8 @@ const ReduxProvider = ({ children, locale }) => {
         const reciterId =
           userPreferences[PreferenceGroup.AUDIO]?.reciter ||
           audioService.getSnapshot().context.reciterId;
-        audioService.send({ type: 'SET_INITIAL_CONTEXT', playbackRate, reciterId });
+        const { volume } = audioService.getSnapshot().context;
```
osamasayed, line 52: I see `audioService.getSnapshot().context` is repeated multiple times for `volume`, `playbackRate` and `reciterId` so perhaps we should put it in a `const` and reuse it for all.

src/components/AudioPlayer/Buttons/VolumeControl/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }, VOLUME_CONTROL_CLOSE_DELAY);
+  };
+
+  return (
+    <div dir={direction}>
+      <PopoverMenu
+        isPortalled
+        isModal
+        isOpen={isOpen}
+        contentClassName={styles.VolumeControl__content}
```
osamasayed, line 63: nitpicking: we don't use `__` in the naming of CSS classes. You can directly set it to `styles.content` and update `src/components/AudioPlayer/Buttons/VolumeControl/VolumeControl.module.scss` to:

```
.content {
  margin-block-end: var(--spacing-large);
  padding: var(--spacing-medium);
}
```

src/components/AudioPlayer/Buttons/VolumeControl/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      volume: newVolume / 100,
+    });
+    setIcon(getVolumeIconType(newVolume));
+
+    // Auto close volume control popover after a value is selected
+    if (timerId.current) clearTimeout(timerId.current);
+    timerId.current = setTimeout(() => {
+      setIsOpen(false);
+    }, VOLUME_CONTROL_CLOSE_DELAY);
+  };
```
osamasayed, line 55: To make sure no memory leak occurs, `setTimeout` and `clearTimeout` should be set inside `useEffect`. For reference, you can take a look at `src/components/Verse/AdvancedCopy/VerseAdvancedCopy.tsx`
Screenshot 2024-03-18 at 12 31 01 AM

src/components/AudioPlayer/Buttons/VolumeControl/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const audioService = useContext(AudioPlayerMachineContext);
+  const volume = useSelector(audioService, (state) => state.context.volume);
+
+  const [isOpen, setIsOpen] = useState(false);
+  const [icon, setIcon] = useState(getVolumeIconType(volume * 100));
+  const timerId = useRef<NodeJS.Timeout>();
+
+  const handleVolumeChange = ([newVolume]: number[]) => {
+    logEvent('audio_player_volume_change');
```
osamasayed, line 43: We have a logger function called `logValueChange` which is better to use in this case to also log volume values e.g. `logValueChange('audio_player_volume', volume, newVolume)`

mauwaz:
Feedback addressed

osamasayed:
> Feedback addressed

Jazak Allahu Khairan!

osamasayed (approved):

PR #2129, opened 2024-04-02
Set the starting verse before navigating to login

amdat99:
### Summary
The current verse is set before navigating to login in case the back button is pressed from login.

### DIFF
--- a/src/components/Verse/Notes/index.tsx
+++ b/src/components/Verse/Notes/index.tsx
@@ -43,14 +43,16 @@ const VerseNotes = ({ verseKey, isTranslationView, hasNotes }: VerseNotesProps) 
     getUserProfile,
   );
 
-  const onItemClicked = () => {
+  const onItemClicked = async () => {
     const isUserLoggedIn = isLoggedIn();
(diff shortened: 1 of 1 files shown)


src/components/Verse/Notes/index.tsx
```diff
@@ -43,14 +43,16 @@ const VerseNotes = ({ verseKey, isTranslationView, hasNotes }: VerseNotesProps)
...
+  const onItemClicked = async () => {
     const isUserLoggedIn = isLoggedIn();
     logButtonClick('verse_actions_menu_note', {
       isTranslationView,
       isLoggedIn,
     });
     if (!isUserLoggedIn) {
-      router.push(getLoginNavigationUrl(getChapterWithStartingVerseUrl(verseKey)));
+      const currentRoute = getChapterWithStartingVerseUrl(verseKey);
+      await router.push(currentRoute)
```
AbdQaadir, line 54: This line is not needed anymore. The next line is what you probably needed.

AbdQaadir (changes requested):

PR #2130, opened 2024-04-02
Fixing the layout shift that happens in the home page and chapterId page

0XYoussefX0:
### Summary

fixed the layout shift that happens when you open the language dropdown, and when you open any of the drawers available on the navbar. the first one is caused by a library used by radix ui under the hood called: react-remove-scroll, and the second one is caused by the usePreventBodyScrolling hook.

Here is the [Before](https://www.loom.com/share/a48d73440c4142e9a5fa2f7f08c6aee5?sid=29360acc-bd1e-4b55-8dc7-dbb9ddf6209c) and [After](https://www.loom.com/share/d9cc43a9b81341fc8b73ccae2dd30e5c?sid=d35d61e2-d69c-4535-ade1-17d8b23541ff)

### DIFF
--- a/src/components/Navbar/Navbar.tsx
+++ b/src/components/Navbar/Navbar.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 
 import classNames from 'classnames';
 import { useSelector, shallowEqual } from 'react-redux';
@@ -7,6 +7,7 @@ import styles from './Navbar.module.scss';
 import NavbarBody from './NavbarBody';
 
 import { useOnboarding } from '@/components/Onboarding/OnboardingProvider';
+import useScrollBarWidth from '@/hooks/useScrollBarWidth';
 import { selectNavbar } from '@/redux/slices/navbar';
 
 const Navbar = () => {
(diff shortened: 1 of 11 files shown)


0XYoussefX0:
@osamasayed @AhmedCodeGuy please can you review this PR ? 
it builds perfectly fine and solves a layout shift issue that will improve the UX for users visiting from a computer.

src/components/Navbar/Navbar.tsx
```diff
@@ -11,13 +11,43 @@ import { selectNavbar } from '@/redux/slices/navbar';
...
-  const { isVisible: isNavbarVisible } = useSelector(selectNavbar, shallowEqual);
+  const {
+    isVisible: isNavbarVisible,
+    isSettingsDrawerOpen,
+    isNavigationDrawerOpen,
+    isSearchDrawerOpen,
+  } = useSelector(selectNavbar, shallowEqual);
   const showNavbar = isNavbarVisible || isActive;
 
+  useEffect(() => {
```
AhmedCodeGuy, line 22: Could you please extract all of the useEffect logic to a custom hook and just use the custom hook in the `Navbar` component.

One more thing regarding perfomance, is to add throttling to the `resize` event listener function.

```
function throttle(fn, time) {
    let timeout = null;
    return function () {
        if(timeout) return;
        const context = this;
        const args = arguments;
        const later = () => {
            fn.call(context, ...args);
            timeout = null;
        }
        timeout = setTimeout(later, time);
    }
}

const handleResizeThrottled = throttle(setScrollbarWidth, 500);
```

AhmedCodeGuy (approved):
LGTM! JAK <3

0XYoussefX0:
@AhmedCodeGuy Done 👍

PR #2158, opened 2024-05-16
validate page id by the selected mushaf id

AhmedCodeGuy:
# Summary

Fixes #QF-433

In creating or updating a goal the button will be disabled if the page number is outside the range of the selected Mushaf settings 

## Type of change

## Test plan

I created and updated a goal using different Mushaf settings and the button is disabled when I type a page number that is greater than the Mushaf pages count:

- Indopak 15 lines
- Indopak 16 lines
- Uthmani

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/components/ReadingGoal/UpdateReadingGoalModal/index.tsx
+++ b/src/components/ReadingGoal/UpdateReadingGoalModal/index.tsx
@@ -175,12 +175,16 @@ const UpdateReadingGoalModal = ({ isDisabled, goal }: UpdateReadingGoalButtonPro
   };
 
   const getIsUpdateDisabled = () => {
-    return !validateReadingGoalData(chaptersData, {
-      type,
-      pages,
-      seconds,
-      range,
-    });
+    return !validateReadingGoalData(
+      chaptersData,
+      {
+        type,
+        pages,
+        seconds,
+        range,
+      },
+      mushaf,
+    );
   };
 
   return (
--- a/src/components/ReadingGoal/UpdateReadingGoalModal/index.tsx
+++ b/src/components/ReadingGoal/UpdateReadingGoalModal/index.tsx
@@ -2,7 +2,6 @@
 import { useCallback, useContext, useEffect, useMemo, useState } from 'react';
 
 import useTranslation from 'next-translate/useTranslation';
-import { useSelector, shallowEqual } from 'react-redux';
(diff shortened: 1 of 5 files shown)


src/utils/validator.ts
```diff
@@ -133,13 +133,35 @@ export const isValidPageId = (pageId: string | number): boolean => {
...
+ */
+export const isValidPageIdBySelectedMushafId = (
+  pageId: string | number,
+  mushafId: Mushaf,
+): boolean => {
+  const pageIdNumber = Number(pageId);
+  const MUSHAF_COUNT = PAGES_MUSHAF_MAP[mushafId];
+  // if it's not a numeric string or it's numeric but out of the range of the selected Mushaf
+  if (Number.isNaN(pageIdNumber) || pageIdNumber > MUSHAF_COUNT || pageIdNumber < 1) {
+    return false;
```
osamasayed, line 160: This bit is almost identical to what is inside `isValidPageId`, I suggest we keep it DRY and move it to a common function and re-use it for both. e.g. `isValidaPageNumber(maximumMushafPages: number)`

src/components/ReadingGoalPage/utils/validator.ts
```diff
@@ -29,9 +30,10 @@ interface ReadingGoalPayload {
 export const validateReadingGoalData = (
   chaptersData: ChaptersData,
   { type, pages, seconds, range }: ReadingGoalPayload,
+  mushafId: Mushaf = Mushaf.QCFV1,
```
osamasayed, line 33: I think we should remove the default value from here since `mushafId` is required always passed.

src/components/ReadingGoal/UpdateReadingGoalModal/index.tsx
```diff
@@ -175,12 +175,16 @@ const UpdateReadingGoalModal = ({ isDisabled, goal }: UpdateReadingGoalButtonPro
...
-    });
+    return !validateReadingGoalData(
+      chaptersData,
+      {
+        type,
+        pages,
+        seconds,
+        range,
+      },
+      mushaf,
```
osamasayed, line 186: Can we also refactor this component to use `useGetMushaf` hook?

osamasayed (approved):

PR #2160, opened 2024-05-21
QF-353

AhmedCodeGuy:
# Summary

Fixes #QF-353

Fixed some words being cut off on the Firefox browser.

## Type of change

## Test plan

I have tested different font sizes on Chrome, and Firefox using the Uthmani V1 and V2 as these are the only fonts affected by the change made.

## Checklist

## Screenshots or videos

| Before | After |
| ------ | ------ |
| Screenshot 2024-05-21 at 5 19 47 PM | Screenshot 2024-05-21 at 5 20 50 PM |


Ayman24x7:
Actually, the problem is with the outdated font. Your solution does not solve 2:232 Waw cutting off. Page 37.

src/components/dls/QuranWord/GlyphWord.tsx
```diff
@@ -51,22 +50,23 @@ const GlyphWord = ({
...
-  const { quranTextFontScale, mushafLines } = quranReaderStyles;
+  const { quranTextFontScale, mushafLines } = useSelector(selectQuranReaderStyles, shallowEqual);
+  const innerHTML = {
+    __html: ` ${getWordText(qpcUthmaniHafs, textCodeV1, textCodeV2, font, isFontLoaded)}`,
+  };
+  const className = classNames(styles.styledWord, {
+    [styles.wordSpacing]: quranTextFontScale < 6,
+    [styles.fallbackText]: !isFontLoaded,
+    [styles[getFontClassName(FALLBACK_FONT, quranTextFontScale, mushafLines, true)]]: !isFontLoaded,
+  });
```
osamasayed, line 61: I understand this might be subjective and while I feel that assigning a value in a separate `const` instead of inline is clearer but if it's only used once, I recommend we should keep it inline as assigning a value in a const will add some work for the garbage collector to collect that const after it's not longer used.

src/components/dls/QuranWord/GlyphWord.tsx
```diff
@@ -51,22 +50,23 @@ const GlyphWord = ({
   font,
   isFontLoaded,
 }: UthmaniWordTextProps) => {
-  const quranReaderStyles = useSelector(selectQuranReaderStyles, shallowEqual);
-  const { quranTextFontScale, mushafLines } = quranReaderStyles;
+  const { quranTextFontScale, mushafLines } = useSelector(selectQuranReaderStyles, shallowEqual);
+  const innerHTML = {
+    __html: ` ${getWordText(qpcUthmaniHafs, textCodeV1, textCodeV2, font, isFontLoaded)}`,
+  };
```
osamasayed, line 56: Adding the extra spacing before the word broke V2 Mushaf's layout  since `GlyphWord` is common for all browsers and also for all Glyph-based Mushafs.
Before:
Screenshot 2024-05-28 at 3 56 13 AM

After:
Screenshot 2024-05-28 at 3 55 58 AM

src/components/dls/QuranWord/GlyphWord.module.scss
```diff
@@ -1,14 +1,28 @@
...
+
 // this is needed to reduce CLS when we are displaying the fallback text by trying to make the font-sizes of the original text and the fallback text as close as possible so that once the original font loads, there will be as little visible CLS as possible.
 .fallbackText {
   font-family: UthmanicHafs !important;
 }
+
+.tajweedTextHighlighted {
+  color: transparent;
+  text-shadow: 0px 0px var(--color-success-medium);
+}
```
osamasayed, line 28: I see we are not using this, we should remove it

osamasayed (changes requested):

src/components/dls/QuranWord/GlyphWord.tsx
AhmedCodeGuy: Sure, would that affect the performance?

osamasayed (approved):

PR #482, opened 2021-09-22
Implement ayah highlighting

muhajirdev:
### Summary
Ready for review, still doing performance testing.

PR #2163, opened 2024-05-29
migrate tajweed mushaf to the latest version

AhmedCodeGuy:
# Summary

Fixes #482

Update redux persist migrations to overwrite `tajweed` quran font to `tajweed_v4`.

## Type of change

## Test plan

I have checked the local storage and redux devtools to make sure that the `quranFont` property is set to the newer version but still showing the old version if I selected the `tajweed` mushaf from the settings.

## Checklist

### DIFF
--- a/src/redux/migrations.ts
+++ b/src/redux/migrations.ts
@@ -234,7 +234,7 @@ export default {
     ...state,
     quranReaderStyles: {
       ...state.quranReaderStyles,
-      quranFont: QuranFont.TajweedV4,
+      quranFont: state.quranFont === QuranFont.Tajweed ? QuranFont.TajweedV4 : state.quranFont,
     },
   }),
 };
--- a/src/redux/migrations.ts
+++ b/src/redux/migrations.ts
@@ -234,7 +234,9 @@ export default {
     ...state,
     quranReaderStyles: {
       ...state.quranReaderStyles,
-      quranFont: state.quranFont === QuranFont.Tajweed ? QuranFont.TajweedV4 : state.quranFont,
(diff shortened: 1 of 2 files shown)


src/redux/migrations.ts
```diff
@@ -230,4 +230,11 @@ export default {
       wordByWordFontScale: initialState.quranReaderStyles.wordByWordFontScale,
     },
   }),
+  31: (state) => ({
+    ...state,
+    quranReaderStyles: {
+      ...state.quranReaderStyles,
+      quranFont: QuranFont.TajweedV4,
+    },
+  }),
```
osamasayed, line 239: If this is applied, it will change every single user's Musahf to Tajweed! We need to check if the user's current setting is Tajweed first before applying this.

AhmedCodeGuy: Right! will add the condition

src/redux/migrations.ts
```diff
@@ -230,4 +230,13 @@ export default {
       wordByWordFontScale: initialState.quranReaderStyles.wordByWordFontScale,
     },
   }),
+  31: (state) => ({
+    ...state,
+    quranReaderStyles: {
+      ...state.quranReaderStyles,
+      ...(state.quranFont === QuranFont.Tajweed && {
```
osamasayed, line 237: `state.quranFont` is not a valid value, it should be `state.quranReaderStyles.quranFont`

AhmedCodeGuy: I was wondering why it is no longer working xD

osamasayed (approved):

PR #2180, opened 2024-06-26
fix navigation sidebar

AhmedCodeGuy:
# Summary

Fixes #576

Fixes the navigation sidebar shifting on mushaf settings changes.

## Type of change

## Test plan

This was tested for web/mobile (chrome), and LTR/RTL.

## Checklist

- [ ] I have commented on my code, particularly in hard-to-understand areas


src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
```diff
@@ -70,10 +70,14 @@ const SidebarNavigation = () => {
...
-        [styles.spaceOnTopMenuExpanded]: isContextMenuExpanded,
-        [styles.spaceOnTopMenuCollapse]: !isContextMenuExpanded,
+        [styles.visibleContainer]: isSidebarVisible === true && isNavbarVisible,
+        [styles.visibleContainerCollapsed]: isSidebarVisible === true && !isNavbarVisible,
+        [styles.containerAuto]: isSidebarVisible === 'auto' && isNavbarVisible,
+        [styles.containerAutoCollapsed]: isSidebarVisible === 'auto' && !isNavbarVisible,
+        [styles.inVisibleContainer]: isNavbarVisible,
+        [styles.inVisibleContainerCollapsed]: !isNavbarVisible,
+        [styles.spaceOnTop]: isSidebarVisible && isNavbarVisible,
+        [styles.spaceOnTopCollapsed]: isSidebarVisible && !isNavbarVisible,
```
osamasayed, line 80: I see there are a few duplications in this section. For example:
- `isSidebarVisible === true && isNavbarVisible` and `isSidebarVisible && isNavbarVisible`
- `isSidebarVisible && !isNavbarVisible` and `isSidebarVisible === true && !isNavbarVisible`

We should keep it DRY and technically they should share the same exact classes

osamasayed (changes requested):

src/components/QuranReader/SidebarNavigation/SidebarNavigation.tsx
AhmedCodeGuy: `isSidebarVisible === true` is different from `isNavbarVisible` as `isNavbarVisible` could be `true` or `auto`, however the first condition applies only when `isNavbarVisible` equals to `true` only, same for the second point.

osamasayed: If this is the case, the code is very hard to follow. We defo need to refactor it with better naming and introduce variables. E.g. extract `isSidebarVisible === 'auto'` to a variable since it's used twice for example.

AhmedCodeGuy: `isSidebarVisible === true` is different from `isSidebarVisible` as `isSidebarVisible` could be `true` or `auto`.

src/components/QuranReader/SidebarNavigation/SidebarNavigation.module.scss
```diff
@@ -41,15 +48,46 @@ $progress-bar-height: var(--spacing-micro);
   display: flex;
   flex-direction: column;
   overflow-y: hidden;
-  @include sidebarInvisible();
+  margin-top: -1px;
```
osamasayed, line 51: If we need a negative value, we usually do `calc(-1*var...)`

osamasayed (changes requested):

osamasayed (approved):

PR #2187, opened 2024-07-18
Feature/video-gen

AhmedCodeGuy:
# Summary

Fixes #640

Video generator 

## Type of change

## Test plan

This has been tested on chrome for all different themes, and different screen sizes.

## Checklist


src/components/MediaMaker/RenderControls/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (isCopied === true) {
+      timeoutId = setTimeout(() => setIsCopied(false), COPY_TIMEOUT_MS);
+    }
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [isCopied]);
+
+  const onCopyLinkClicked = () => {
+    logButtonClick('video-generation-copy-link');
```
osamasayed, line 52: the conversion of the vent should have `_` -> `video_generation_copy_link`

osamasayed (changes requested):

src/utils/validator.ts
```diff
@@ -269,6 +271,58 @@ export const isValidVerseKey = (chaptersData: ChaptersData, verseKey: string): b
...
+  }
+  if (getVerseNumberFromKey(endVerseKey) < startVerseKey) {
+    return false;
+  }
+  if (getChapterNumberFromKey(endVerseKey) !== Number(chapterID)) {
+    return false;
+  }
+  return true;
+};
+
```
osamasayed, line 325: Add unit tests

src/utils/media/utils.ts
```diff
@@ -162,3 +169,41 @@ export const mutateGeneratedMediaCounter = (currentData) => {
     },
   };
 };
+
+export const getVerseValue = (
```
osamasayed, line 173: Unit tests should be added.

src/components/MediaMaker/Settings/VideoSettings.tsx
```diff
@@ -89,27 +70,6 @@ const VideoSettings: React.FC<Props> = ({
     removeQueryParam(Object.values(QueryParam));
   }, [dispatch, removeQueryParam, seekToBeginning]);
 
-  const onSettingsUpdate = useCallback(
```
osamasayed: Is there a reason why this was moved to `src/pages/quran-media-maker/index.tsx` as that file has so many lines of code already.

src/components/MediaMaker/Settings/VideoSettings.tsx
```diff
@@ -26,41 +25,23 @@ import ImageIcon from '@/icons/photo.svg';
...
   reciters: Reciter[];
   seekToBeginning: () => void;
+  onSettingsUpdate: (settings: ChangedSettings, key?: keyof MediaSettings, value?: any) => void;
   getCurrentFrame: () => void;
   isFetching: boolean;
   inputProps: any;
   mediaSettings: MediaSettings;
 };
 
-const MEDIA_SETTINGS_TO_QUERY_PARAM = {
```
osamasayed: Is there a reason why this was moved to `src/pages/quran-media-maker/index.tsx` as that file has so many lines of code already.

src/pages/quran-media-maker/index.tsx
```diff
@@ -99,6 +123,61 @@ const MediaMaker: NextPage<MediaMaker> = ({
...
+    [QueryParam.OPACITY]: String(opacity),
+    [QueryParam.BORDER_COLOR]: borderColor,
+    [QueryParam.BORDER_SIZE]: String(borderSize),
+    [QueryParam.FONT_COLOR]: fontColor,
+    [QueryParam.VIDEO_ID]: String(videoId),
+    [QueryParam.QURAN_TEXT_FONT_SCALE]: String(quranTextFontScale),
+    [QueryParam.QURAN_TEXT_FONT_STYLE]: String(quranTextFontStyle),
+    [QueryParam.TRANSLATION_FONT_SCALE]: String(translationFontScale),
+    [QueryParam.ORIENTATION]: orientation,
+  };
```
osamasayed, line 145: 2 notes:

1. Do we need to use `String(...)` for most of the values, won't it be string by default?
2. If we need to, it seems `orientation`, `fontColor` and `borderColor` were missed.

src/hooks/useGetQueryParamOrReduxValue.ts
```diff
@@ -201,22 +199,30 @@ const useGetQueryParamOrReduxValue = (
   }
   // @ts-ignore
   const reduxValue = useSelector(...useSelectorArguments);
-  // TODO: this bit is identical to the one in useGetQueryParamOrXstateValue.ts, keep it DRY
+  const surahReduxValue = useSelector(
+    QUERY_PARAMS_DATA[QueryParam.SURAH].reduxSelector,
+    QUERY_PARAMS_DATA[QueryParam.SURAH].reduxEqualityFunction,
+  );
```
osamasayed, line 205: This check should not be a general one that applies to every query param but rather it's specific to when it's `from` or `to` query params.

src/hooks/useGetQueryParamOrReduxValue.ts
```diff
@@ -201,22 +199,30 @@ const useGetQueryParamOrReduxValue = (
...
     const paramStringValue = String(query[queryParam]);
     const isValidValue = validate(paramStringValue, chaptersData, query);
+    const parsedQueryParamValue = getQueryParamValueByType(paramStringValue, valueType);
+    const isQueryParamDifferent = getIsQueryParamDifferent(paramStringValue, valueType, reduxValue);
+
+    if (queryParam === QueryParam.VERSE_TO || queryParam === QueryParam.VERSE_FROM) {
+      return {
+        value: getVerseValue(query, queryParam, chaptersData, QUERY_PARAMS_DATA, surahReduxValue),
+        isQueryParamDifferent,
+      };
```
osamasayed, line 218: is there a more dynamic way to specify the `value` rather than adding this if condition check inside this dynamic function?

src/utils/media/utils.ts
```diff
@@ -162,3 +169,41 @@ export const mutateGeneratedMediaCounter = (currentData) => {
...
+  query,
+  queryParam,
+  chaptersData,
+  QUERY_PARAMS_DATA,
+  surahReduxValue,
```
osamasayed, line 178: Typescript types should be added

src/utils/media/utils.ts
```diff
@@ -162,3 +169,41 @@ export const mutateGeneratedMediaCounter = (currentData) => {
     },
   };
 };
+
```
osamasayed, line 172: JsDoc should be added.

src/utils/media/utils.ts
```diff
@@ -162,3 +169,41 @@ export const mutateGeneratedMediaCounter = (currentData) => {
...
+  QUERY_PARAMS_DATA,
+  surahReduxValue,
+) => {
+  const queryParamProps = QUERY_PARAMS_DATA[queryParam];
+  const surahParamProps = QUERY_PARAMS_DATA[QueryParam.SURAH];
+  const verseFromStringValue = String(query[QueryParam.VERSE_FROM]);
+  const verseToStringValue = String(query[QueryParam.VERSE_TO]);
+  const surahStringValue = String(query[QueryParam.SURAH]);
+
+  const chapterID = surahParamProps.validate(surahStringValue) ? surahStringValue : surahReduxValue;
```
osamasayed, line 186: `chapterId` -> `surahId`

osamasayed (changes requested):

src/components/MediaMaker/Root.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * To make things work (e.g. run npx remotion studio), I had to:
+ *   - hardcode locale to 'en' in util.ts and chapter.ts
+ *   - install scss loaders to make things work - https://www.remotion.dev/docs/webpack#enable-sassscss-support
+ *   - install querystring module?
+ *   -
+ *   -
+ *   -
+ *   -
+ *   -
+ */
```
osamasayed, line 29: This bit needs to be removed

src/components/MediaMaker/Settings/VideoSettings.tsx
AhmedCodeGuy: I need to invoke `setIsUpdating` from inside `onSettingsUpdate,` so I thought moving it here would be better.

src/pages/quran-media-maker/index.tsx
AhmedCodeGuy: These values come from `mediaSettings`, so I have to convert number values to strings, for the `orientation`, `fontColor`, `borderColor`, and `backgroundColor` they are already strings.

src/hooks/useGetQueryParamOrReduxValue.ts
AhmedCodeGuy: This is true, but as you know we can't use hooks conditionally, in our case `useSelector`.

osamasayed: Why cannot we make `reduxValue` select whatever `surahReduxValue` is trying to select when there is a `customHandler`?

AhmedCodeGuy: We need both the redux value, which will be the ayah, and the surah redux value simultaneously.

vitest.config.ts
```diff
@@ -14,6 +14,8 @@ export default defineConfig({
   },
   resolve: {
     alias: {
+      // src: path.resolve(__dirname, './src'),
+      '@/types': path.resolve(__dirname, './types'),
```
osamasayed, line 18: We should remove these

PR #2250, opened 2024-11-30
QF 1272 search autocomplete

osamasayed:
# Summary

Fixes #QF-1272

This PR uses quick search mode as the default for the search across the app.


AhmedCodeGuy (approved):

src/components/Search/SearchResults/SearchResults.module.scss
```diff
@@ -21,4 +21,8 @@
 }
 .navigationItemContainer {
   margin-inline-end: var(--spacing-small);
+  em {
```
AhmedCodeGuy, line 24: I can't see this class being used anywhere.

PR #2273, opened 2025-01-06
Protects next.js proxy from unauthorized usage

mohsinayoob:
# Summary
Fixes #[QF-1552]

Protects next.js proxy from unauthorized usage. For integration or client side it checks for the origin of the request and for server side it checks signature for each request.
New ENV Vars:
- ALLOWED_ORIGINS=quran.com,test.quran.com
- PROXY_SIGNATURE_TOKEN=123456

## Type of change

## Test plan

This should state how this PR have been tested.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings
- [ ] I have commented on my code, particularly in hard-to-understand areas

## Screenshots or videos

| Before | After |
| ------ | ------ |
| IMAGE HERE | IMAGE HERE |

[QF-1552]: https://quranfoundation.atlassian.net/browse/QF-1552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/api/proxy/[...path].ts
```diff
@@ -3,23 +3,90 @@ import { EventEmitter } from 'events';
...
 EventEmitter.defaultMaxListeners = Number(process.env.PROXY_DEFAULT_MAX_LISTENERS) || 100;
 
-// This file sets up a proxy middleware for API requests. It is needed to forward requests from the frontend
-// to the backend server, allowing for features like cookie handling and request body fixing, which are essential
-// for maintaining session state and ensuring correct request formatting while in a cross domain env.
+const isOriginAllowed = (origin: string | undefined): boolean => {
+  if (!origin) return false;
+  const url = new URL(origin);
+  const { hostname } = url;
+  return ALLOWED_DOMAINS.some((domain) => hostname.endsWith(domain));
```
Comment (osamasayed) [inline src/pages/api/proxy/[...path].ts:34]: Why do we need `endsWith` here? `ALLOWED_DOMAINS` will be exact match + all subdomains already, I believe we should do `return ALLOWED_DOMAINS.includes(hostname);` issue with `endsWith` is that any it would accept any domain that happens to end with one of the configured ones for example `copy-quran.com` which we don't want.

osamasayed (dismissed):

src/pages/api/proxy/[...path].ts
Comment (osamasayed) [inline src/pages/api/proxy/[...path].ts:34]: Why do we need `endsWith` here? `ALLOWED_DOMAINS` will be exact match + all subdomains already, I believe we should do `return ALLOWED_DOMAINS.includes(hostname);` issue with `endsWith` is that any it would accept any domain that happens to end with one of the configured ones for example `copy-quran.com` which we don't want.

osamasayed (changes requested):

src/pages/api/proxy/[...path].ts
Comment (mohsinayoob) [reply src/pages/api/proxy/[...path].ts:34]: I agree. I did it for testing but forgot to revert. Masha Allah you caught this. Thanks

osamasayed (approved):

osamasayed (approved):

PR #1974, opened 2023-05-24
Fix inaccurate navigation

ahmedriad1:
### Summary
Navigation in the translation view was totally inaccurate. This is due to:
1. We render virtuoso list items as pages, not verses, to fetch data in one place. Therefore, in `useScrollToVirtualizedTranslationView`, when we scroll to an ayah, we try to approximate where it is. 
2. When ayahs are loaded, the scroll position changes due to a difference in the real size vs. the skeleton one, therefore, showing the user a different ayah. 

### Solution
1. We render ayahs individually, and SWR will cache requests for the same page. Now, when we navigate, we can pass the ayah index directly to Virtuoso.
2. When we scroll to un-loaded ayahs, we re-adjust the scroll position after they're loaded.

PR #2349, opened 2025-03-12
QF-1974 - P2

AhmedCodeGuy:
# Summary

Fixes #1974
- add all query params to the url


AhmedCodeGuy:
@osamasayed I merged this one as it was required to be on testing immediately, but feel free to leave any comments and I will address them on a follow up PR.

PR #2363, opened 2025-03-19
updated page title and meta desc, previously they were generic

isadreddin:
but ow we use the question and the summary of the answer

| IMAGE HERE | IMAGE HERE |
image
image

### DIFF
--- a/src/pages/[chapterId]/answers/[questionId].tsx
+++ b/src/pages/[chapterId]/answers/[questionId].tsx
@@ -24,6 +24,7 @@ import {
 } from '@/utils/staticPageGeneration';
 import { isValidVerseKey } from '@/utils/validator';
 import ChaptersData from 'types/ChaptersData';
+import { toLocalizedVerseKey } from '@/utils/locale';
 
 type QuestionPageProps = {
   hasError?: boolean;
@@ -44,7 +45,7 @@ const QuestionPage: NextPage<QuestionPageProps> = ({
   questionData,
   verseKey,
 }) => {
-  const { lang } = useTranslation('question');
+  const { t, lang } = useTranslation('question');
 
   if (hasError) {
     return <Error statusCode={500} />;
@@ -56,7 +57,7 @@ const QuestionPage: NextPage<QuestionPageProps> = ({
   return (
     <>
       <NextSeoWrapper
-        title={'Quran ' + String(verseKey) + ' - ' + String(body)}
+        title={t('quran-reader:q-and-a.quran') + ' ' + toLocalizedVerseKey(verseKey, lang) + ' - ' + body}
         image={getExploreAnswersOgImageUrl({
           locale: lang,
         })}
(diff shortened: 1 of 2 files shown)


AhmedCodeGuy (approved):

src/pages/[chapterId]/answers/[questionId].tsx
```diff
@@ -44,27 +44,27 @@ const QuestionPage: NextPage<QuestionPageProps> = ({
...
+        title={'Quran ' + String(verseKey) + ' - ' + String(body)}
         image={getExploreAnswersOgImageUrl({
           locale: lang,
         })}
         imageWidth={1200}
         imageHeight={630}
         canonical={getCanonicalUrl(lang, navigationUrl)}
         languageAlternates={getLanguageAlternates(navigationUrl)}
-        description={`${verseKey} ${body} ${t('questions-meta-desc')}`}
+        description={String(summary)}
```
Comment (osamasayed) [inline src/pages/[chapterId]/answers/[questionId].tsx:67]: No need to add `String` since summary is a string already.

src/pages/[chapterId]/answers/[questionId].tsx
```diff
@@ -44,27 +44,27 @@ const QuestionPage: NextPage<QuestionPageProps> = ({
...
 
-  const { type, theme: themes, body } = questionData as Question;
+  const { type, theme: themes, body, summary } = questionData as Question;
   const navigationUrl = getAnswerNavigationUrl(questionId, verseKey);
 
   return (
     <>
       <NextSeoWrapper
-        title={t('quran-reader:q-and-a.explore_answers')}
+        title={'Quran ' + String(verseKey) + ' - ' + String(body)}
```
Comment (osamasayed) [inline src/pages/[chapterId]/answers/[questionId].tsx:59]: "`Quran`" needs to be localized since switching to other locales will show it with "Quran" word, same for `verseKey`, we have a function called `toLocalizedVerseKey` which you should use.

osamasayed (changes requested):

src/pages/[chapterId]/answers/[questionId].tsx
```diff
@@ -125,4 +125,4 @@ export const getStaticPaths: GetStaticPaths = async () => ({
   fallback: 'blocking', // will server-render pages on-demand if the path doesn't exist
 });
 
-export default QuestionPage;
+export default QuestionPage;
```
Comment (AhmedCodeGuy) [inline src/pages/[chapterId]/answers/[questionId].tsx:128]:
```suggestion
export default QuestionPage;

```

src/pages/[chapterId]/answers/[questionId].tsx
Comment (isadreddin) [reply src/pages/[chapterId]/answers/[questionId].tsx]: done and did the same for body

src/pages/[chapterId]/answers/[questionId].tsx
Comment (isadreddin) [reply src/pages/[chapterId]/answers/[questionId].tsx]: Did the following: 
1. import { toLocalizedVerseKey } from '@/utils/locale';
2. updated quran-reader.json with a new key 'quran'
3. t('quran-reader:q-and-a.quran')

AhmedCodeGuy (approved):

PR #2462, opened 2025-07-13
feat: add static props with chapters data to about, developers and support pages

AhmedCodeGuy:
# Summary

Closes: [QF-2855](https://quranfoundation.atlassian.net/browse/QF-2855)


osamasayed (changes requested):
@AhmedCodeGuy why don't we just add the same solution to all static pages? e.g. `src/pages/beyond-ramadan/index.tsx`

PR #2614, opened 2025-12-02
fix(bookmarks): proper cache invalidation for cross-page updates

AhmedCodeGuy:
## Summary

Closes: [QF-3097](https://quranfoundation.atlassian.net/browse/QF-3097) and [QF-3930](https://quranfoundation.atlassian.net/browse/QF-3930)

- Replace `useSWRImmutable` with `useSWR` for bookmark hooks since bookmarks are mutable data
- Add centralized cache invalidation via `useBookmarkCacheInvalidator` hook
- Add `BOOKMARK_CACHE_PATHS` constants for pattern-based cache invalidation
- Enable `revalidateOnFocus` for cross-tab sync
- Fix Save to Collection modal not updating after removing verse from collection

### Why `useSWR` over `useSWRImmutable`

Bookmarks are mutable data that can be changed from multiple pages (reader, profile, collections). `useSWRImmutable` doesn't properly invalidate cache when cleared via `globalMutate()`.

### Cache Invalidation

Pattern-based invalidation using `BOOKMARK_CACHE_PATHS`:
- `AYAHS_RANGE` - Bulk bookmark fetch for reader pages
- `BOOKMARK` - Single bookmark check  
- `BOOKMARKS_LIST` - Bookmarks list
- `COLLECTIONS` - Collections list
- `BOOKMARK_COLLECTIONS` - Which collections a bookmark belongs to

## Test Plan

- [x] Bookmark a verse in reader view
- [x] Navigate to profile page, delete the bookmark
- [x] Navigate back to reader - bookmark icon should be unfilled
- [x] Add verse to collection from reader
- [x] Remove verse from collection page
- [x] Open Save to Collection modal - collection should be unchecked
- [x] Test cross-tab: bookmark in one tab, switch to another tab - should sync on focus

[QF-3097]: https://quranfoundation.atlassian.net/browse/QF-3097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QF-3930]: https://quranfoundation.atlassian.net/browse/QF-3930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useBookmarkCollections.ts
+++ b/src/hooks/useBookmarkCollections.ts
@@ -1,4 +1,4 @@
-import { useCallback } from 'react';
+import { useCallback, useRef } from 'react';
 
 import useTranslation from 'next-translate/useTranslation';
 import useSWR, { useSWRConfig } from 'swr';
@@ -30,19 +30,11 @@ interface UseBookmarkCollectionsProps {
 
 interface UseBookmarkCollectionsReturn {
   collectionIds: string[];
-  isLoading: boolean;
   addToCollection: (collectionId: string) => Promise<boolean>;
   removeFromCollection: (collectionId: string) => Promise<boolean>;
   mutateBookmarkCollections: (newIds?: string[]) => void;
 }
 
-/**
- * Custom hook for managing which collections a bookmark belongs to
- * Handles adding/removing bookmarks from collections with proper cache invalidation
- *
- * @param {UseBookmarkCollectionsProps} props - Hook configuration
- * @returns {UseBookmarkCollectionsReturn} Bookmark collection state and management functions
- */
 const useBookmarkCollections = ({
   mushafId,
   key,
@@ -54,21 +46,13 @@ const useBookmarkCollections = ({
   const { mutate: globalMutate } = useSWRConfig();
   const toast = useToast();
   const { t } = useTranslation('common');
+  const isPendingRef = useRef(false);
 
-  const {
-    data: collectionIds,
-    isValidating,
-    mutate: mutateBookmarkCollections,
-  } = useSWR<string[]>(
+  const { data: collectionIds, mutate: mutateBookmarkCollections } = useSWR<string[]>(
     isLoggedIn ? makeBookmarkCollectionsUrl(mushafId, key, type, verseNumber) : null,
-    async () => {
-      const response = await getBookmarkCollections(mushafId, key, type, verseNumber);
-      return response;
-    },
+    async () => getBookmarkCollections(mushafId, key, type, verseNumber),
     mutatingFetcherConfig,
   );
-
-  const isLoading = isValidating && !collectionIds;
 
   const showErrorToast = useCallback(
     (err: unknown) => {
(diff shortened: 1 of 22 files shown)


src/hooks/useBookmarkCollections.ts
```diff
@@ -1,8 +1,7 @@
 import { useCallback } from 'react';
 
 import useTranslation from 'next-translate/useTranslation';
-import { useSWRConfig } from 'swr';
-import useSWRImmutable from 'swr/immutable';
+import useSWR, { useSWRConfig } from 'swr';
```
osamasayed, line 4: Any reason why we no longer use immutable?

src/hooks/usePageBookmark.ts
```diff
@@ -1,7 +1,7 @@
 import { useCallback, useMemo } from 'react';
 
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
-import useSWRImmutable from 'swr/immutable';
+import useSWR from 'swr';
```
osamasayed, line 4: Ditto: any reason why we stopped using useImmutable?

src/hooks/useVerseBookmark.ts
```diff
@@ -1,8 +1,7 @@
 import { useCallback, useMemo } from 'react';
 
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
-import { useSWRConfig } from 'swr';
-import useSWRImmutable from 'swr/immutable';
+import useSWR, { useSWRConfig } from 'swr';
```
osamasayed, line 4: Ditto: any reason why we stopped using useImmutable?

src/hooks/usePageBookmark.ts
AhmedCodeGuy: # Why `useSWR` over `useSWRImmutable` for Bookmarks

## TL;DR

Bookmarks are **mutable, user-specific data** that can change from multiple pages. `useSWR` provides automatic cache invalidation and cross-tab sync out of the box, while `useSWRImmutable` would require manual implementation of these features.

## What's the difference?

`useSWRImmutable` is `useSWR` with these options disabled:

```typescript
{
  revalidateIfStale: false,
  revalidateOnFocus: false,
  revalidateOnReconnect: false
}
```

## Can `useSWRImmutable` achieve the same behavior?

Yes, but it requires manual handling:

| Feature | `useSWR` | `useSWRImmutable` |
|---------|----------|-------------------|
| Cache invalidation | `mutate(key)` triggers refetch | Must pass exact new data or `mutate(key, undefined, { revalidate: true })` |
| Cross-tab sync | Automatic via `revalidateOnFocus` | Manual (BroadcastChannel API or localStorage events) |
| Reconnect handling | Automatic | Manual |

## Comparison

| Aspect | `useSWR` | `useSWRImmutable` |
|--------|----------|-------------------|
| Network requests | More (revalidates on focus/reconnect) | Fewer |
| Cross-tab sync | Automatic | Manual implementation |
| Code complexity | Simpler | More complex |
| Risk of stale data | Low | Higher if you miss an update |
| Best for | Mutable, user-specific data | Static/immutable data |

## Why `useSWR` is right for bookmarks

1. **Mutable data** - Bookmarks change frequently from multiple sources (reader, profile, collections)
2. **Cross-tab sync is required** - Users expect changes in one tab to reflect in another (QF-3930)
3. **Consistency over fewer requests** - Bookmarks are critical user data; stale state causes confusion
4. **Minimal overhead** - Extra requests only happen on tab focus/reconnect, not continuous polling

## When to use `useSWRImmutable`

Use it for truly immutable content:

- Quran text
- Translations
- Tafsir
- Static configuration

## References

- [SWR Documentation](https://swr.vercel.app/docs/revalidation)
- [useSWRImmutable](https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations)

AhmedCodeGuy: # Reducing Network Requests with `useSWR`

```typescript
useSWR(key, fetcher, {
  revalidateOnReconnect: false, // Skip reconnect revalidation
  dedupingInterval: 10000,      // 10s deduping window
  // Keep revalidateOnFocus: true for cross-tab sync
});
```

Combined with **optimistic updates**:

- Updates UI immediately
- Passes new data to cache with `{ revalidate: false }`
- Only refetches on error rollback

This gives cross-tab sync while minimizing requests.

AhmedCodeGuy: Also added `focusThrottleInterval: 30000` - to limit focus revalidations to once per 30s.

PR #2641, opened 2025-12-08
Fix: Page number tracking in Reading and Translation modes

AhmedCodeGuy:
## Summary

Closes: [QF-3925](https://quranfoundation.atlassian.net/browse/QF-3925)

Fixed page number not updating immediately when navigating to different content (chapters, juz, pages, hizb, etc.). Previously, users had to scroll before the page number would update in the context menu.

## Problem

The intersection observer's `rootMargin` (`-131.6px 0px -68% 0px`) excludes elements at the top of the viewport, so the first verse is never "intersecting" on initial page load. This meant the page number only updated after scrolling.

## Solution

Set the page state directly from `initialData.verses[0]` on navigation, rather than relying on the intersection observer for initial state. The first verse from server-side data contains all required information: `verseKey`, `chapterId`, `pageNumber`, and `hizbNumber`.

### Key Changes

1. **`useSyncChapterPage.ts`** - Simplified to use `initialData.verses[0]` directly with `useBrowserLayoutEffect` for synchronous state update before paint

2. **`useSyncReadingProgress.ts`** - Added null guard for elements without `data-verse-key` attribute to prevent runtime errors

3. **`TranslationPageVerse.tsx`** - Removed obsolete `pageNumber` and `hizbNumber` props from ChapterHeader

## Coverage

This fix works for **all 98 navigation scenarios**:
- 7 content types (Surah, Verse, Juz, Page, Hizb, Rub, Range)
- 7 destination types (navigating between any of the above)
- 2 reading modes (Reading Mode, Translation Mode)

## Testing

- ✅ Page number updates immediately on navigation (no scroll required)
- ✅ Page number continues to update correctly while scrolling
- ✅ Works in both Reading and Translation modes
- ✅ TypeScript compilation passes
- ✅ ESLint passes
- ✅ No runtime errors

[QF-3925]: https://quranfoundation.atlassian.net/browse/QF-3925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/QuranReader/ReadingView/Line.tsx
+++ b/src/components/QuranReader/ReadingView/Line.tsx
@@ -1,10 +1,9 @@
-import { memo, RefObject, useCallback, useContext, useEffect, useRef } from 'react';
+import { memo, RefObject, useContext, useEffect } from 'react';
 
 import { useSelector as useXstateSelector } from '@xstate/react';
 import classNames from 'classnames';
 import { shallowEqual, useSelector } from 'react-redux';
 
-import { QURAN_READER_OBSERVER_ID } from '../observer';
 import { verseFontChanged } from '../utils/memoization';
 
 import styles from './Line.module.scss';
@@ -12,7 +11,6 @@ import styles from './Line.module.scss';
 import ChapterHeader from '@/components/chapters/ChapterHeader';
 import { useOnboarding } from '@/components/Onboarding/OnboardingProvider';
 import VerseText from '@/components/Verse/VerseText';
-import useIntersectionObserver from '@/hooks/useObserveElement';
 import useScroll, { SMOOTH_SCROLL_TO_CENTER } from '@/hooks/useScrollToElement';
 import { selectEnableAutoScrolling } from '@/redux/slices/AudioPlayer/state';
 import { selectInlineDisplayWordByWordPreferences } from '@/redux/slices/QuranReader/readingPreferences';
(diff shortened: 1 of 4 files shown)


src/components/QuranReader/ReadingView/Line.tsx
```diff
@@ -40,6 +42,22 @@ const Line = ({ lineKey, words, isBigTextLayout, pageIndex, lineIndex }: LinePro
   const [scrollToSelectedItem, selectedItemRef]: [() => void, RefObject<HTMLDivElement>] =
     useScroll(SMOOTH_SCROLL_TO_CENTER);
 
+  // Register with intersection observer for page tracking
+  const observerRef = useRef<HTMLDivElement | null>(null);
+  useIntersectionObserver(observerRef, QURAN_READER_OBSERVER_ID);
+
+  // Merge refs for both auto-scroll and observer
+  const mergedRef = useCallback(
```
mzunain: please follow DRY. I found ditto same in src/components/QuranReader/TranslationView/TranslationViewCell.tsx

src/components/QuranReader/TranslationView/TranslationViewCell.tsx
```diff
@@ -71,9 +73,30 @@ const TranslationViewCell: React.FC<TranslationViewCellProps> = ({
   const translationsCount = verse.translations?.length || 0;
   const wordVerse: WordVerse = constructWordVerse(verse, translationsLabel, translationsCount);
 
+  // Register this cell with the global intersection observer for page tracking
+  const observerRef = useRef<HTMLDivElement | null>(null);
+  useIntersectionObserver(observerRef, QURAN_READER_OBSERVER_ID);
+
+  // Callback ref to merge both selectedItemRef and observerRef
+  const mergedRef = useCallback(
```
mzunain: same as above I mentioned

mzulqarnain118 (changes requested):

src/components/QuranReader/ReadingView/Line.tsx
AhmedCodeGuy: Good catch! JAK.

mzulqarnain118 (approved):

PR #2673, opened 2025-12-17
fix(markdown-editor): increase blockquote line-height for Urdu RTL

basit3407:
- Increase blockquote paragraph line-height for Urdu RTL pages to prevent glyph collisions in large Urdu text.
- This applies a language- and direction-specific override without affecting LTR or non-Urdu content.


basit3407:
@osamasayed it's been some time since this was created. Any comments ?

osamasayed (approved):

PR #2703, opened 2025-12-23
add flexbox mixin

afifvdin:
add flexbox mixin

### DIFF
--- a/src/styles/mixins/_flexbox.scss
+++ b/src/styles/mixins/_flexbox.scss
@@ -1,0 +1,248 @@
+// use it like
+// @use 'src/styles/mixins/_flexbox' as flexbox-mixins;
+// @include flexbox-mixins.flex-col;
+
+/* Define mixins */
+@mixin flex {
+  display: flex;
+}
+
+@mixin col-center {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
+
+@mixin col-end {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+}
+
+@mixin col-start {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+}
+
+@mixin inline-flex {
+  display: inline-flex;
+}
+
+@mixin flex-col {
+  display: flex;
+  flex-direction: column;
+}
+
+@mixin flex-row {
+  display: flex;
+  flex-direction: row;
+}
+
+@mixin justify-center {
+  justify-content: center;
+}
+
+@mixin justify-start {
+  justify-content: flex-start;
+}
+
+@mixin justify-end {
+  justify-content: flex-end;
+}
+
+@mixin justify-between {
+  justify-content: space-between;
+}
+
+@mixin justify-around {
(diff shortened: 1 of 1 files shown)


mzulqarnain118 (approved):

AhmedCodeGuy (changes requested):
## PR #2703 - Code Review

### Summary
Adds a flexbox mixin file with 248 lines of mixins and utility classes. However, this PR has significant issues that need to be addressed before it can be merged.

---

### 🔴 CRITICAL ISSUES

#### 1. Dead Code - No Usage Anywhere
The PR only adds a file but **nothing uses it**. This is dead code that adds to the bundle without providing any value. Before adding utility code, there should be at least one component using it.

#### 2. Minimal Value Mixins
Most mixins wrap single CSS properties, providing no benefit over writing CSS directly:

```scss
@mixin justify-center {
  justify-content: center;
}

@mixin align-start {
  align-items: flex-start;
}
```

**Why this is problematic:**
- Adds indirection without benefit
- Developers must learn custom mixin names instead of standard CSS
- No compile-time optimization gained
- Makes code harder to read and maintain

**Compare to existing useful mixins in `_utility.scss`:**
```scss
@mixin center-vertically {
  display: flex;
  flex-direction: column;
  justify-content: center;
}
```
This is useful because it combines multiple properties into a semantic unit.

---

### 🟠 MEDIUM ISSUES

#### 3. Duplicate Functionality
The codebase already has `@mixin center-vertically` in `src/styles/_utility.scss:13-17`. Adding `@mixin col-center` creates duplication.

#### 4. CSS Classes in a Mixin File
The file mixes two concepts:
- SCSS mixins (`@mixin flex-col`)
- Utility classes (`.flexCol`, `.rowCenter`)

These should be separate files. Also, utility classes won't work globally with CSS modules - each component would need to import and reference them explicitly.

#### 5. Architectural Mismatch
The codebase uses **CSS modules with component-specific styles**. Adding Tailwind-like utility classes is an anti-pattern here. If you need shared flexbox patterns, use mixins that components can `@include`.

---

### 🟡 LOW ISSUES

#### 6. Inconsistent Naming Convention
```scss
// With flex- prefix
@mixin flex-col
@mixin flex-row
@mixin flex-wrap

// Without prefix
@mixin col-center
@mixin row-center
@mixin justify-center
```

#### 7. Missing Newline at End of File

#### 8. No PR Description
The PR body just says "add flexbox mixin" with no context on:
- Why is this needed?
- Which components will use it?
- What problem does it solve?

---

### 🎯 VERDICT

**Do Not Merge** - This PR needs fundamental rethinking:

1. **Remove single-property mixins** - They add no value
2. **Remove utility classes** - They don't fit the CSS modules architecture
3. **Keep only compound mixins** that combine multiple properties semantically
4. **Show usage** - At least one component should use these mixins
5. **Avoid duplicating** existing mixins in `_utility.scss`

**Suggested approach:** If flexbox utilities are needed, extend `_utility.scss` with a few well-designed compound mixins, and refactor at least one component to use them as proof of value.

afifvdin:
i'll close this, as we will add utilities when we need it

PR #3278, opened 2026-03-30
Explore: refine article detail layout and card metadata display

RooTooRD:
- Update explore article detail UX: change back button text to `Back to Explore`
- Move page title above the hero image and avoid duplicate top heading from article body
- Improve explore cards: fallback title from slug, sanitize/truncate description text, and hide description when missing
- Add description spacing only when description exists


RooTooRD:
Closing this PR because it targets the wrong upstream repository.

PR #3294, opened 2026-05-14
fix: add missing @/styles path alias to storybook's webpack

junaidanjum:
## summary

storybook's webpack alias list in `.storybook/main.js` was missing `@/styles`. scss modules using `@use '@/styles/...'` (e.g. `dls/Button`) failed with `SassError: Can't find stylesheet to import`, breaking the storybook preview build.

added `'styles'` to the `otherAliases` array. now matches how next.js resolves the alias via tsconfig paths.

## type of change

- [x] bug fix

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -33,7 +33,7 @@ module.exports = {
     ];
 
     // @/ root alias doesn't work in storybook, so we have to write the aliases manually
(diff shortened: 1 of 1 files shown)


Mir-Zairan:
I think same issue is already resolved in https://github.com/quran/quran.com-frontend-next/pull/3272

PR #3316, opened 2026-07-11
Fix missing reading colors in qiraat ReadersPanel when reader overlaps readings

### LINKED ISSUE #3286: Qiraat reader panel hides colors when reader mappings overlap
## Problem

On `https://quran.com/2:245/qiraat`, the junction `فَيُضَـٰعِفَهُۥ` shows four color-coded reading variants in the reading cards, but the left-side reader/transmitter panel only shows three colors. The pink variation is present in the reading cards but absent from the reader panel.

## Reproduction

1. Open `https://quran.com/2:245/qiraat`.
2. Select the junction for `فَيُضَـٰعِفَهُۥ`.
3. Compare the reading-card colors with the reader/transmitter tags on the left panel.

## Actual Behavior

The reading cards show four colors: white, green, blue, and pink.

The left panel shows green and blue correctly, but Abū Jaʿfar and Ibn Kathīr remain white. This causes the pink reading `فَيُضَعِّفُهُ` to have no visible reader/transmitter association in the left panel.

## Expected Behavior

For this junction, the left panel should map the colors as follows:

```txt
green: Ibn ʿĀmir, Yaʿqūb
blue:  Nāfiʿ, Ḥamzah, Khalaf, Abū ʿAmr, al-Kisāʾī
pink:  Abū Jaʿfar, Ibn Kathīr
white: ʿĀṣim only
```

## Root Cause

The live payload has reader-level overlap and no transmitter-level disambiguation for this junction. `matrix.transmitters` is empty for the readings, so the frontend falls back to reader-level matching.

For `2:245`, the live data includes:

```txt
white فَيُضَاعِفَهُ readers: [5, 8, 6]
pink  فَيُضَعِّفُهُ readers: [8, 6]
```

Reader `6` is Abū Jaʿfar and reader `8` is Ibn Kathīr, so both appear in both the white and pink readings.

The frontend currently uses first-match lookup:

```ts
readings.find(({ matrix }) => matrix?.readers?.includes(reader.id))
```

Because the white reading appears before the pink reading, Abū Jaʿfar and Ibn Kathīr are assigned white and the lookup never reaches pink.

The same first-match assumption also exists in the transmitter click/scroll fallback.

## Scope

A live scan of qiraat payloads found this is not isolated to `2:245`. The same ambiguous reader-level mapping pattern appears in these verses:

```txt
2:245
6:145
7:161
10:35
12:23
18:44
19:25
20:63
22:39
```

A related duplicate transmitter-level first-match case appears in:

```txt
7:165
```

## Proposed Frontend Mitigation

Keep direct transmitter matches authoritative when `matrix.transmitters` is populated.

For reader-level fallback, do not stop at the first matching reading when the reader appears in multiple reading colors. Instead:

1. Build all candidate readings/colors for the reader in the selected junction.
2. Assign unambiguous reader/transmitter colors first.
3. Track which reading colors are already represented in the left panel.
4. For an ambiguous reader, if the first matching candidate color is already represented and a later candidate color is not represented yet, continue to the later candidate and use that color.
5. If all candidate colors are already represented, fall back to the current first-match behavior.

This would make `2:245` resolve correctly: ʿĀṣim keeps white, while Abū Jaʿfar and Ibn Kathīr continue past the already-represented white candidate and map to the otherwise missing pink candidate.

## Longer-Term Data Fix

The cleanest fix would be for the API payload to provide transmitter-level disambiguation or avoid overlapping reader-level mappings where the UI expects a unique reader-to-reading color. The frontend mitigation above prevents missing colors in the panel while preserving the current payload shape.

Mubashir78:
Fixes #3286

## Summary

When a junction has reader-level overlap (a reader appearing in multiple reading colors) and no transmitter-level disambiguation, the Readers panel hides some reading colors because the color lookup uses first-match semantics via `readings.find()`.

For example, on `https://quran.com/2:245/qiraat`, readers Abū Jaʿfar (6) and Ibn Kathīr (8) appear in both the white reading and the pink reading. Because the white reading appeared first in the payload, both readers were assigned white, hiding the pink color entirely from the panel.

## Changes

**ReadersPanel/index.tsx**: Added a `readerColorMap` that:
1. Assigns unambiguous readers (matching exactly one reading) first
2. For ambiguous readers, prefers a candidate color not yet used by another reader
3. Falls back to first-match only when all candidate colors are already represented

**ReaderItem.tsx**: Added an optional `readerColor` prop. When provided, it takes priority over the inherited first-match fallback in `getTransmitterColor`, allowing the parent panel to control the color assignment.

## Result

All reading colors now appear in the Readers panel even when readers overlap across multiple readings. The fix is not specific to 2:245 and applies to all verses with the same ambiguous reader-level mapping pattern (6:145, 7:161, 10:35, 12:23, 18:44, 19:25, 20:63, 22:39).

### DIFF
--- a/src/components/QuranReader/ReadingView/StudyModeModal/tabs/StudyModeQiraatTab/ReadersPanel/ReaderItem.tsx
+++ b/src/components/QuranReader/ReadingView/StudyModeModal/tabs/StudyModeQiraatTab/ReadersPanel/ReaderItem.tsx
@@ -10,6 +10,7 @@ interface ReaderItemProps {
   reader: QiraatReader;
   transmitters: QiraatTransmitter[];
   readings: QiraatReading[];
+  readerColor?: string;
   onInfoClick?: () => void;
   onTransmitterClick?: (transmitterId: number) => void;
   isClickable?: boolean;
@@ -26,6 +27,7 @@ const ReaderItem: React.FC<ReaderItemProps> = ({
   reader,
   transmitters,
   readings,
+  readerColor,
   onInfoClick,
   onTransmitterClick,
   isClickable = false,
@@ -57,12 +59,14 @@ const ReaderItem: React.FC<ReaderItemProps> = ({
 
     if (transmitterReading) return transmitterReading.color || DEFAULT_COLOR;
 
-    // 2. If not found directly, the transmitter inherits color from its parent reader
-    // Find a reading where this reader appears in the matrix
+    // 2. If the parent component has pre-computed a color for this reader, use it
+    if (readerColor) return readerColor;
+
+    // 3. If not found directly, the transmitter inherits color from its parent reader
     const readerReading = readings.find(({ matrix }) => matrix?.readers?.includes(reader.id));
     if (readerReading) return readerReading.color || DEFAULT_COLOR;
 
-    // 3. Default fallback - no association found
+    // 4. Default fallback - no association found
     return DEFAULT_COLOR;
   };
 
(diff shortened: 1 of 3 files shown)


Mubashir78:
Fixed.

Issue #2208, opened 2024-09-23
[bug]: Security policy is missing


rehandaphedar:
### Is there an existing issue for this?

- [X] There is no exsiting issue for this bug

### Describe the bug

Not sure where to post this, but the [create an issue](https://github.com/quran/quran.com-frontend-next/issues/new/choose) page links to a Security Policy, which is missing.

### Steps to reproduce

NA

### Environment (please complete the following information) and Add any other context about this bug

NA

### Environment

Production(quran.com)

successbyte:
Salamu Alaikum brother
would you please assign me to this issue

PR #3324, opened 2026-08-15
docs: add security policy (#2208)

wakqasahmed:
@/tmp/quran-issue-2208-prbody.md

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -2,16 +2,29 @@
 
 ## Supported Versions
 
-Only the latest version deployed on [quran.com](https://quran.com) (the `production` branch) is supported with security updates.
+Only the latest version deployed on [quran.com](https://quran.com) (the `production` branch) is
+supported with security updates. Tagged releases (e.g. `v0.0.12`) are not separately patched; if a
+fix is needed, it lands on `production`.
 
 ## Reporting a Vulnerability
 
-If you discover a security vulnerability in this project, please **do not** open a public GitHub issue.
+If you discover a security vulnerability in this project, please **do not** open a public GitHub
+issue.
 
-Instead, report it privately using GitHub's [private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability) feature:
+[Private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+is not yet enabled on this repository. Until it is, please report vulnerabilities directly to a
+maintainer:
 
-1. Go to the [Security tab](https://github.com/quran/quran.com-frontend-next/security) of this repository.
-2. Click "Report a vulnerability" to open a draft security advisory.
-3. Provide as much detail as possible (steps to reproduce, affected versions, potential impact).
+1. Open a new issue titled `Security: request for private contact` with **no vulnerability details**
+   in the body, just a note that you have a security report to share privately.
+2. A maintainer will respond on that issue to arrange a private channel (e.g. email or the
+   [Quran.com Discord](https://discord.gg/SpEeJ5bWEQ)) for the actual report.
 
-We will review the report and respond as soon as possible.
+Once private vulnerability reporting is enabled for this repository, this document will be updated
+to point directly to the Security tab's "Report a vulnerability" flow instead.
+
+When reporting, please include as much detail as possible: steps to reproduce, affected versions,
+and potential impact.
+
+A maintainer will acknowledge your report and start triage. Response times are best-effort, as this
+is not a project with a contracted security SLA.
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,1 +1,17 @@
-https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository
+# Security Policy
+
+## Supported Versions
+
+Only the latest version deployed on [quran.com](https://quran.com) (the `production` branch) is supported with security updates.
+
+## Reporting a Vulnerability
+
+If you discover a security vulnerability in this project, please **do not** open a public GitHub issue.
+
+Instead, report it privately using GitHub's [private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability) feature:
+
+1. Go to the [Security tab](https://github.com/quran/quran.com-frontend-next/security) of this repository.
+2. Click "Report a vulnerability" to open a draft security advisory.
+3. Provide as much detail as possible (steps to reproduce, affected versions, potential impact).
+
+We will review the report and respond as soon as possible.


.github/SECURITY.md
```diff
@@ -1 +1,17 @@
-https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository
\ No newline at end of file
+# Security Policy
+
+## Supported Versions
+
+Only the latest version deployed on [quran.com](https://quran.com) (the `production` branch) is supported with security updates.
```
wakqasahmed: This line is 128 characters. `.prettierrc` sets `printWidth: 100` + `proseWrap: "always"`, and lint-staged runs `prettier --write --ignore-unknown` over `"*"`, so Markdown **is** covered — contrary to the PR description. `prettier --check` fails on this file. Please run Prettier.

Separately, the claim itself: the repo does publish tags (`v0.0.12`, `v0.0.11`, …). Worth saying explicitly that tagged releases are **not** separately patched and only the deployed `production` branch is, so the sentence can't be misread as "the latest tag is supported".

.github/SECURITY.md
```diff
@@ -1 +1,17 @@
...
+
+## Reporting a Vulnerability
+
+If you discover a security vulnerability in this project, please **do not** open a public GitHub issue.
```
wakqasahmed: 103 characters — over the configured `printWidth: 100`. Prettier would rewrap this.

.github/SECURITY.md
```diff
@@ -1 +1,17 @@
...
+
+Instead, report it privately using GitHub's [private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability) feature:
```
wakqasahmed: **Blocking.** This directs reporters to GitHub's private vulnerability reporting, but that feature is switched off for this repository:

```
$ gh api repos/quran/quran.com-frontend-next/private-vulnerability-reporting
{"enabled":false}
```

So the documented channel does not currently exist. Also 227 chars on one line — over `printWidth: 100`.

.github/SECURITY.md
```diff
@@ -1 +1,17 @@
...
+
+1. Go to the [Security tab](https://github.com/quran/quran.com-frontend-next/security) of this repository.
+2. Click "Report a vulnerability" to open a draft security advisory.
```
wakqasahmed: With PVR disabled, there is no "Report a vulnerability" button rendered on the Security tab, so this step cannot be completed. A reporter who reaches this dead end will most plausibly fall back to filing a public issue — the exact outcome the policy above forbids.

This PR can't fix that itself; a maintainer has to enable Settings → Security → "Private vulnerability reporting". Please either get that confirmed before merge, or add an email fallback so the policy is actionable on day one. (Line 13 is also 106 chars — over `printWidth: 100`.)

.github/SECURITY.md
```diff
@@ -1 +1,17 @@
...
+3. Provide as much detail as possible (steps to reproduce, affected versions, potential impact).
+
+We will review the report and respond as soon as possible.
```
wakqasahmed: "as soon as possible" is unenforceable and gives a reporter nothing to plan around. Consider a concrete target (e.g. initial acknowledgement within N business days) and naming the responsible party, since the whole value of a security policy is telling a reporter what to expect before they decide whether to disclose privately.

wakqasahmed (commented):
## Cold-start review — changes requested

I verified the diff against the repo's actual configuration rather than taking the PR description at face value. The change is directionally right (the old `.github/SECURITY.md` was a bare docs URL, so `/security/policy` rendered no policy at all — which is exactly what #2208 reported), file location is correct, and the title matches `semantic.yml`. But there is one blocking accuracy problem and one CI/hygiene problem.

**1. BLOCKING — the reporting flow this file documents does not exist on this repo.**

Private vulnerability reporting is currently **disabled**:

```
$ gh api repos/quran/quran.com-frontend-next/private-vulnerability-reporting
{"enabled":false}
```

With PVR off, there is no "Report a vulnerability" button on the Security tab, so steps 1–3 of the new policy are a dead end. A reporter who follows them and finds nothing will most likely fall back to opening a public issue — precisely what the policy tells them not to do. That makes this change arguably worse than the status quo for the one scenario it exists to handle.

A PR cannot flip that toggle; it is a repo setting (Settings → Security → "Private vulnerability reporting"). So this needs one of:
- a maintainer enabling PVR, with the PR held until that is confirmed, or
- a real contact channel (e.g. a security@ address owned by Quran Foundation) documented as the primary or fallback route.

Either way the PR description should state which, so a reviewer isn't merging an unverifiable claim.

**2. Prettier — the PR description's lint claim is incorrect and the file is unformatted.**

The PR says "Markdown-only file not covered by lint". `package.json` lint-staged is:

```json
"lint-staged": {
  "*.{tsx,ts}": "eslint --fix",
  "*.scss": "stylelint --fix",
  "*": "prettier --write --ignore-unknown"
}
```

The `"*"` glob covers Markdown, and `.prettierrc` sets `printWidth: 100` with `proseWrap: "always"`. The file as committed fails the check:

```
$ npx prettier@2 --print-width 100 --prose-wrap always --check .github/SECURITY.md
[warn] SECURITY.md
[warn] Code style issues found in the above file. Forgot to run Prettier?
```

Four lines exceed 100 chars (128, 103, 227, 106). The pre-commit hook would normally have rewrapped these, so it looks like it was bypassed. Please run Prettier and re-commit, and correct the claim in the PR body.

**3. Non-blocking**
- The PR body drops the template's `Closes: [QF-XXXX]` Jira link and several Pre-Review checklist items (`yarn test`, `yarn build`). Trivially N/A for a docs change, but the repo template asks for the Jira reference.
- No scope statement (are `api.quran.com`, the mobile apps, and sibling `quran/*` repos in or out of scope for this policy?) and no safe-harbour/disclosure-expectations wording. Worth adding while the file is being written, not worth blocking on.

Happy to re-review once PVR is confirmed enabled (or a contact address is added) and Prettier has been run.

wakqasahmed:
Thanks for the thorough review — addressed all points in `e6f50c46`:

1. **PVR disabled (blocking):** confirmed `{"enabled":false}` still holds. I don't have the org permission to flip that repo setting, and there's no existing security@ address anywhere in the repo/README/org profile to use as a drop-in replacement, so I didn't invent one. Instead the policy now documents a real fallback: report via a contact-only issue (no vulnerability details in the body) and a maintainer arranges a private channel from there, with a note to switch back to GitHub's native PVR flow once it's enabled. Flagging that enabling PVR itself is a maintainer/settings action outside this PR's scope.
2. **Prettier:** ran `npx prettier@2 --print-width 100 --prose-wrap always --write .github/SECURITY.md` against the repo's actual `.prettierrc`; `--check` now passes. Corrected the PR description's incorrect claim that Markdown isn't covered by lint-staged.
3. **Vague response wording:** replaced "as soon as possible" with the concrete process above (maintainer acknowledges + triages, best-effort since there's no contracted SLA) rather than inventing a numeric SLA the maintainers haven't agreed to.
4. **Supported versions:** now explicitly states tagged releases (e.g. `v0.0.12`) are not separately patched — only the deployed `production` branch is supported.
5. **PR template compliance (non-blocking):** added `Closes: N/A (not a Jira-tracked change)` and filled in the Pre-Review checklist items reasonably (test/build marked N/A for a docs-only change).

Scope/safe-harbour wording (mentioned as non-blocking extra) intentionally left out of this pass to keep the diff minimal — happy to add in a follow-up if wanted.

wakqasahmed:
Hi @yousefdergham @osamasayed — noticed this PR doesn't have a reviewer assigned yet — it's been about 5 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

wakqasahmed:
Hi @yousefdergham @osamasayed — circling back on this one — still no reviewer assigned, about 7 days since the last activity. No pressure, just don't want it to get lost.

PR #1741, opened 2022-08-08
Start xstate implementation

muhajirdev:
TODO

- [x] remaining count of the repeat butotn
- [x] change reciter from UI
- [x] fix broken localstorage persistance
- [x] play audio without changing reciter (need to update xstate implementation)
- [x] remove unused selectors  & actions from redux
- [x] persist reciter data & playback rate to DB
- [x] Inifinite repeat settings
- [x] Load reciterId from database instead of localstorage and save to database via (onSettingsChangeXstate)
- [x] Optimize rerendering, replace useActor with useSelector
- [x] update typing file, use the existing one
- [x] fetch data using existing fetcher -> the implication is we'll use verseTiming instaed of verse_timing, verseKey instead of verse_key

PR #3325, opened 2026-08-15
Fix stale filter value on repeat-ayah combobox open (#2088)

### LINKED ISSUE #2088: [bug]: Cannot choose ayah past 19 in repeat settings

### Describe the bug

When I attempt to set the starting ayah, I cannot type in the box, and the dropdown options only include ayahs 1 to 19, with some missing, depending on the surah. I have seen this issue in multiple surahs, across multiple browsers and systems (Chrome on ChromeOS, Edge on Windows 11), all up to date, for both the main and beta versions of the website.

### Relevant log output

_No response_

### Steps to reproduce

1. Open any surah on the website
2. Open repeat settings
3. Attempt to type in any of the boxes or open the dropdown menu

ChromeOS, Chrome v121.0.6167.159 (Official Build) (64-bit)
Windows 11, Edge v121.0.2277.128 (Official build) (64-bit)

### Environment

Production(quran.com)

wakqasahmed:
## Summary

An earlier version of this PR added a test-only regression check for `generateChapterVersesKeys()` and claimed the bug reported in #2088 (repeat-settings "to ayah" dropdown capped around ayah 19) had already been fixed by the 2022 xstate rewrite of the repeat modal. That claim was wrong, and a cold-start review on this PR traced it down: the 2022 rewrite (`c46b62df`) happened ~18 months *before* the issue was even filed, and `generateChapterVersesKeys` was already uncapped at the time of the report too. That function was never the cause.

This update fixes the actual root cause and corrects the PR accordingly.

## Real root cause

`src/components/dls/Forms/Combobox/index.tsx` seeds `inputValue` from `initialInputValue` on mount, and filters the dropdown's visible items by a **substring** match against `inputValue`:

```ts
const [inputValue, setInputValue] = useState<string>(initialInputValue || '');
...
setFilteredItems(
  !inputValue ? items : items.filter((item) => item.label.toLowerCase().includes(inputValue.toLowerCase())),
);
```

`SelectRepetitionMode.tsx` passes `initialInputValue={toLocalizedVerseKey(verseKey, lang)}` and `AdvancedCopy/SelectorContainer.tsx` passes `initialInputValue={value}` — i.e. the currently selected verse key (e.g. `"36:1"` for a surah whose selection starts at ayah 1). Opening the combobox (`onSelectorClicked`) only toggled `isOpened`; it never cleared `inputValue`. So the dropdown stayed filtered by the stale selected value, and only items whose label *contained* `"36:1"` as a substring matched: `36:1, 36:10, 36:11 … 36:19`. Ayahs 2-9 and 20+ were invisible — exactly the reported "capped around 19, with some missing depending on the surah."

## Fix

`onSelectorClicked` now clears `inputValue` when the combobox transitions from closed to open (single-select only; multi-select already cleared it on close), so every item is visible again until the user actually types to filter. `initialInputValue` still drives the closed-state display label (via `closeCombobox`'s restore-from-`valueToLabelMap` logic), so nothing about the collapsed view changes.

Verified against both a long surah (Ya-Sin, 83 verses, selection seeded at `"36:1"`) and a short surah (An-Nas, 6 verses) to confirm the short case isn't broken.

## What changed here

- `src/components/dls/Forms/Combobox/index.tsx` — clear `inputValue` on open (the actual fix).
- `src/components/dls/Forms/Combobox/index.test.tsx` — new component-level regression test: renders `Combobox` seeded with a stale selected value, opens it, and asserts ayahs outside the old substring match (e.g. `36:2`, `36:20`, `36:83`) are present; also covers the short-surah case and confirms typing still filters correctly. Verified these tests fail without the fix and pass with it.
- `src/utils/verse.test.ts` — kept the earlier long-surah coverage for `generateChapterVersesKeys` (harmless, just not the actual fix), but removed the false "regression test for #2088" framing and tightened the assertion to a full expected-array comparison per review feedback.

## Test plan

- [x] `npx vitest run src/components/dls/Forms/Combobox src/utils/verse.test.ts` — 26/26 pass
- [x] Confirmed the new Combobox tests fail on the pre-fix code and pass after the fix
- [x] `npx eslint` clean on all changed files
- [x] Manually traced both `SelectRepetitionMode.tsx` (repeat-ayah modal) and `AdvancedCopy/SelectorContainer.tsx` (advanced copy range selectors) — both feed `Combobox` through the same `initialInputValue` prop, so both consumers are fixed by this change

Closes #2088

### DIFF
--- a/src/utils/verse.test.ts
+++ b/src/utils/verse.test.ts
@@ -198,4 +198,19 @@ describe('generateChapterVersesKeys', () => {
     const result = generateChapterVersesKeys(chaptersData, chapterId, false);
     expect(result).toEqual(expectedKeys);
   });
+
+  // Regression test for https://github.com/quran/quran.com-frontend-next/issues/2088
+  // The repeat-settings "to ayah" dropdown must include every ayah in the surah,
+  // not just the first 19, for surahs that have more than 19 verses.
+  it('should generate keys for every verse of a surah with more than 19 verses', async () => {
+    const chapterId = '2'; // Al-Baqarah has 286 verses
+    const chaptersData = await getAllChaptersData();
+    const result = generateChapterVersesKeys(chaptersData, chapterId, false);
+
+    expect(result).toHaveLength(286);
+    expect(result[0]).toEqual('2:1');
+    expect(result[18]).toEqual('2:19');
+    expect(result[19]).toEqual('2:20');
+    expect(result[result.length - 1]).toEqual('2:286');
+  });
 });
--- a/src/utils/verse.test.ts
+++ b/src/utils/verse.test.ts
@@ -199,18 +199,15 @@ describe('generateChapterVersesKeys', () => {
     expect(result).toEqual(expectedKeys);
   });
 
-  // Regression test for https://github.com/quran/quran.com-frontend-next/issues/2088
-  // The repeat-settings "to ayah" dropdown must include every ayah in the surah,
-  // not just the first 19, for surahs that have more than 19 verses.
+  // Coverage for a long surah: the existing tests above only exercised
+  // chapters 1, 112, and the not-found case, none of which have more than 19
+  // verses, so a hypothetical 19-item cap would not have been caught here.
   it('should generate keys for every verse of a surah with more than 19 verses', async () => {
     const chapterId = '2'; // Al-Baqarah has 286 verses
     const chaptersData = await getAllChaptersData();
     const result = generateChapterVersesKeys(chaptersData, chapterId, false);
+    const expectedKeys = Array.from({ length: 286 }, (unused, index) => `2:${index + 1}`);
 
-    expect(result).toHaveLength(286);
-    expect(result[0]).toEqual('2:1');
-    expect(result[18]).toEqual('2:19');
-    expect(result[19]).toEqual('2:20');
-    expect(result[result.length - 1]).toEqual('2:286');
+    expect(result).toEqual(expectedKeys);
   });
 });
--- a/src/components/dls/Forms/Combobox/index.test.tsx
+++ b/src/components/dls/Forms/Combobox/index.test.tsx
@@ -1,0 +1,113 @@
+import React from 'react';
+
+import { cleanup, fireEvent, render, screen, within } from '@testing-library/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { DropdownItem } from './ComboboxItem';
+
+import Combobox from '.';
+
+vi.mock('./Icons/SearchInputIcon', () => ({ default: () => <span /> }));
+vi.mock('./Icons/ClearInputIcon', () => ({ default: () => null }));
+vi.mock('./Icons/CaretInputIcon', () => ({ default: () => null }));
+
+// jsdom doesn't implement scrollIntoView; ComboboxItems calls it on open.
+window.HTMLElement.prototype.scrollIntoView = vi.fn();
+
+// Builds the same shape of items the repeat-settings / advanced-copy verse
+// selectors build via `generateChapterVersesKeys`: one entry per ayah, value
+// and label both being the localized verse key (e.g. "36:1").
+const buildVerseItems = (chapterNumber: number, versesCount: number): DropdownItem[] =>
+  Array.from({ length: versesCount }, (unused, index) => {
+    const verseKey = `${chapterNumber}:${index + 1}`;
+    return {
+      id: verseKey,
+      value: verseKey,
+      name: verseKey,
+      label: verseKey,
+    };
+  });
+
+describe('Combobox', () => {
+  afterEach(cleanup);
+
+  // Regression test for https://github.com/quran/quran.com-frontend-next/issues/2088
+  // Root cause: `inputValue` was seeded from `initialInputValue` (the currently
+  // selected verse key) and never cleared on open, so the substring filter kept
+  // matching only labels containing that stale value (e.g. selecting "36:1" then
+  // reopening only showed "36:1", "36:10"-"36:19").
+  it('shows every ayah of a long surah after opening, not just ones matching the pre-seeded selection', () => {
+    // Ya-Sin (36) has 83 verses; the reported bug reproduces when the
+    // currently selected verse is the surah's first ayah (e.g. "36:1"),
+    // since every ayah from 10-19 also contains "36:1" as a substring while
+    // 2-9 and 20+ do not.
+    const items = buildVerseItems(36, 83);
+
+    render(
+      <Combobox
+        id="single"
+        items={items}
+        value="36:1"
+        initialInputValue="36:1"
+        placeholder="Search..."
+      />,
+    );
+
+    // Before opening, the combobox is closed; open it the same way a user
+    // would, by clicking the selector.
+    fireEvent.click(screen.getByRole('combobox'));
+
+    const listbox = screen.getByRole('listbox');
+
+    // These ayahs never contained "36:1" as a substring, so they were hidden
+    // by the stale filter before the fix.
+    expect(within(listbox).getByText('36:2')).not.toBeNull();
+    expect(within(listbox).getByText('36:20')).not.toBeNull();
+    expect(within(listbox).getByText('36:83')).not.toBeNull();
+  });
+
+  it('still shows every ayah for a short surah after opening', () => {
+    // An-Nas (114) has 6 verses; make sure the fix doesn't break the short case.
+    const items = buildVerseItems(114, 6);
+
+    render(
+      <Combobox
+        id="single"
+        items={items}
+        value="114:1"
+        initialInputValue="114:1"
+        placeholder="Search..."
+      />,
+    );
+
+    fireEvent.click(screen.getByRole('combobox'));
+
+    const listbox = screen.getByRole('listbox');
+
+    expect(within(listbox).getByText('114:1')).not.toBeNull();
+    expect(within(listbox).getByText('114:6')).not.toBeNull();
+  });
+
+  it('still filters items once the user types after opening', () => {
+    const items = buildVerseItems(36, 83);
+
+    render(
+      <Combobox
+        id="single"
+        items={items}
+        value="36:1"
+        initialInputValue="36:1"
+        placeholder="Search..."
+      />,
+    );
+
+    const input = screen.getByRole('combobox');
+    fireEvent.click(input);
+    fireEvent.change(input, { target: { value: '36:5' } });
+
+    const listbox = screen.getByRole('listbox');
+
+    expect(within(listbox).getByText('36:5')).not.toBeNull();
+    expect(within(listbox).queryByText('36:20')).toBeNull();
+  });
+});
--- a/src/components/dls/Forms/Combobox/index.tsx
+++ b/src/components/dls/Forms/Combobox/index.tsx
@@ -231,7 +231,20 @@ const Combobox: React.FC<Props> = ({
   );
 
   const onSelectorClicked = () => {
-    setIsOpened((prevIsOpened) => !prevIsOpened);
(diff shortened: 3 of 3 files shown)


src/utils/verse.test.ts
```diff
@@ -198,4 +198,19 @@ describe('generateChapterVersesKeys', () => {
     const result = generateChapterVersesKeys(chaptersData, chapterId, false);
     expect(result).toEqual(expectedKeys);
   });
+
+  // Regression test for https://github.com/quran/quran.com-frontend-next/issues/2088
+  // The repeat-settings "to ayah" dropdown must include every ayah in the surah,
+  // not just the first 19, for surahs that have more than 19 verses.
+  it('should generate keys for every verse of a surah with more than 19 verses', async () => {
```
wakqasahmed, line 205: This comment asserts the framing I disagree with. `generateChapterVersesKeys` never limited the list to 19 — `src/utils/verse.ts@0d7f030e` (Nov 2023, the state of the file when #2088 was filed on 2024-02-21) already read `range(data[chapterNumberString].versesCount)`, and `RepeatAudioModal.tsx@e2c59b0e` (Apr 2023) already consumed it. Linking this test to #2088 as a *regression* test claims a causal relationship the history does not support.

Suggest either dropping the issue reference and keeping it as plain coverage ("long-surah case"), or wording it as "verify the key list is driven by `versesCount`" without asserting that a 19-cap ever lived here.

src/utils/verse.test.ts
```diff
@@ -198,4 +198,19 @@ describe('generateChapterVersesKeys', () => {
...
+    const chapterId = '2'; // Al-Baqarah has 286 verses
+    const chaptersData = await getAllChaptersData();
+    const result = generateChapterVersesKeys(chaptersData, chapterId, false);
+
+    expect(result).toHaveLength(286);
+    expect(result[0]).toEqual('2:1');
+    expect(result[18]).toEqual('2:19');
+    expect(result[19]).toEqual('2:20');
```
wakqasahmed: The assertions check length + four sampled positions, so they'd still pass on a mutation that produced 286 entries with a wrong/duplicated key somewhere in the middle (e.g. an off-by-one or a `Set`-dedup bug in the 20–285 range). Given the surrounding tests already use `toEqual` on the full expected array, the strongest and simplest form here is a full-range comparison:

```ts
const expectedKeys = Array.from({ length: 286 }, (_, i) => `2:${i + 1}`);
expect(result).toEqual(expectedKeys);
```

That subsumes all five current assertions (`result[18]`/`result[19]` are redundant once the full array is compared) and actually pins the contiguity the issue is about.

src/utils/verse.test.ts
```diff
@@ -198,4 +198,19 @@ describe('generateChapterVersesKeys', () => {
...
+    expect(result[result.length - 1]).toEqual('2:286');
+  });
 });
```
wakqasahmed, line 213: Three factual corrections for the PR description while you're here:

1. "the existing `generateChapterVersesKeys` tests only exercised Al-Fatihah (7 verses)" — the file also covers chapter `112` (4 verses) and the not-found `200` case. The substantive point (no surah with >19 verses was covered) stands, but the claim as written is inaccurate.
2. The stated pipeline lists `VersesRangeSelector.tsx` and `SelectorContainer.tsx` as if they were part of `RepeatAudioModal/`. They aren't — `SelectRepetitionMode.tsx` imports both from `@/components/Verse/AdvancedCopy/`. Worth stating precisely, since that shared component is where I believe the real bug lives.
3. "Edge Cases Verified: Surah with fewer than 19 ayahs (An-Nas, 6 verses)" is checked, but no An-Nas test is added in this diff — the existing short-surah test uses 112, not 114. Please don't tick a verification box for coverage the diff doesn't contain.

wakqasahmed (commented):
Cold-start review. I verified the code claims independently against `production` rather than taking the PR description at face value, and I think the premise of this PR is wrong in a way that matters: **#2088 was never caused by `generateChapterVersesKeys`, and the underlying bug looks like it still reproduces on `production` today.** Merging this would auto-close (`Closes #2088`) a live bug report on the strength of a test that cannot fail for it.

### What I confirmed is correct

- `src/utils/verse.ts:31-41` — `generateChapterVersesKeys` is `range(data[chapterNumberString]?.versesCount)`, no cap of 19 or anything else.
- `data/chapters/en.json` — chapter `2` really is `versesCount: 286` (and 1→7, 112→4, 114→6), so the new test's expectations are numerically right and it does exercise the real data source via `getAllChaptersData()` rather than a fixture. It is not tautological.
- No `slice`/limit in `src/components/dls/Forms/Combobox` either.

### Why the PR's narrative does not hold

The description says the bug "was already fixed as a side effect of the later rewrite of the repeat modal (move to the xstate-backed `RepeatAudioModal` + generic `Combobox`)". That rewrite is commit `c46b62df` "Start xstate implementation (#1741)", **August 2022** — ~18 months *before* the issue was filed (2024-02-21). I checked the tree as it stood when the issue was reported:

- `src/utils/verse.ts@0d7f030e` (Nov 2023) already had `range(data[chapterNumberString].versesCount)` — unbounded.
- `RepeatAudioModal.tsx@e2c59b0e` (Apr 2023, the last change to that directory before the report) already built `comboboxVerseItems` from `generateChapterVersesKeys`.

So the ayah list was already generated in full at the moment the user hit the bug. Nothing about that function was ever broken, and no rewrite "fixed" it. A regression test on it therefore guards a layer that provably was not the cause.

### The likely real, still-present root cause

`src/components/dls/Forms/Combobox/index.tsx`:

```ts
const [inputValue, setInputValue] = useState<string>(initialInputValue || '');
...
setFilteredItems(
  !inputValue ? items : items.filter((item) => item.label.toLowerCase().includes(inputValue.toLowerCase())),
);
```

The dropdown is filtered by a **substring** match against `inputValue`, and `inputValue` is *pre-seeded* with `initialInputValue`. `SelectRepetitionMode.tsx` passes `initialInputValue={toLocalizedVerseKey(verseKey, lang)}` and `AdvancedCopy/SelectorContainer.tsx` passes `initialInputValue={value}` — i.e. the currently selected verse key. Opening the selector does not clear it (`onSelectorClicked` only toggles `isOpened`), and `closeCombobox` restores the selected label, so it stays filtered on reopen.

Concretely, open repeat settings on a surah whose range starts at ayah 1: `inputValue = "36:1"` → the visible options are exactly `36:1, 36:10, 36:11 … 36:19` (Ya-Sin has 83 verses, so nothing above 19 matches). That is *precisely* the reported symptom — "the dropdown options only include ayahs 1 to 19, **with some missing**, depending on the surah" (2–9 are the missing ones; "depending on the surah" falls out of whether 1xx keys exist). It also explains the reporter's second complaint, that typing in the box behaves oddly.

I could not run a browser here, so please treat this as a high-confidence hypothesis rather than a confirmed repro — but it should be checked before anything closes #2088.

### Requested changes (self-review; GitHub disallows REQUEST_CHANGES on one's own PR, so this is posted as a review comment)

1. Drop `Closes #2088` (or close this PR). Do not auto-close a bug report that most likely still reproduces.
2. Either repurpose this into an actual fix for the Combobox pre-seeded-filter behaviour (e.g. don't filter until the user actually types / clear `inputValue` on open), or close this and post the analysis as a comment on #2088.
3. If a test-only PR is still wanted afterwards, uncheck "🐛 Bug fix" — this changes no production code — and correct the description inaccuracies noted inline.

No objection to the test itself as a piece of coverage; the objection is to what it is claimed to do.

wakqasahmed:
Thanks for the review — the premise was wrong and I've fixed this properly.

Confirmed independently: `generateChapterVersesKeys` was never the cause (already uncapped 18 months before #2088 was filed). Traced the real bug to `Combobox/index.tsx`: `inputValue` is seeded from `initialInputValue` (the currently selected verse key) on mount and never cleared when the dropdown opens, and the item list is filtered by a substring match against `inputValue`. For a selection like "36:1", opening the dropdown only shows items whose label contains "36:1" as a substring — `36:1, 36:10-36:19` — matching the reported symptom exactly.

Fix: `onSelectorClicked` now clears `inputValue` when transitioning from closed to open (single-select case), so all items show until the user actually types to filter. The closed-state label is unaffected since `closeCombobox` already restores it from the selected value.

Added a component-level regression test on `Combobox` itself (seeded with a stale value, opens it, asserts previously-hidden ayahs like `36:20`/`36:83` are now visible, plus a short-surah case and a still-filters-when-typing case) — confirmed it fails on the pre-fix code and passes after. Also removed the false causal framing from the earlier `verse.test.ts` addition and tightened it to a full-array comparison per your suggestion; kept it as plain coverage since it's harmless.

`npx vitest run src/components/dls/Forms/Combobox src/utils/verse.test.ts` — 26/26 pass. Lint clean on all changed files. Title/body updated to reflect the real fix; keeping `Closes #2088` since I'm confident this addresses the reported behavior for both consumers (`SelectRepetitionMode.tsx` and `AdvancedCopy/SelectorContainer.tsx`, both of which feed `Combobox` through `initialInputValue`).

wakqasahmed:
Hi @chatgpt-codex-connector — just checking in on this one, it's been about 4 days with CI green and mergeable. Is there anything needed from our side to help move it forward, or any rough idea on timing? No rush, just following up. Thanks for the review!

src/utils/verse.test.ts
```diff
@@ -198,4 +198,19 @@ describe('generateChapterVersesKeys', () => {
...
     expect(result).toEqual(expectedKeys);
   });
+
+  // Regression test for https://github.com/quran/quran.com-frontend-next/issues/2088
+  // The repeat-settings "to ayah" dropdown must include every ayah in the surah,
+  // not just the first 19, for surahs that have more than 19 verses.
+  it('should generate keys for every verse of a surah with more than 19 verses', async () => {
+    const chapterId = '2'; // Al-Baqarah has 286 verses
+    const chaptersData = await getAllChaptersData();
+    const result = generateChapterVersesKeys(chaptersData, chapterId, false);
```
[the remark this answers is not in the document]
wakqasahmed: Addressed in f4a47ef: added a component-level regression test directly on Combobox (src/components/dls/Forms/Combobox/index.test.tsx) that renders it seeded with a stale selection (e.g. "36:1"), fires the same click a user would use to open it, and asserts previously-hidden ayahs (36:2, 36:20, 36:83) are now visible. Verified this fails on the pre-fix code and passes after. Didn't render the full RepeatAudioModal/AdvancedCopy tree since the bug is entirely inside Combobox's own open/filter logic and both real consumers feed it through the same initialInputValue prop, so a direct test on the component is the tighter regression guard.

src/utils/verse.test.ts
wakqasahmed: Fixed — description and the verse.test.ts comment no longer claim a causal link to #2088. The comment now reads as plain long-surah coverage ("Coverage for a long surah: the existing tests above only exercised chapters 1, 112, and the not-found case...") and the real root cause (Combobox's stale inputValue filter) is fixed and covered separately.

src/utils/verse.test.ts
wakqasahmed: Fixed — replaced the five sampled assertions with a full-array comparison: expect(result).toEqual(Array.from({ length: 286 }, (_, i) => `2:${i + 1}`)), matching the toEqual pattern already used by the surrounding tests.

src/utils/verse.test.ts
wakqasahmed: Fixed — updated the PR description: corrected which chapters the pre-existing tests covered (1, 112, and the not-found case), corrected the component path (SelectRepetitionMode.tsx imports from components/Verse/AdvancedCopy/, not RepeatAudioModal/), and removed the unverified An-Nas checkbox claim — the Combobox test suite now actually includes an An-Nas (114, 6 verses) case, so that coverage exists and is accurately described.

